PR #75, opened 2020-01-24
Sapuzhak: solution

Mikhail-88:


src/transformStateWithClones.js
```diff
@@ -63,8 +63,43 @@
...
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const newState = { ...state };
+  const changesInNewState = [];
+
+  function addProperties(obj) {
+    for (const key in obj) {
+      newState[key] = obj[key];
+    }
+    changesInNewState.push({ ...newState });
```
irenhh: add a new line before block of code

src/transformStateWithClones.js
```diff
@@ -63,8 +63,43 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const newState = { ...state };
+  const changesInNewState = [];
```
irenhh:
```suggestion
  const transformedState = [];
```
it is enough

src/transformStateWithClones.js
```diff
@@ -63,8 +63,43 @@
...
+    changesInNewState.push({ ...newState });
+  }
+
+  function clear() {
+    for (const key in newState) {
+      delete newState[key];
+    }
+    changesInNewState.push({ ...newState });
+  }
+
```
irenhh: they look very similar, think about how to combine them

PR #78, opened 2020-01-26
solution

BohdanSalenik:


src/transformStateWithClones.js
```diff
@@ -64,7 +64,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const clone = Object.assign({}, state);
```
irenhh: you can also use ```const clone = {...state}```

irenhh:
```suggestion
  const clonedState = Object.assign({}, state);
```

irenhh: or
```suggestion
  const stateClone = Object.assign({}, state);
```

PR #84, opened 2020-01-27
task was solved

LeonidShv:
https://github.com/LeonidShv/js_stateful-clones/blob/develop/src/transformStateWithClones.js

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -64,7 +64,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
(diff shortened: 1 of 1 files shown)


src/transformStateWithClones.js
```diff
@@ -64,7 +64,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const changedState = { ...state };
+  const result = [];
+
+  for (let i = 0; i < transforms.length; i++) {
+    switch (transforms[i].operation) {
```
irenhh: use destructuring

PR #87, opened 2020-01-28
solution

maximkacmm:


src/transformStateWithClones.js
```diff
@@ -64,7 +64,47 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const newArr = [];
+  const newObject = Object.assign({}, state);
+
+  for (let i = 0; i < transforms.length; i++) {
+    const operations = transforms[i].operation;
+    const properties = transforms[i].properties;
```
irenhh: the same as for the prev homework

src/transformStateWithClones.js
```diff
@@ -64,7 +64,47 @@
...
+
+    switch (operations) {
+      case 'addProperties' :
+        let transitObj = {};
```
irenhh: why do you need this?

src/transformStateWithClones.js
irenhh: you didn't fix this

src/transformStateWithClones.js
```diff
@@ -64,7 +64,46 @@
...
+    const properties = transforms[i].properties;
+
+    switch (operations) {
+      case 'addProperties' :
+
+        for (const key in properties) {
+          newObject[key] = properties[key];
+        }
+
+        transitObj = Object.assign({}, newObject);
```
irenhh: why do you need this extra object?

PR #95, opened 2020-03-30
Solution

SeiAlek:


src/transformStateWithClones.js
```diff
@@ -64,7 +64,63 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const initialState = {};
+  const result = [];
```
vpolets: history?

SeiAlek: May be 'history' was a better name for this variable, but I have thought that was started state for every iteration and named it initialState.

PR #160, opened 2020-05-27
Solution

ogsevko:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,63 @@
...
+        statesArray.push(JSON.parse(JSON.stringify(stateClone)));
+        break;
+
+      case 'clear':
+        for (const key in stateClone) {
+          if (stateClone.hasOwnProperty(key)) {
+            delete stateClone[key];
+          }
+        };
+        statesArray.push(JSON.parse(JSON.stringify(stateClone)));
```
solaryasha, line 95: You repeated this block of code three times.
Better to right it one time after the switch loop.

src/transformStateWithClones.js
```diff
@@ -65,7 +65,63 @@
...
+        break;
+    }
+  }
+
```
solaryasha, line 99: Here.

src/transformStateWithClones.js
ogsevko: Yeah, makes sense.

PR #189, opened 2020-08-03
solved

DmJSCourse:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,31 @@
...
+        const toDelete = action['properties'][key];
+
+        delete lastState[toDelete];
+      }
+    }
+
+    if (action['operation'] === 'clear') {
+      for (const element in lastState) {
+        delete lastState[element];
+      }
```
vsm2k17, line 87: clear always delete everything, so we can just put in clones empty object
```suggestion
       lastState = {};
```

PR #193, opened 2020-08-03
solution

krajnova:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,34 @@
...
+      case 'removeProperties':
+        for (const key of currentTransform.properties) {
+          delete add[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in add) {
+          delete add[key];
+        }
```
vsm2k17, line 88:
```suggestion
      case 'clear':
      add = {};
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
...
+          delete currentTransform[key];
+        }
+        break;
+
+      case 'clear':
+        currentTransform = {};
+        break;
+
+      default:
+        break;
```
vsm2k17, line 88: no need to write break after default and you can throw an error here, or just show in the console that something wrong

PR #195, opened 2020-08-03
Solution

KaterinaSavvova:


src/transformStateWithClones.js
```diff
@@ -66,6 +66,30 @@
  */
 function transformStateWithClones(state, transforms) {
   // write code here
+  const ans = [];
+  const obj = { ...state };
+
+  for (const i of (transforms)) {
```
vkryvytskyy, line 72: what does i means?

src/transformStateWithClones.js
```diff
@@ -66,6 +66,30 @@
...
+    if (i.operation === 'addProperties') {
+      Object.assign(obj, i.properties);
+    }
+
+    if (i.operation === 'clear') {
+      for (const k in obj) {
```
vkryvytskyy, line 78: what does k means?

src/transformStateWithClones.js
```diff
@@ -66,6 +66,30 @@
...
+        delete obj[k];
+      }
+    }
+
+    if (i.operation === 'removeProperties') {
+      for (const a of (i.properties)) {
```
vkryvytskyy, line 84: what does a means?

src/transformStateWithClones.js
```diff
@@ -66,6 +66,30 @@
  */
 function transformStateWithClones(state, transforms) {
   // write code here
+  const ans = [];
+  const obj = { ...state };
+
+  for (const i of (transforms)) {
+    if (i.operation === 'addProperties') {
```
vkryvytskyy, line 73: use switch..case or if..else instead of separate ifs

src/transformStateWithClones.js
```diff
@@ -66,6 +66,26 @@
  */
 function transformStateWithClones(state, transforms) {
   // write code here
+  const ans = [];
```
vkryvytskyy, line 69: ans? what is ans?
```suggestion
  const result = [];
```

PR #196, opened 2020-08-03
stateful-clones

Elena-brv:


src/transformStateWithClones.js
```diff
@@ -64,8 +64,31 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const array = [];
```
vsm2k17, line 69: I think we can use better names for variables, array doesn't explain anything.

src/transformStateWithClones.js
```diff
@@ -64,8 +64,31 @@
...
+  let result = { ...state };
+
+  for (const obj of transforms) {
```
vsm2k17, line 72:
```suggestion
  for (const transform of transforms) {
```

Elena-brv: changed

src/transformStateWithClones.js
Elena-brv: changed

src/transformStateWithClones.js
```diff
@@ -64,8 +64,33 @@
...
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const result = [];
+  let transformed = { ...state };
+
+  for (const transform of transforms) {
+    switch (transform.operation) {
+      case 'addProperties':
+        Object.assign(transformed, transform.properties);
+        result.push({ ...transformed });
```
vsm2k17, line 76: you can write this code only ones outside of the switch

Elena-brv: done

PR #197, opened 2020-08-04
Solution

b0hdnan:


src/transformStateWithClones.js
```diff
@@ -66,6 +66,29 @@
  */
 function transformStateWithClones(state, transforms) {
   // write code here
+  const result = [];
+  const newState = { ...state };
+
+  for (let i = 0; i < transforms.length; i++) {
```
vsm2k17, line 72: consider of using "for of" loop

PR #199, opened 2020-08-04
solution

furude-rika:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,27 @@
...
+      for (const prop of elem.properties) {
+        delete arrElem[prop];
+      }
+      arr.push({ ...arrElem });
+    } else if (elem.operation === 'clear') {
+      for (const obj in arrElem) {
+        delete arrElem[obj];
+      }
+      arr.push({ ...arrElem });
+    }
```
rmnkk, line 85: If you will write arr.push({ ...arrElem }); after all conditions (after line 85) it will also work.

furude-rika: Thank you, I'll fix it

src/transformStateWithClones.js
```diff
@@ -65,7 +65,25 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const arr = [];
+  const arrElem = { ...state };
```
vsm2k17, line 69: arr? I thought that an object.

src/transformStateWithClones.js
```diff
@@ -65,7 +65,25 @@
...
+    if (elem.operation === 'addProperties') {
+      Object.assign(arrElem, elem.properties);
+    } else if (elem.operation === 'removeProperties') {
+      for (const prop of elem.properties) {
+        delete arrElem[prop];
+      }
+    } else if (elem.operation === 'clear') {
+      for (const obj in arrElem) {
+        delete arrElem[obj];
+      }
```
vsm2k17, line 81: arrElem = {}

src/transformStateWithClones.js
furude-rika: it's an object but it's also an arr's elem, that's why I called it in such a way
but know I think it's better to rename it

PR #200, opened 2020-08-04
solved

nanccyy:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,38 @@
...
-  // write code here
+  const clone = {};
+  const arrayWithClones = [];
+
+  Object.assign(clone, state);
+
+  for (const key in transforms) {
+    if (transforms[key].operation === 'addProperties') {
+      Object.assign(clone, transforms[key].properties);
+      arrayWithClones.push({ ...clone });
```
vkryvytskyy, line 76: do it once in the end of for loop after all ifs

src/transformStateWithClones.js
```diff
@@ -65,7 +65,38 @@
...
+    }
+
+    if (transforms[key].operation === 'removeProperties') {
+      for (const propertyKey in transforms[key].properties) {
+        for (const value in clone) {
+          if (transforms[key].properties[propertyKey] === value) {
+            delete clone[value];
+          }
+        }
+      } arrayWithClones.push({ ...clone });
```
vkryvytskyy, line 86: don't need inner loop here

src/transformStateWithClones.js
```diff
@@ -65,7 +65,38 @@
...
+
+    if (transforms[key].operation === 'clear') {
+      for (const prop in clone) {
+        if (clone.hasOwnProperty(prop)) {
+          delete clone[prop];
+        }
+      }
+      arrayWithClones.push({ ...clone });
+    }
+  }
```
vkryvytskyy, line 97: use switch case or if..else instead of separate ifs

src/transformStateWithClones.js
```diff
@@ -65,7 +65,38 @@
...
+          if (transforms[key].properties[propertyKey] === value) {
+            delete clone[value];
+          }
+        }
+      } arrayWithClones.push({ ...clone });
+    }
+
+    if (transforms[key].operation === 'clear') {
+      for (const prop in clone) {
+        if (clone.hasOwnProperty(prop)) {
```
vkryvytskyy, line 91: this if also could be removed

PR #205, opened 2020-08-04
Solution

RusTron:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  let currentArray = Object.assign({}, state);
+
+  for (let i = 0; i < transforms.length; i++) {
```
vkryvytskyy, line 70:
```suggestion
  for (const transform of transforms) {
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
...
+    switch (transforms[i].operation) {
```
vkryvytskyy, line 71:
```suggestion
    switch (transforms.operation) {
```
use for..of instead of usual for loop

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
...
+      case 'addProperties':
+        for (const key in transforms[i].properties) {
+          currentArray[key] = transforms[i].properties[key];
+        }
+        transforms[i] = Object.assign({}, currentArray);
```
vkryvytskyy, line 76:
```suggestion
        transforms.push({...currentArray});
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  let currentArray = Object.assign({}, state);
```
vkryvytskyy, line 68: i believe this is not an array

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
...
+      case 'addProperties':
+        for (const key in transforms[i].properties) {
+          currentArray[key] = transforms[i].properties[key];
+        }
+        transforms[i] = Object.assign({}, currentArray);
+        break;
+
+      case 'clear':
+        currentArray = {};
+        transforms[i] = Object.assign({}, currentArray);
```
vkryvytskyy, line 81: do you need to assign two empty objects?

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
...
+        break;
+
+      case 'removeProperties':
+        for (let j = 0; j < transforms[i].properties.length; j++) {
+          if (currentArray.hasOwnProperty(transforms[i].properties[j])) {
```
vkryvytskyy, line 86: what happens if you try remove property that doesn't exist on currentArray?

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
...
+  let currentObject = Object.assign({}, state);
+
+  for (let transform of transforms) {
+    switch (transform.operation) {
+      case 'addProperties':
+        for (const key in transform.properties) {
+          currentObject[key] = transform.properties[key];
+        }
+        transform = Object.assign({}, currentObject);
+        transforms.push(transform);
```
vkryvytskyy, line 77:
```suggestion
        transforms.push({ ...currentObject });
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
...
+
+      case 'clear':
+        currentObject = {};
+        transforms.push({});
+        break;
 
+      case 'removeProperties':
+        for (let j = 0; j < transform.properties.length; j++) {
+          delete currentObject[transform.properties[j]];
+        }
```
vkryvytskyy, line 88: use for..of loop instead of usual for

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
...
+        transform = Object.assign({}, currentObject);
+        transforms.push(transform);
+    }
+  }
+  transforms.splice(0, transforms.length / 2);
```
vkryvytskyy, line 93: transforms should not be changed

src/transformStateWithClones.js
```diff
@@ -65,7 +65,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
-}
+  let currentObject = Object.assign({}, state);
+
+  const reTransforms = transforms;
```
vkryvytskyy, line 70: image
do you believe that this is different objects?

src/transformStateWithClones.js
```diff
@@ -65,7 +65,34 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const property in transform.properties) {
+          delete currentObject[transform.properties[property]];
+        }
+        reTransforms.push({ ...currentObject });
+    }
+  }
+  reTransforms.splice(0, reTransforms.length / 2);
```
vkryvytskyy, line 94: you do not need to use splice in this solution

src/transformStateWithClones.js
```diff
@@ -83,15 +85,14 @@ function transformStateWithClones(state, transforms) {
         break;
 
       case 'removeProperties':
-        for (let j = 0; j < transform.properties.length; j++) {
-          delete currentObject[transform.properties[j]];
+        for (const property in transform.properties) {
```
vkryvytskyy, line 88: use for..of instead of for..in, since transform.properties is an array

RusTron: transform.properties is an object, so I use for ... in loop

PR #209, opened 2020-08-04
Develop

tykhan:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,25 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const modifiedObject = { ...state };
+  const resultArr = [];
+
+  for (let i = 0; i < transforms.length; i++) {
```
vkryvytskyy, line 71: use for..of instead of usual loop

PR #216, opened 2020-08-04
solution

zarichnyi:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,29 @@
...
+        for (const item of transform.properties) {
+          delete clone[item];
+        }
+        break;
 
+      case 'clear':
+        for (const item in clone) {
+          delete clone[item];
+        }
+    }
```
vkryvytskyy, line 87: always add default case to swtich

PR #222, opened 2020-08-04
Solution

uni-8:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
...
+          ...clonedState,
+          ...command.properties,
+        };
+        break;
+
+      case 'removeProperties':
+        for (const key of command.properties) {
+          if (clonedState[key]) {
+            delete clonedState[key];
+          }
```
vkryvytskyy, line 84: redundant if

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
...
+        }
+        break;
+
+      case 'clear':
+        clonedState = {};
+    }
```
vkryvytskyy, line 90: always add default case to switch

PR #227, opened 2020-08-05
one commit

zarva6596:


src/transformStateWithClones.js
```diff
@@ -66,6 +66,39 @@
...
+      case 'addProperties':
+        for (const key in transform.properties) {
+          stateCopy[key] = transform.properties[key];
+        }
+
+        break;
+      case 'removeProperties':
+        for (const key in transform.properties) {
+          delete stateCopy[transform.properties[key]];
+        }
```
vkryvytskyy, line 85:
```suggestion
        for (const key of transform.properties) {
          delete stateCopy[key];
        }
```

src/transformStateWithClones.js
```diff
@@ -66,6 +66,39 @@
...
+
+        break;
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+
+        break;
+    }
```
vkryvytskyy, line 94: always add default case to switch

PR #228, opened 2020-08-05
solution

ChrisZinch:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,34 @@
...
+        break;
+
+      // eslint-disable-next-line no-fallthrough
+      case 'clear':
+        for (const prop in obj) {
+          delete obj[prop];
+        }
+        arr.push({ ...obj });
+        break;
+    }
```
vkryvytskyy, line 92: add default case to switch

src/transformStateWithClones.js
```diff
@@ -65,7 +65,34 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key in properties) {
+          delete obj[properties[key]];
+        }
+        arr.push({ ...obj });
+        break;
+
+      // eslint-disable-next-line no-fallthrough
```
vkryvytskyy, line 85: I believe you can get rid of it

PR #234, opened 2020-08-06
solution

Oleg-Ozerov:


src/transformStateWithClones.js
```diff
@@ -66,6 +66,28 @@
...
+        break;
+      case 'removeProperties':
+        for (const property of transform.properties) {
+          delete stateCopy[property];
+        }
+        break;
+      case 'clear':
+        stateCopy = {};
+        break;
+    }
```
vkryvytskyy, line 85: always add default case to switch

PR #309, opened 2020-11-20
Solution

AndrewKreshchenko:
Firstly I was trying to push every object to my array "result" inside of for in loop.


vpolets (approved):
good! Наступного разу використовуй switch case

PR #320, opened 2020-12-08
Transform with clones - solution

VinnieJ-2k20:


Lara-l-art (approved):
Nice!

src/transformStateWithClones.js
```diff
@@ -65,7 +65,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const result = [];
+  let subject = {};
```
varseniuk, line 69: What if there will be a not empty 'state' in arguments? What if you will receive as an argument
  'state' = {foo: 'bar', bar: 'foo'} and as a 'transform' you will receive [ {
  operation: 'removeProperties', properties: ['bar']
}]

How will you remove the property 'bar' from your empty object 'subject' if it's already declared as empty?

src/transformStateWithClones.js
```diff
@@ -65,7 +65,30 @@
...
+
+  Object.assign(subject, state);
+
+  for (const transform of transforms) {
+    if (transform.operation === 'addProperties') {
```
varseniuk, line 74: Let's use `switch...case` instead of `if`

varseniuk (changes requested):

src/transformStateWithClones.js
VinnieJ-2k20: Sorry, I don't completely understand.
'subject' here is used as a shell for first instance of 'state'.
In line 71 `Object.assign(subject, state);` it is always assigned with all 'state' properties.
In case you described everything works fine (screenshot below).
https://prnt.sc/vzltny

Please provide additional clarification.

varseniuk: If you want to use 'subject' as a shell for 'state', so it should be A COPY of 'state', not just an empty object.
On line 69 you are declaring EMPTY object, from which you cannot remove anything in case 'removeProperties', because there is nothing there. And if 'state' contains something, the 'subject' is empty (as you declared). So it's not a shell in this case.

In the example, you displayed on the screenshot, you are declaring the variable 'stateTest' with fields in it. Of course, it will work. But it won't work with an empty object

varseniuk (changes requested):
Commented about empty 'shell'

src/transformStateWithClones.js
VinnieJ-2k20: Yes, on line 69 it's empty, but after line 71 it becomes a copy of 'state'.
And a loop afterwards operates on this copy.

varseniuk: You're right!)
That's why we want you to write more simple code. If you would just write
`let subject = { ...state };`
it would be MUCH MORE readable and understandable and wouldn't confuse me!
Thank you for the discussion and sorry for my inattention.

Good job!

varseniuk (approved):
Good job!

src/transformStateWithClones.js
VinnieJ-2k20: No problem. I`ll definately take it into consideration in the future tasks.

PR #329, opened 2020-12-10
solution

5Mountains:


xipholena (approved):
Класна робота! Але я б подумала над іншим ім’ям замість result. Наприклад, updatedState або changedState

src/transformStateWithClones.js
```diff
@@ -65,7 +65,28 @@
...
+      case 'addProperties':
+        Object.keys(item.properties).forEach(key => {
+          currentState[key] = item.properties[key];
+        });
+        break;
+      case 'removeProperties':
+        item.properties.forEach(key => delete currentState[key]);
+        break;
+      case 'clear':
+        Object.keys(currentState).forEach(key => delete currentState[key]);
```
varseniuk, line 82: It's ok but you also may use 'for...in' loop to iterate through object keys and not creating a new array of them using 'Object.keys;
```        
for (const key in currentState) {
  delete currentState[key];
}
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,28 @@
...
+        break;
```
varseniuk, line 83: Add 'default' case

varseniuk (changes requested):

src/transformStateWithClones.js
5Mountains: Okay, I'll try it!

src/transformStateWithClones.js
5Mountains: my bad (:

varseniuk (approved):
Great job 👍

PR #330, opened 2020-12-10
Solution

vovador:


olesiatk (approved):

src/transformStateWithClones.js
```diff
@@ -65,7 +65,55 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const states = [];
+
+  for (let i = 0; i < transforms.length; i++) {
```
vsm2k17, line 70: "for of" loop looks better here

src/transformStateWithClones.js
```diff
@@ -65,7 +65,55 @@
...
+    if (transforms[i].operation === 'addProperties') {
+      let temporalState;
```
vsm2k17, line 72: you can declare this variable a little bit upper. no need to repeat this code in every if-block

src/transformStateWithClones.js
```diff
@@ -65,7 +65,55 @@
...
+
+      if (i === 0) {
+        temporalState = { ...state };
+      } else {
+        temporalState = { ...states[i - 1] };
+      }
```
vsm2k17, line 78: redundant

src/transformStateWithClones.js
```diff
@@ -65,7 +65,55 @@
...
+
+      for (const key in transforms[i].properties) {
+        temporalState[key] = transforms[i].properties[key];
+      }
+      states.push(temporalState);
```
vsm2k17, line 83: DRY (don't repeat yourself), you can push at the end of the loop

src/transformStateWithClones.js
```diff
@@ -65,7 +65,55 @@
...
+
+      if (i === 0) {
+        temporalState = { ...state };
+      } else {
+        temporalState = { ...states[i - 1] };
+      }
+
+      for (let j = 0; j < properties.length; j++) {
+        delete temporalState[properties[j]];
+      }
```
vsm2k17, line 96: "for of" loop looks better here

src/transformStateWithClones.js
```diff
@@ -65,7 +65,55 @@
...
+      let temporalState;
+
+      if (i === 0) {
+        temporalState = { ...state };
+      } else {
+        temporalState = { ...states[i - 1] };
+      }
+
+      for (const key in temporalState) {
+        if (temporalState.hasOwnProperty(`${key}`)) {
```
vsm2k17, line 108: redundant check

vsm2k17 (changes requested):
great work, but consider use switch case construction or "if-else" at least because at this moment your code can be triggered on every if block (it's impossible, we know it, but still no need to organize logic this way), need to write blocks of code counting on behavior when only one block can be triggered. that's why I'm suggesting to use a switch case.

src/transformStateWithClones.js
```diff
@@ -65,7 +65,41 @@
...
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const states = [];
+  let temporalState = { ...state };
+
+  for (let i = 0; i < transforms.length; i++) {
+    if (i !== 0) {
+      temporalState = { ...states[i - 1] };
+    }
```
Denchepornyuk, line 74: це тут точно потрібно писати?

src/transformStateWithClones.js
```diff
@@ -65,7 +65,41 @@
...
+
+    switch (transforms[i].operation) {
+      case 'addProperties':
+
+        for (const key in transforms[i].properties) {
+          temporalState[key] = transforms[i].properties[key];
+        }
```
Denchepornyuk, line 81: для addProperties не потрібно робити цикл, зроби це через Object.assign

Denchepornyuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -65,7 +65,39 @@
...
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const states = [];
+  const temporalState = { ...state };
+
+  for (const instruction of transforms) {
+    // if (i !== 0) {
+    //   temporalState = { ...states[i - 1] };
+    // }
```
aionova, line 74: remove comments

src/transformStateWithClones.js
```diff
@@ -65,7 +65,39 @@
...
+      case 'addProperties':
+
+        Object.assign(temporalState, instruction.properties);
+        break;
+
+      case 'removeProperties' :
+
+        const { properties } = instruction;
+
+        for (const y of properties) {
```
aionova, line 86:
```suggestion
        for (const property of properties) {
```

aionova (approved):
Good job!

PR #331, opened 2020-12-10
solution

xipholena:


5Mountains:
Well, job! It looks very compact. 😇

5Mountains (approved):

src/transformStateWithClones.js
```diff
@@ -65,7 +65,31 @@
...
+        for (const empty of properties) {
+          delete stateClone[empty];
+        };
+        break;
+
+      case 'clear':
+        for (const erased in stateClone) {
+          delete stateClone[erased];
+        };
+        break;
```
varseniuk, line 87: Add 'default' case

varseniuk (changes requested):
Very good solution! 🚀 
Just get used to adding 'default' case for 'switch...case' statement

varseniuk (approved):
👍

PR #333, opened 2020-12-10
solution

dima-bond:


khrulenko (approved):

src/transformStateWithClones.js
```diff
@@ -65,7 +65,35 @@
...
+    }
+
+    for (const key in transforms[i]) {
+      if (key === 'properties') {
+        continue;
+      }
+
+      if (transforms[i][key] === 'addProperties') {
+        for (const addprops in transforms[i]['properties']) {
+          resultingArray[i][addprops] = transforms[i]['properties'][addprops];
```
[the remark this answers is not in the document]
vsm2k17: add prop sound like a function name, if you call variable. use something like prop or propForAdd or addedProp

src/transformStateWithClones.js
```diff
@@ -65,7 +65,35 @@
...
+        }
+      } else if (transforms[i][key] === 'removeProperties') {
+        for (let j = 0; j < transforms[i]['properties'].length; j++) {
```
vsm2k17, line 87: "for of" loop looks better here

src/transformStateWithClones.js
```diff
@@ -65,7 +65,35 @@
...
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const resultingArray = [];
+
+  for (let i = 0; i < transforms.length; i++) {
+    if (i === 0) {
+      resultingArray[i] = { ...state };
+    } else {
+      resultingArray[i] = { ...resultingArray[i - 1] };
+    }
```
vsm2k17, line 75: just create a temporary object and push it at the end of the loop

src/transformStateWithClones.js
```diff
@@ -65,7 +65,28 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const resultingArray = [];
+  const objTmp = { ...state };
```
```suggestion
  const temporaryState = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,28 @@
...
+
+  for (const arrayItem of transforms) {
```
aionova, line 71:
```suggestion
  for (const transformation of transforms) {
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,28 @@
...
+    if (arrayItem.operation === 'addProperties') {
+      for (const addprops in arrayItem.properties) {
```
aionova, line 73:
```suggestion
      for (const property in transformation.properties) {
```
When I read 'addprops in arrayItem.properties', I can't understand anything about these variables. Give names that help understand what these variables contain

src/transformStateWithClones.js
```diff
@@ -66,31 +66,24 @@
...
+        objTmp[addprops] = arrayItem.properties[addprops];
+      }
+    } else if (arrayItem.operation === 'removeProperties') {
+      for (const propertiesItem of arrayItem.properties) {
+        delete objTmp[propertiesItem];
+      }
+    } else if (arrayItem.operation === 'clear') {
+      for (const deleteKeys in objTmp) {
+        delete objTmp[deleteKeys];
```
aionova, line 82: deleteKeys is not a noun, so it's not a good name for a variable. Moreover, you can just write 
`objTmp = {}` instead of this loop.

src/transformStateWithClones.js
```diff
@@ -66,31 +66,24 @@
  */
 function transformStateWithClones(state, transforms) {
   const resultingArray = [];
```
aionova, line 68: No need to include 'Array' in the name of your variable. 'clonedStates' would describe this variable better.

aionova (changes requested):

aionova (approved):
Good job!

PR #334, opened 2020-12-10
solution

NazarDovhanchuk:


MarchD (approved):
Good solution!

src/transformStateWithClones.js
```diff
@@ -65,7 +65,28 @@
...
+      case 'removeProperties' :
+        for (const key of transform.properties) {
+          delete clone[key];
+        }
+        break;
+
+      case 'clear' :
+        clone = {};
+        break;
+    }
```
varseniuk, line 85: Add 'default' case

varseniuk (changes requested):
All is ok. Just add 'default' case for 'switch...case' statement.

varseniuk (approved):

PR #339, opened 2020-12-11
Solution

Layrovell:


olesiatk (approved):

src/transformStateWithClones.js
```diff
@@ -64,8 +64,44 @@
...
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, transforms) {
-  // write code here
+  // const stateClone = {};
+
+  // for (const key in state) {
+  //   stateClone[key] = state[key];
+  // }
```
varseniuk, line 73: Remove this. Or do you need it?)

src/transformStateWithClones.js
```diff
@@ -64,8 +64,44 @@
...
+
+  const stateClone = { ...state };
+
+  const arr = [];
```
varseniuk, line 77: Again, try to give more meaningful names to variables

src/transformStateWithClones.js
```diff
@@ -64,8 +64,44 @@
...
+        break;
+
+      case 'clear':
+        for (const prop in stateClone) {
+          delete stateClone[prop];
+        }
+        break;
+
+      default:
+        // do nothing
```
varseniuk, line 98: Don't leave 'default' state empty. Add at least 'break'.
It is useful to add 'default' case because there may be a lot of different kinds of problems so there always should be a default case to break the function's execution and exclude unpredictable consequences.

varseniuk (approved):
Good solution 👍 
Read comments and take them into consideration for future

PR #346, opened 2020-12-12
Solution

Roman-Matsuk:


YuriiFitsyk (approved):
Interesting idea!

src/transformStateWithClones.js
```diff
@@ -65,7 +65,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const resulArr = [];
```
aionova, line 68:
```suggestion
  const clonedStates = [];
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,41 @@
...
+  const copy = { ...state };
+  let add;
+  let remove;
+  let deleteAll;
+
+  for (const { operation, properties } of transforms) {
```
aionova, line 74: great use of destructuring!

src/transformStateWithClones.js
```diff
@@ -65,7 +65,41 @@
...
+    switch (operation) {
+      case 'addProperties':
+        for (const key in properties) {
+          copy[key] = properties[key];
+        }
+        add = { ...copy };
+        resulArr.push(add);
```
aionova, line 81:
```suggestion
        resulArr.push({ ...copy });
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,41 @@
...
+        break;
+
+      case 'clear':
+        for (const key in copy) {
+          delete copy[key];
+        }
+        deleteAll = { ...copy };
+        resulArr.push(deleteAll);
+        break;
+    }
```
aionova, line 99: You can add resulArr.push({ ...copy }) after your switch case, and then you won't need to duplicate it in every case.

aionova (changes requested):

aionova (changes requested):
Don't repeat yourself. As I've written previously, you don't need to repeat clonedStates.push({ ...copy })

aionova (approved):
Good job!

PR #358, opened 2020-12-13
Solution

ira-cherpak:


src/transformStateWithClones.js
```diff
@@ -66,6 +66,29 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key in transform.properties) {
+          delete newState[transform.properties[key]];
+        }
+        break;
+
+      case 'clear':
+        newState = {};
```
MarchD, line 85: You need to use "default" case

src/transformStateWithClones.js
```diff
@@ -66,6 +66,29 @@
  */
 function transformStateWithClones(state, transforms) {
   // write code here
+  const stateArr = [];
+  let newState = Object.assign({}, state);
```
MarchD, line 70: I think it is better to use `{...state}` in this situation

MarchD (approved):
good job! I left some comments

src/transformStateWithClones.js
aionova: Agree with Marina. Destructuring is shorter and is easy to read.

src/transformStateWithClones.js
aionova: Agree again) It's better to always include default case for error handling

src/transformStateWithClones.js
```diff
@@ -66,6 +66,29 @@
...
+        for (const key in transform.properties) {
+          delete newState[transform.properties[key]];
+        }
+        break;
+
+      case 'clear':
+        newState = {};
+    }
+
+    stateArr.push(Object.assign({}, newState));
```
aionova, line 88:
```suggestion
    stateArr.push({ ...newState });
```

src/transformStateWithClones.js
```diff
@@ -66,6 +66,29 @@
  */
 function transformStateWithClones(state, transforms) {
   // write code here
+  const stateArr = [];
```
```suggestion
  const clonedStates = [];
```
Choose meaningful names

aionova (changes requested):

Denchepornyuk (approved):

PR #359, opened 2020-12-13
Solution

taniazt:


src/transformStateWithClones.js
```diff
@@ -66,6 +66,29 @@
  */
 function transformStateWithClones(state, transforms) {
   // write code here
+  const result = [];
+  const clone = Object.assign({}, state);
+
+  for (const obj of transforms) {
+    if (obj.operation === 'addProperties') {
```
MarchD, line 73: It is better to use 'switch case'

MarchD (changes requested):
Good job! Change if else:)

MarchD (approved):

Denchepornyuk (approved):

PR #426, opened 2021-03-02
Solution

herkoss:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const newState = { ...state };
+  const resArr = [];
```
```suggestion
  const states = [];
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,29 @@
...
+
+  for (let i = 0; i < transforms.length; i++) {
```
aionova, line 71: 'for of' loop will fit better here

src/transformStateWithClones.js
```diff
@@ -65,7 +65,29 @@
...
+    if (transforms[i]['operation'] === 'addProperties') {
```
aionova, line 72: let's use switch case

aionova (changes requested):
Good job! Let's refactor it now

aionova (approved):
Looks much better now!
'states.push({ ...newState })' can be written only once after switch case construction

PR #442, opened 2021-03-04
Created function transformStateWithClones

haiduk2019:


src/transformStateWithClones.js
```diff
@@ -66,6 +66,38 @@
  */
 function transformStateWithClones(state, transforms) {
   // write code here
-}
+  const newStates = [];
+  const newState = {
+    ...state,
+  };
+
+  for (const transform of transforms) {
```
OlenaMelnyk, line 74: You can do desctructuring
for (const [operation, poperties] of transforms)
and then will be less repeated code

src/transformStateWithClones.js
```diff
@@ -66,6 +66,38 @@
...
+    switch (transform.operation) {
+      case 'addProperties':
+        for (const key in transform.properties) {
+          newState[key] = transform.properties[key];
+        }
+        // newState = Object.assign(newState, transform.properties);
```
OlenaMelnyk, line 80: do not leave commented code

src/transformStateWithClones.js
```diff
@@ -66,6 +66,38 @@
...
+        newStates.push(Object.assign({}, newState));
```
OlenaMelnyk, line 81: you can use spread operator as well (see line 70). Better use spread operator and not mixing them.

OlenaMelnyk (changes requested):

danheim (approved):

PR #451, opened 2021-03-12
Solution

vasylevskaya:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const result = [];
+  const newObj = { ...state };
+
+  for (const obj in transforms) {
```
vkryvytskyy, line 71:
```suggestion
  for (const transform of transforms) {
```
use of with arrays

src/transformStateWithClones.js
```diff
@@ -65,7 +65,34 @@
...
+    if (transforms[obj].operation === 'addProperties') {
```
vkryvytskyy, line 72:
```suggestion
    if (transform.operation === 'addProperties') {
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,34 @@
...
+          delete newObj[key];
+        }
+      }
+    }
+
+    if (transforms[obj].operation === 'clear') {
+      for (const key in newObj) {
+        delete newObj[key];
+      }
+    }
```
vkryvytskyy, line 90: consider to use switch for such cases:
```
switch(transform.operation) {
  case 'addProperties':

  case 'removeProperties':
}

vkryvytskyy (approved):

PR #452, opened 2021-03-13
solution

SerhiiKolesnyk:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const result = [];
```
victor-arutiunov, line 68: I think that it will be better to replace "result" list name by more expressive name, like "statesList"

src/transformStateWithClones.js
```diff
@@ -65,7 +65,34 @@
...
+  const newObj = { ...state };
+
+  for (const obj in transforms) {
+    if (transforms[obj].operation === 'addProperties') {
+      for (const key in transforms[obj].properties) {
+        newObj[key] = transforms[obj].properties[key];
+      }
+    }
+
+    if (transforms[obj].operation === 'removeProperties') {
```
victor-arutiunov, line 78: For my opinion it would be better to use if else or switch case

src/transformStateWithClones.js
```diff
@@ -65,7 +65,34 @@
...
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const result = [];
+  const newObj = { ...state };
+
+  for (const obj in transforms) {
+    if (transforms[obj].operation === 'addProperties') {
+      for (const key in transforms[obj].properties) {
+        newObj[key] = transforms[obj].properties[key];
```
victor-arutiunov, line 74: You can push the result immediately in result list instead of using newObj;

src/transformStateWithClones.js
aliceblunt8: 'result' sounds okay

src/transformStateWithClones.js
```diff
@@ -65,7 +65,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const result = [];
+  const newObj = { ...state };
+
+  for (const obj in transforms) {
+    if (transforms[obj].operation === 'addProperties') {
```
aliceblunt8, line 72: let's use switch case

src/transformStateWithClones.js
aliceblunt8: yes, use switch case

src/transformStateWithClones.js
```diff
@@ -65,7 +65,34 @@
...
+      for (const key in transforms[obj].properties) {
+        newObj[key] = transforms[obj].properties[key];
+      }
+    }
+
+    if (transforms[obj].operation === 'removeProperties') {
+      for (const key of transforms[obj].properties) {
+        if (newObj.hasOwnProperty(key)) {
```
aliceblunt8, line 80: no need for hasOwnProperty check

src/transformStateWithClones.js
aliceblunt8: you can use Object.assign here

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #454, opened 2021-03-16
Solution

Andrey203:


src/transformStateWithClones.js
```diff
@@ -66,6 +66,42 @@
  */
 function transformStateWithClones(state, transforms) {
   // write code here
+  const arr = [];
+  const obj = { ...state };
```
aionova, line 70:
```suggestion
  const states = [];
  const stateClone = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -66,6 +66,42 @@
...
+
+  transforms.forEach((item) => {
```
aionova, line 72:
```suggestion
  transforms.forEach((transform) => {
```

aionova (changes requested):
Good job overall!
But arr and obj are no good names for variables) Let's fix it

aionova (approved):
You can write 'states.push({ ...stateClone });' only once in each iteration after switch case construction.

PR #458, opened 2021-03-22
Solution

senkiv-oleh:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const array = [];
```
Ihor-Karpyn, line 68: historyChanges will better for this variables

src/transformStateWithClones.js
```diff
@@ -65,7 +65,31 @@
...
+      array.push({ ...temp });
+    }
+
+    if (transforms[i].operation === 'clear') {
+      for (const keys in temp) {
+        delete temp[keys];
+      }
+      array.push({ ...temp });
+    }
+  }
```
Ihor-Karpyn, line 90: In this case switch will better.

src/transformStateWithClones.js
```diff
@@ -65,7 +65,31 @@
...
+    if (transforms[i].operation === 'addProperties') {
+      Object.assign(temp, transforms[i].properties);
+      array.push({ ...temp });
+    }
+
+    if (transforms[i].operation === 'removeProperties') {
+      for (let k = 0; k < (transforms[i].properties).length; k++) {
+        delete temp[transforms[i].properties[k]];
+      }
+      array.push({ ...temp });
```
Ihor-Karpyn, line 81:
```suggestion
      array.push({ ...temp });
```

You do this 3 time. you can push changed state at the end of each loop

Ihor-Karpyn (changes requested):

Ihor-Karpyn (approved):
Good job!

VinnieJ-2k20 (changes requested):
One small edit left.
`switch/case` block should always have `default` option in case of irregular inputs.
Even if it doesn't perform anything, it should be there.

VinnieJ-2k20 (approved):
For now `default` case can do nothing, but usually it returns an error.

PR #470, opened 2021-04-21
Solution

igor-stasiv:


Ihor-Karpyn (commented):
Similar to the previous task

Ihor-Karpyn (approved):

VinnieJ-2k20 (changes requested):
For this task it's also better to use `switch/case` instead of `if`.

VinnieJ-2k20 (approved):
Good job 👍

PR #471, opened 2021-04-21
solution

Ihor-Karpyn:


igor-stasiv (approved):

src/transformStateWithClones.js
```diff
@@ -65,7 +65,35 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of transform.properties) {
+          delete initialState[key];
+        }
+        changesList.push({ ...initialState });
+        break;
+    }
+  }
```
aliceblunt8, line 94: you can push changed state at the end of each loop

```suggestion
  for (const transform of transforms) {
    switch (transform.operation) {
      case 'clear':
        initialState = {};
        break;

      case 'addProperties':
        Object.assign(initialState, transform.properties);
        break;
        
      case 'removeProperties':
        for (const key of transform.properties) {
          delete initialState[key];
        }
        break;
    }
    
     changesList.push({ ...initialState });
  }
```

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #472, opened 2021-04-21
first solution

scarlettjohanssons:


src/transformStateWithClones.js
```diff
@@ -66,6 +66,44 @@
...
+  for (let i = 0; i < transforms.length; i++) {
+    const obj = transforms[i];
+
+    if (obj.operation === 'addProperties') {
+      const object = obj.properties;
+
+      for (const key in object) {
+        result[key] = object[key];
+      }
+      massResult.push({ ...result });
```
Ihor-Karpyn, line 84: You push result 3 times.Do this at the end of the for

Ihor-Karpyn (approved):
Excepе this comments, read comments to previous task

src/transformStateWithClones.js
```diff
@@ -66,6 +66,44 @@
  */
 function transformStateWithClones(state, transforms) {
   // write code here
+  let result = {};
+
+  result = Object.assign(result, state);
```
aliceblunt8, line 71:
```suggestion
  const result = {...state};
```

src/transformStateWithClones.js
aliceblunt8: agree, move push to the end of the for loop

src/transformStateWithClones.js
```diff
@@ -66,6 +66,44 @@
...
+
+  const massResult = [];
+
+  for (let i = 0; i < transforms.length; i++) {
```
aliceblunt8, line 75: let's use 'for of' loop and switch case instead of if's

src/transformStateWithClones.js
```diff
@@ -66,6 +66,44 @@
...
+    }
+
+    if (obj.operation === 'removeProperties') {
+      const mass = obj.properties;
+
+      for (let j = 0; j < mass.length; j++) {
+        for (const key in result) {
+          if (key === mass[j]) {
+            delete result[key];
+          }
```
aliceblunt8, line 99: just iterate through properties, get rid of nested loop

aliceblunt8 (changes requested):

src/transformStateWithClones.js
```diff
@@ -66,6 +66,30 @@
...
+        break;
+      case 'removeProperties':
+        for (const properties of transform.properties) {
+          delete result[properties];
+        }
+        massResult.push({ ...result });
+        break;
+      case 'clear':
+        result = {};
+        massResult.push({ ...result });
```
aliceblunt8, line 87: move push to the end of the for loop

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #473, opened 2021-04-21
solution

victor-arutiunov:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,45 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const statesList = [];
+
+  for (const transform of transforms) {
+    if (transform.operation === 'addProperties') {
```
aionova, line 71: switch case will fit perfect here

aionova (approved):
Great naming!
I recommend using switch case here though

PR #475, opened 2021-04-22
solution

vitalya2626:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const clone = Object.assign({}, state);
+  const final = [];
```
IvanGrekov, line 69: I think `final ` is uncertain name for our array with changes of state

src/transformStateWithClones.js
```diff
@@ -65,7 +65,31 @@
...
+        for (const key of transform.properties) {
+          delete clone[key];
+        }
+        break;
+
+      case 'clear' :
+        for (const key in clone) {
+          delete clone[key];
+        }
+        break;
```
IvanGrekov, line 87: `break` in the end again)

IvanGrekov (changes requested):
Everything is good, but there are a couple of remarks)

VinnieJ-2k20 (approved):
Good job 👍

PR #476, opened 2021-04-22
solution

IvanKolodiazhnyi:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,37 @@
...
+          delete clone[value];
+        }
+        states.push(Object.assign({}, clone));
+        break;
+
+      case 'clear':
+        for (const key in clone) {
+          delete clone[key];
+        }
+        states.push(Object.assign({}, clone));
```
aliceblunt8, line 93: no need to delete every properties in clone, just add to states an empty object

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #478, opened 2021-04-22
solution

provolovskyi-illia:


nazarmatsevych (approved):
Great !! 👍

VinnieJ-2k20 (changes requested):
`switch/case` block should always have `default` option in case of irregular inputs.
Even if it doesn't perform anything, it should be there.

VinnieJ-2k20 (approved):
Good job 👍

PR #479, opened 2021-04-22
Solution v0.1

SkyPKWolf:


bohdanklius (approved):
Круто!
Можна використати конструкцію "switch" замість "if"))

src/transformStateWithClones.js
```diff
@@ -65,7 +65,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const transrormsHistory = [];
```
aliceblunt8, line 68:
```suggestion
  const transformsHistory = [];
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,31 @@
...
+  const copyState = { ...state };
+
+  for (const transform of transforms) {
+    if (transform.operation === 'addProperties') {
```
aliceblunt8, line 72: I think it's better to use switch case for code's readability, but no strong opinion ;)

src/transformStateWithClones.js
```diff
@@ -65,7 +65,31 @@
...
+      Object.assign(copyState, transform.properties);
+      transrormsHistory.push({ ...copyState });
```
aliceblunt8, line 74: you push state 3 times in each if, but you can do it only once at the end of each iteration

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #481, opened 2021-04-22
Solution

IvanVasylenko:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const stateArr = [];
+  const transformationState = { ...state };
```
illisogori, line 69: Лучше не писать Arr в переменных.
```suggestion
  const transformedObjects = [];
  const transformedObject = { ...state };
```

illisogori (approved):
Хорошо.

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
...
+        transformationState[prop] = item.properties[prop];
+      }
+      stateArr.push({ ...transformationState });
+    }
+
+    if (item.operation === 'removeProperties') {
+      for (const prop of item.properties) {
+        delete transformationState[prop];
+      }
+      stateArr.push({ ...transformationState });
```
VinnieJ-2k20, line 83: Don't repeat yourself, Igor's comment is useful.

VinnieJ-2k20 (changes requested):
Let's use `switch/case` for this solution.
Also, don't forget about `default` option in case of irregular inputs.
In this task it can do nothing or throw an error.

varseniuk (approved):
Good job! 👍

PR #484, opened 2021-04-22
Solution

TimurHrynenko:


BoikoEva:
Good job!

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const newStates = [];
+  const newState = { ...state };
+
+  transforms.forEach(move => {
```
aionova, line 71:
```suggestion
  transforms.forEach({ operation, properties } => {
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
...
+    switch (move.operation) {
+      case 'addProperties':
+        Object.assign(newState, move.properties);
+        newStates.push({ ...newState });
```
aionova, line 75: newStates.push({ ...newState }) can be written only once after switch case construction

aionova (approved):
Good job!

PR #500, opened 2021-04-23
test

C-y-y-y:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  let copy = { ...state };
+  const arr = [];
```
aliceblunt8, line 69:
```suggestion
  const transformChanges = [];
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  let copy = { ...state };
```
aliceblunt8, line 68:
```suggestion
  let stateCopy = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
...
+  const arr = [];
+
+  for (const transform of transforms) {
+    copy = { ...copy };
```
aliceblunt8, line 72: you've already created a copy of state earlier, no need to do it on each loop iteration

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
...
+    arr.push(copy);
```
aliceblunt8, line 73: let's push to state at the end of each loop, not at the beginning

aliceblunt8 (changes requested):

src/transformStateWithClones.js
```diff
@@ -65,7 +65,33 @@
...
 function transformStateWithClones(state, transforms) {
-  // write code here
+  let stateCopy = { ...state };
+  const transformChanges = [];
+
+  for (const transform of transforms) {
+    stateCopy = { ...stateCopy };
+    transformChanges.push(stateCopy);
+
+    switch (transform.operation) {
```
mykola-yarchuk, line 75: usually in switch case construction developers adds default case, it's a common issue that there is no default case, but it would be nice to have something like 
```
default: 
  // Do nothing
```

or 
```
default: 
```

mykola-yarchuk (approved):
Good job. Left some comment as advice

PR #503, opened 2021-04-23
js_stateful-clones

Bogdan-Kotsupey:


Grygoriy-Shytikov (approved):
Виглядає добре, але краще було б з використанням swich case (на мою думку)

src/transformStateWithClones.js
```diff
@@ -65,7 +65,36 @@
...
+  const copy = { ...state };
+  const result = [];
+
+  for (const { operation, properties } of transforms) {
+    if (operation === 'addProperties') {
+      for (const key in properties) {
+        copy[key] = properties[key];
+      }
+
+      result.push({ ...copy });
```
mykola-yarchuk, line 77: you can push a copied object into results, outside of the loop, in that case, you write this line only once instead of three

src/transformStateWithClones.js
```diff
@@ -65,7 +65,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const copy = { ...state };
```
mykola-yarchuk, line 68: I think `copy` is not the best name, think about something more specific, like `copyOfState`, or `temporaryState`, or something else

mykola-yarchuk (approved):
good job left some comments as advice

PR #507, opened 2021-04-24
done

cheytac200:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  let newState = { ...state };
+  const cloneState = [];
```
vitaliy-buchinchik, line 69: It`s better to use another name of the variable for massive, like "states" for example. It`s because your newState is actually the state clone

vitaliy-buchinchik (approved):

src/transformStateWithClones.js
ArtemTopolskyi, line 69:
```suggestion
  let cloneState = { ...state };
  const states = [];
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,29 @@
...
+
+  for (const transform of transforms) {
+    switch (transform.operation) {
+      case 'addProperties':
+        Object.assign(newState, transform.properties);
+        break;
+
+      case 'removeProperties':
+        for (const key of transform.properties) {
+          delete newState[key];
```
ArtemTopolskyi, line 79:
```suggestion
        for (const property of transform.properties) {
          delete newState[property];
```

ArtemTopolskyi (changes requested):

ArtemTopolskyi (approved):

PR #511, opened 2021-04-25
solution

Namart:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const result = [];
+  let copy = { ...state };
+
+  for (const transform of transforms) {
+    switch (transform.operation) {
```
mykola-yarchuk, line 72: usually in switch case construction developers adds default case, it's a common issue that there is no default case, but it would be nice to have something like 
```
default: 
  // Do nothing
```

or 
```
default: 
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const result = [];
+  let copy = { ...state };
```
mykola-yarchuk, line 69: I think `copy` is not the best name, think about something more specific, like `copyOfState`, or `temporaryState`, or something else

mykola-yarchuk (approved):
Good job. Left some comments as advice

PR #512, opened 2021-04-25
first commit for review

NikolayButylkov:


src/transformStateWithClones.js
```diff
@@ -65,7 +65,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const result = [];
+  let stateClone = Object.assign({}, state);
```
vitaliy-buchinchik, line 69: there You can use spread operator ... instead of Object.asign. Something like that: ``` let stateClone = { ... state }

src/transformStateWithClones.js
```diff
@@ -65,7 +65,35 @@
...
+
+  for (let i = 0; i < transforms.length; i++) {
+    const transformsIndex = transforms[i];
+
+    if (transformsIndex.operation === 'addProperties') {
```
vitaliy-buchinchik, line 74: Better to use switch-case and not to use if()

src/transformStateWithClones.js
```diff
@@ -65,7 +65,35 @@
...
+      for (const add in transformsIndex.properties) {
+        stateClone[add] = transformsIndex.properties[add];
+      }
+
+      result.push(Object.assign({}, stateClone));
```
vitaliy-buchinchik, line 79: You can use push method in the end of the loop

vitaliy-buchinchik (changes requested):

src/transformStateWithClones.js
```diff
@@ -65,7 +65,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, transforms) {
-  // write code here
+  const result = [];
```
ArtemTopolskyi, line 68:
```suggestion
  const states = [];
```

src/transformStateWithClones.js
ArtemTopolskyi, line 79:
```suggestion
      states.push(Object.assign({}, stateClone));
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,35 @@
...
+
+      result.push(Object.assign({}, stateClone));
+    }
+
+    if (transformsIndex.operation === 'removeProperties') {
+      for (const remov of transformsIndex.properties) {
+        delete stateClone[remov];
+      }
+
+      result.push(Object.assign({}, stateClone));
```
ArtemTopolskyi, line 87:
```suggestion
      states.push(Object.assign({}, stateClone));
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,35 @@
...
+    }
+
+    if (transformsIndex.operation === 'clear') {
+      stateClone = {};
+      result.push({});
```
ArtemTopolskyi, line 92:
```suggestion
      states.push({});
```

src/transformStateWithClones.js
```diff
@@ -65,7 +65,35 @@
...
+    }
+  }
+
+  return result;
```
ArtemTopolskyi, line 96:
```suggestion
  return states;
```

ArtemTopolskyi (changes requested):

ArtemTopolskyi (approved):

PR #514, opened 2021-04-26
Solution

MinchukO:


src/transformStateWithClones.js
```diff
@@ -66,6 +66,31 @@
...
+        for (const key of properties) {
+          delete newState[key];
+        }
+        break;
+      case 'clear':
+        for (const key in newState) {
+          delete newState[key];
+        }
+        break;
+      default:
```
mykola-yarchuk, line 87: usually in switch case construction developers adds default case, it's a common issue that there is no default case, but it would be nice to have something like 
```
default: 
  // Do nothing
```

or 
```
default: 
```

mykola-yarchuk (approved):
Good job. Left some comments as advice

PR #600, opened 2021-06-27
Solution

IrynaKhyzhynska:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+
+    if (actions[act].type === 'clear') {
+      for (const j in temp) {
+        delete temp[j];
+      }
+    }
+    res.push({ ...temp });
+  }
+
+  return res;
```
vpolets, line 32: for ... of + switch will be better solution here

vpolets (approved):

PR #613, opened 2021-06-29
Solution

alina-alya:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,29 @@
...
+      for (const key of actions[i].keysToRemove) {
+        delete copyObj[key];
+      }
+    }
+
+    if (actions[i].type === 'clear') {
+      for (const key in copyObj) {
+        delete copyObj[key];
+      }
+    }
```
vkryvytskyy, line 29: use switch case instead of separate ifs

vkryvytskyy (approved):

PR #619, opened 2021-07-01

AlexandraBozhan:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const res = [];
+  const temp = { ...state };
+
+  for (const act in actions) {
```
vpolets, line 13:
```suggestion
  for (const act of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+    if (actions[act].type === 'addProperties') {
```
vpolets, line 14:
```suggestion
    if (act.type === 'addProperties') {
```

vpolets (approved):
switch case will be better solution here

PR #643, opened 2021-08-02
Solution

ivanchenko-kirill:


Moroz-Dmytro (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+  const stateCopy = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(stateCopy, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const keyToRemove of action.keysToRemove) {
+        if (stateCopy[keyToRemove]) {
```
varseniuk, line 20:
```suggestion
```
redundant

varseniuk (changes requested):
Let's use `switch case` here as well

varseniuk (approved):
👍

PR #654, opened 2021-08-03
Solution

denyschemeriakov:


lera-yaskevych (commented):
Looks good and clear! But I suggest you try using switch case in this task and the one that comes before this, since I've heard students saying they've got their works rejected for this reason, so why not foreseeing it now)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const changesOfState = [];
+  const clone = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(clone, action.extraData);
+      changesOfState.push({ ...clone });
```
varseniuk, line 16: Don't repeat yourself.
You can wright `changesOfState.push({ ...clone });` only once (in the end of the iterarion)

varseniuk (changes requested):
Let's not repeat yourself and use `switch...case` here

denyschemeriakov:
@lera-yaskevych  Thank you! =)

src/transformStateWithClones.js
denyschemeriakov: Done!

varseniuk (changes requested):
Add `default` statement

varseniuk (approved):

PR #656, opened 2021-08-04
Stateful Clones

andrii-hryhoruk:


alex-glotov67 (approved):
Great solution!

varseniuk (changes requested):
All is good, but don't forget about `default` case

PR #657, opened 2021-08-04
initial solution

sasha-morozov:


src/transformStateWithClones.js
```diff
@@ -7,7 +8,32 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete modifiedState[key];
+        }
+        break;
+
+      case 'clear':
+        for (const property in modifiedState) {
+          delete modifiedState[property];
+        }
```
DmytroMykoliv, line 29: Here you can use one line: modifiedState = {}, without using a loop )

Anna-Tkachenko (approved):
Looks good⭐

DmytroMykoliv (approved):
looks good for me ))

src/transformStateWithClones.js
sasha-morozov: Ok, will try your solution

sasha-morozov: Yeah, without loop it looks much more better;)

Ihor-Karpyn (approved):
LGTM 👌

PR #658, opened 2021-08-04
Stateful clones

UlaDutchak:


alex-glotov67 (approved):
Good job!

varseniuk (changes requested):
All is good, but don't forget to add `default` case

varseniuk (approved):

PR #661, opened 2021-08-04
Solution

Crystallord:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete objectStep[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in objectStep) {
+          delete objectStep[key];
+        }
```
andrii-hryhoruk, line 28: It's not necessary to create loop for it.
case 'clear':
        cloneState = {};

andrii-hryhoruk (approved):
Looks Great!

src/transformStateWithClones.js
Crystallord: It`s true! In the last task, it was impossible. But, if we use clone object it`s work! Thanks!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const collectionObjects = [];
+  let objectStep = { ...state };
```
varseniuk, line 11:
```suggestion
  let clonedState = { ...state };
```
as a variant. Not really clear what `objectStep` should mean

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+      case 'clear':
+        objectStep = {};
+        break;
+
+      default:
+        break;
+    }
+
+    collectionObjects.push(objectStep);
+    objectStep = Object.assign({}, objectStep);
```
varseniuk, line 34:
```suggestion
    collectionObjects.push({ ...objectStep });
```

varseniuk (changes requested):
Some small fixes left

src/transformStateWithClones.js
Crystallord: Give good variable`s name it`s art!)

varseniuk (approved):
🔥

PR #662, opened 2021-08-04
Stateful Clones Solution

ihor-shapoval:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateClone[key];
+        }
+
+        break;
+
+      case 'clear':
+        stateClone = {};
+    }
```
denyschemeriakov, line 29: add -break- to this statement

denyschemeriakov (changes requested):
Good work! But add -break- to the last statement. And some of our mentors require to add -default- statement.

src/transformStateWithClones.js
ihor-shapoval: Thanks, added

danheim (approved):

PR #664, opened 2021-08-04
Solution

Clonvhs:
v1.0

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateHistory = [];
+  const copyOfState = { ...state };
+
+  for (const action of actions) {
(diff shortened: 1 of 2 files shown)


AndreySerhovetc (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
-  // write code here
+  const stateHistory = [];
+  const copyOfState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const addPropertiesData in action.extraData) {
+          copyOfState[addPropertiesData] = action.extraData[addPropertiesData];
+        }
```
varseniuk, line 18:
```suggestion
```
Let's get rid of this loop and use Object.assign instead

varseniuk (changes requested):

varseniuk (approved):
👍

PR #665, opened 2021-08-04
Stateful-clones solved

Andrii-Pinchuk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete newState[key];
+        }
+
+        break;
+
+      case 'clear':
```
sasha-morozov, line 26: Here instead of the loop, you could have written "newState = {}"

sasha-morozov, line 26: here instead of the loop, you could have written: "newState = {}". It would be a bit shorter:)

sasha-morozov (approved):
Good job!

src/transformStateWithClones.js
Ihor-Karpyn: Agree with sasha

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        for (const property in newState) {
+          delete newState[property];
+        }
```
Ihor-Karpyn, line 29:
```suggestion
        newState = {};
```

Ihor-Karpyn (approved):
Pay attention to prev comment

Andrii-Pinchuk:
> Pay attention to prev comment

Already fixed it.

PR #672, opened 2021-08-04
Develop

roman-hado:


TarasHoliuk (changes requested):
It is better to use Switch, not If

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClone = { ...state };
+  const arrayForm = [];
+
+  for (let i = 0; i < actions.length; i++) {
```
varseniuk, line 13:
```suggestion
  for (const action of actions) {
```
As a variant

varseniuk (approved):
Good!
Left one suggestion

PR #675, opened 2021-08-04
solution

derelicttt:


Clonvhs (commented):
Looks Good for me

varseniuk (approved):
All is good, but you forgot to add `default` case to your `switch..case` construction

PR #728, opened 2021-09-15
solution

AVronskyi:


Yashkindzhus (approved):
Great! ✨

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  const newState = { ...state };
+  const arrayStates = [];
```
Anna-Tkachenko, line 11: Don't use names of data types in variable name
```suggestion
  const states = [];
```

Anna-Tkachenko (changes requested):
Rewrite it using switch/case

Anna-Tkachenko (approved):
Well Done!⭐️

PR #732, opened 2021-09-15
transformStateWithClone solution

VitalikStrog:


vlad-shulzhenko (approved):
Хорошая работа)

src/transformStateWithClones.js
```diff
@@ -6,8 +6,32 @@
  *
  * @return {Object[]}
  */
+// eslint-disable-next-line no-shadow
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arr = [];
+  const obj = { ...state };
```
Anna-Tkachenko, line 12: Let's create more meaningful names for variables instead of obj

Anna-Tkachenko (approved):
Good job!👍

PR #737, opened 2021-09-16
done

Yashkindzhus:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,31 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const clone = { ...state };
+  const resArr = [];
```
Anna-Tkachenko, line 12: Let's create more meaningful variable name instead of resArr

Anna-Tkachenko (approved):
Good job👍

PR #745, opened 2021-09-16
--done

PresidentComanch:


CouldSweeT (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newstate = { ...state };
+  const stateclones = [];
+
+  for (const action of actions) {
+    switch (action['type']) {
+      case 'addProperties':
+        Object.assign(newstate, action.extraData);
+        stateclones.push({ ...newstate });
```
Anna-Tkachenko, line 17: Don't repeat yourself, call push method only once at the end of the cycle iteration (after switch/case) instead of calling it in every case

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Well Done!👍

PR #751, opened 2021-09-17
solution

Tarasulko:


Bogdansgit (approved):

Bogdansgit:
good job!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+  const modifiedStates = actions.map(action => {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          currentState[key] = action.extraData[key];
+        }
+
+        const withAdd = Object.assign({}, currentState);
+
+        return withAdd;
```
Anna-Tkachenko, line 20: Don't repeat yourself: you can move this logic after switch/case instead of doing this in every case

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!👍

PR #765, opened 2021-10-05
solution

TRSKZK:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+  for (const action of actions) {
+    const { extraData, keysToRemove } = action;
+
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateDublicate, extraData);
+        break;
+
+      case 'removeProperties':
+        keysToRemove.forEach(prop => delete stateDublicate[prop]);
```
Igor-MateAcademy, line 22: Don't slice variables name: prop => property.
On my opinion, better to write code inside { } for cases.

Igor-MateAcademy (changes requested):

Igor-MateAcademy (approved):

PR #768, opened 2021-10-05
solution

podlesnyi-pavel:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
-  // write code here
+  const versionsOfState = [];
+  const copyState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const keyToAdd in action.extraData) {
+          copyState[keyToAdd] = action.extraData[keyToAdd];
+        }
```
Olexandr6, line 18: You can do it just using `Object.assign(...)`

Olexandr6 (changes requested):
In general, good solution👍 But your code needs some optimization.

VinnieJ-2k20 (approved):
Good job 👍

PR #820, opened 2021-11-09
stateful clones task solution

JamesSunderlanddd:


vlad-tuzenko (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,58 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let newObj;
```
illiarerun, line 10: So if a newObj is an array, than there is no need in using `let` keyword, `const` would be enough

illiarerun: Second, if it is an array, it's not only one new object, it's a lot of `states`, something like `stateHistory` would be great

src/transformStateWithClones.js
```diff
@@ -7,7 +7,58 @@
...
+  let count = 0;
```
illiarerun, line 11: Then this count, why do you need it, you can do a copy of a previous value of 'state', on each iteration form an array that you've declared above

illiarerun (changes requested):
Good start 🔥

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+
+  for (const obj of actions) {
+    if (obj.type === 'addProperties') {
+      Object.assign(changedState, obj.extraData);
+      stateHistory.push({ ...changedState });
+    }
+
+    if (obj.type === 'removeProperties') {
+      for (const key of obj.keysToRemove) {
+        if (changedState[key] !== undefined) {
```
illiarerun, line 21: It's be more convenient to use `key in` or `hasOwnProperty`

illiarerun (approved):
Good job 🔥  Also, check this one comment:

PR #821, opened 2021-11-09
solution

mariia-bila:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,23 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const transformHistory = [];
+  let transformStep = { ...state };
+
+  for (const element of actions) {
+    if (element.type === 'addProperties') {
+      Object.assign(transformStep, element.extraData);
```
dmtrhrytsak: You could use destructuring here, since you can reassing variable `transformStep`

```suggestion
      transformStep = { 
        ...transformStep,
        ...action.extraData,
      };
```

Looks more concise)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,23 @@
...
+    } else if (element.type === 'removeProperties') {
+      for (const removeKeys of element.keysToRemove) {
+        delete transformStep[removeKeys];
+      }
+    } else if (element.type === 'clear') {
+      transformStep = {};
+    }
+    transformHistory.push({ ...transformStep });
```
dmtrhrytsak, line 26: If I'm not wrong, you can push the object directly, no need for destrucuring

```suggestion
    transformHistory.push(transformStep);
```

dmtrhrytsak (approved):

illiarerun (approved):
LGTM 🔥

src/transformStateWithClones.js
illiarerun: Yeah you can push it directly, but what will happen if we change some values in this object afterwards?)

mariia-bila: @illiarerun nothing good 💯 we already had conversation with Dima about that 👍

PR #822, opened 2021-11-09
Transform state with clones solution

Vitaliy-0:


tmrsl (approved):
Looks good, just  use more informative names for variables

AlinaMatuschak (changes requested):

AlinaMatuschak (approved):

PR #824, opened 2021-11-09
Solution

msaikun:


msaikun:
Solution

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const result = [];
```
volodymyr-soltys97, line 11: ```result``` isn't informative. Try to choose another name for the variable.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+
+  for (const action of actions) {
```
volodymyr-soltys97, line 13:
```suggestion
  for (const action of actions) {
     const {type, extraData, keysToRemove} = action;
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+    if (action.type === 'addProperties') {
```
volodymyr-soltys97, line 14: try use switch case instead ```if / else if```

volodymyr-soltys97 (approved):
Good start, but you should use the switch in the task

AlinaMatuschak (changes requested):
fix @volodymyr-soltys97 comments 👆

src/transformStateWithClones.js
```diff
@@ -6,8 +6,38 @@
...
+  const arrResult = [];
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
+        for (const key in extraData) {
+          newState[key] = extraData[key];
+        }
```
AlinaMatuschak, line 22: you can use object assign

AlinaMatuschak (approved):

PR #828, opened 2021-11-09
Solution

Vitalii-Lytvyn01:
Just added some code to the solution of the previous task

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,49 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clone = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    const actionType = action.type;
+
+    switch (actionType) {
+      case 'addProperties':
+        addProperties(clone, action.extraData);
+        result.push({ ...clone });
+        break;
+      case 'removeProperties':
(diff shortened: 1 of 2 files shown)


sviatoslav-myhal (commented):
strange decision to do everything with funcs) but looks great!

sviatoslav-myhal (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,49 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clone = { ...state };
+  const result = [];
```
illiarerun, line 11: Don't use names like result in your code, it's a bad practice, something like stateHistory would be a perfect fit

src/transformStateWithClones.js
```diff
@@ -7,7 +7,49 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clone = { ...state };
```
illiarerun, line 10: Be more specific with names
```suggestion
  const stateClone = { ...state };
```

illiarerun (changes requested):
Good job 🔥  Make sure to check comments:

illiarerun (approved):
Oops

PR #831, opened 2021-11-09
solution

Val-space:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+
+      case 'removeProperties':
+        actions[index].keysToRemove.forEach(key => {
+          delete prevStates[index][key];
+        });
+        break;
+
+      case 'clear':
+        Object.keys(prevStates[index])
+          .forEach(key => delete prevStates[index][key]);
```
dmtrhrytsak, line 30: Clear means deleting every key of an object, right? Then you could just make the current state reference an empty object, and the previous one would be simply deleted by the garbage collector of a javascript engine)

`prevStates[index] = {}`

dmtrhrytsak (approved):

src/transformStateWithClones.js
Val-space: oh yeah, I was going to but forgot

illiarerun (approved):
LGTM 🔥

PR #832, opened 2021-11-09
Solution

olya-shyian:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        for (const key of keysToRemove) {
+          delete newState[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in newState) {
+          delete newState[key];
+        }
+    }
```
dmtrhrytsak, line 31: Clear means deleting every key of an object, right? Then you could just make the current state reference an empty object, and the previous one would be simply deleted by the garbage collector of a javascript engine)

`newState = {}`

dmtrhrytsak (approved):

AlinaMatuschak (approved):

PR #837, opened 2021-11-09
Solution

Veronika-Korona:


alinaShchavinska (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let cloneState = { ...state };
+  const result = [];
```
illiarerun, line 11: Don't use names like result in your code, it's a bad practice, something like stateHistory would be a perfect fit

illiarerun (approved):
Good job 🔥  Check this one comment:

PR #840, opened 2021-11-09
solution

MagicBox15:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copy = { ...state };
+  const result = [];
+
+  for (const obj of actions) {
```
alinaShchavinska, line 13: maybe, better name (const action of  actions)

alinaShchavinska (approved):

src/transformStateWithClones.js
MagicBox15: hmm... agree

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copy = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case
+        'addProperties':
```
illiarerun, line 16: Again, don't put the value of a case on a newline

illiarerun (approved):
Good job 🔥

PR #841, opened 2021-11-09
Solution

SerhiiKirik:


DudarDaryna (approved):
🔥

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let copy = { ...state };
+  const result = [];
+
+  for (const obj of actions) {
```
illiarerun, line 13: If we iterate through an array of `actions` then the element of this array would be called `action`, right?

illiarerun (approved):
Good job 🔥

PR #842, opened 2021-11-09
Clone solution

skachkov23:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let copyObject = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```
volodymyr-soltys97, line 14: I think it will be better.
```suggestion
  for (const action of actions) {
   const {type, extraData, keysToRemove} = action;
   
   switch (type) {
```

volodymyr-soltys97 (approved):

AlinaMatuschak (approved):
🤘

PR #850, opened 2021-11-10
Solution

dlugash:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
+          const key = keysToRemove[i];
+
+          delete newState[key];
+        }
+        break;
+      }
+
+      case 'clear' : {
+        newState = {};
+        break;
```
sviatoslav-myhal, line 39: there is no need to write "break" in the last case

sviatoslav-myhal (approved):
good job!

dlugash:
> good job!

agree, thanks!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clones = [];
+  let lastState = state;
+
+  for (const action of actions) {
+    let newState;
```
illiarerun, line 14: Too much variables, here. You can work this out with only lastState variable

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+        break;
+      }
+
+      case 'removeProperties' : {
+        const keysToRemove = action.keysToRemove;
+
+        newState = Object.assign({}, lastState);
+
+        for (let i = 0; i < keysToRemove.length; i++) {
+          const key = keysToRemove[i];
```
illiarerun, line 30: Use for ... of loop here, no need to create a new variable every time

illiarerun (changes requested):
Good job 🔥  Let's work on fixes:

src/transformStateWithClones.js
dlugash: you mean rewriting lastState every time ?

illiarerun: Or just changing object 😉

illiarerun (changes requested):

src/transformStateWithClones.js
aionova: agree

aionova (approved):
Nice!

PR #874, opened 2021-12-14
Solution

destroythemoon:
🥲

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const previousVersions = [];
+  const stateCopy = { ...state };
+
+  for (const properties of actions) {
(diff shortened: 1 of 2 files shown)


ArturErmolenko (commented):
[CODE KNOWLEDGE] - use object desructuring for getting values from object.

ArturErmolenko (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      for (const i of properties.keysToRemove) {
+        delete stateCopy[i];
+      }
+    }
+
+    if (properties.type === 'clear') {
+      for (const i in stateCopy) {
+        delete stateCopy[i];
+      }
+    }
```
Denchepornyuk, line 28: use switch case instead of three `if` conditions

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #887, opened 2021-12-14
solution

Viktor9r:


Sesile26 (approved):
Добав default в кінці свіча

src/transformStateWithClones.js
```diff
@@ -8,6 +8,31 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const result = [];
```
Olexandr6, line 11:
```suggestion
  const previousStates = [];
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,31 @@
...
+  const copy = { ...state };
```
Olexandr6, line 12:
```suggestion
  const currentState = { ...state };
```

Olexandr6 (approved):
Good job 👍  

Check @Sesile26 comment

PR #902, opened 2021-12-15
clone solution

OleksandrOstrovskyi:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
+
+      default:
+    }
```
orest-mykhailyshyn, line 34: default is empty. Why?

orest-mykhailyshyn (approved):
Maybe
default:
        return null;
or something similar)

src/transformStateWithClones.js
Olexandr6: Agree, you should write some logic for default

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const stateCopiesArr = [];
```
Olexandr6, line 11:
```suggestion
  const stateCopies = [];
```

Olexandr6 (approved):

PR #921, opened 2022-01-10
Solution

VovaPain741:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const addProperties = 'addProperties';
+  const removeProperties = 'removeProperties';
+  const clear = 'clear';
```
aliceblunt8, line 12: no need for these variables, using just string in this case it's enough

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+  let newObj = { ...state };
+  const newArr = [];
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === addProperties) {
```
aliceblunt8, line 17: use switch case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+    if (actions[i].type === removeProperties) {
+      for (const key of actions[i].keysToRemove) {
+        delete newObj[key];
+      }
+      newArr.push(Object.assign({}, newObj));
+    }
+
+    if (actions[i].type === clear) {
+      newObj = {};
+      newArr.push({});
```
aliceblunt8, line 31: push only once after all blocks

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #929, opened 2022-01-18
Solution

Oleksii-Pinchuk:


pogrebnyakp (approved):

src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const statesArray = [];
```
Anna-Tkachenko, line 11: don't use data type names in variable names
```suggestion
  const states = [];
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
+
+      default:;
```
Anna-Tkachenko, line 34:
```suggestion
      default:
      	break;
```

Anna-Tkachenko (approved):

PR #966, opened 2022-01-30
Solution

KovalVladuslav:


src/transformStateWithClones.js
```diff
@@ -64,7 +64,45 @@
...
+      }
+
+      case 'removeProperties': {
+        const obj = { ...acc };
+
+        properties.forEach(key => {
+          delete obj[key];
+        });
+
+        result.push(obj);
```
Ihor-Karpyn, line 94: `result.push(obj);` don't repeat yourself

Ihor-Karpyn (changes requested):

PR #974, opened 2022-02-08
solution for #transformStateWithClones

jstmpelowycz:


krolich500grn (approved):
good job!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
+        temp = {};
+        break;
+      case 'removeProperties':
+        Object
+          .values(action.keysToRemove)
+          .forEach(K => delete temp[K]);
+        break;
```
Ihor-Karpyn, line 22: for in

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+
+      default:
+        for (const [K, V] of Object.entries(action.extraData)) {
+          temp[K] = V;
+        }
```
Ihor-Karpyn, line 27: use for in, it's not default it case 'add'

Ihor-Karpyn (changes requested):

Ihor-Karpyn (approved):

PR #981, opened 2022-02-09
solution

Marta1993-droid:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(copy, action.extraData);
+        break;
+
+      case 'removeProperties':
+
+        for (const key of action.keysToRemove) {
```
SanyaBratashchuk, line 21:
```suggestion
        for (const key of action.keysToRemove) {
```
"keysToRemove is an array. You shouldn't name a variable as key, better to name it as element or smth like that. But it is just a suggestion

SanyaBratashchuk (approved):

Ihor-Karpyn (approved):

PR #996, opened 2022-02-11
solution

AnnSerdechna:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arr = [];
+  const newState = { ...state };
```
solaryasha, line 11: Let's use better naming for variables

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+
+  for (const value of actions) {
```
solaryasha, line 13:
```suggestion
  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+    if (value['type'] === 'addProperties') {
```
solaryasha, line 14: Let's practice `switch` statement

solaryasha (changes requested):

PR #997, opened 2022-02-11
new solution

Danylo-Kotliarenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newArr = [];
+  let newBox = { ...state };
```
barannikovArtem, line 11: I think u may to name it better

barannikovArtem (approved):
I like it  C:

src/transformStateWithClones.js
solaryasha: Agree with Artem, let's use better naming for variables

solaryasha (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newArray = [];
+  let arrayResult = { ...state };
```
solaryasha, line 11: https://mateacademy.slack.com/archives/C02T596SHUZ/p1644920876764629

solaryasha (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let joinedState = { ...state };
```
solaryasha, line 11:
```suggestion
  let copiedState = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
solaryasha, line 10:
```suggestion
  const transformedStates = [];
```

solaryasha (approved):

PR #1004, opened 2022-02-20
Solution

derenivska:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,45 @@
...
+      }
+
+      case 'clear': {
+        for (const key in newObj) {
+          delete newObj[key];
+        }
+
+        const cloneObject = { ...newObj };
+
+        clones.push(cloneObject);
```
aliceblunt8, line 43: push at the end of every iteration, not inside each case

src/transformStateWithClones.js
```diff
@@ -8,6 +8,45 @@
...
+        break;
+      }
+
+      case 'clear': {
+        for (const key in newObj) {
+          delete newObj[key];
+        }
+
+        const cloneObject = { ...newObj };
```
aliceblunt8, line 41: also no need to do this, just push a copy to clones
```
clones.push({...newObj});
```

aliceblunt8 (changes requested):

PR #1007, opened 2022-02-22
task is done

alexsom64:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const objects = [];
+  let newObj = { ...state };
+
+  for (const obj of actions) {
+    switch (obj.type) {
+      case 'addProperties':
+        newObj = { ...newObj };
```
vitaliy-martynyuk, line 16: there is no need to copy object to itself

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+          if (key in newObj) {
+            delete newObj[key];
+          }
+        }
+        objects.push(newObj);
+        break;
+      case 'clear':
+        newObj = { ...newObj };
+        Object.keys(newObj).forEach(key => delete newObj[key]);
+        objects.push(newObj);
```
vitaliy-martynyuk, line 33:
```suggestion
        newObj = {};
```

vitaliy-martynyuk (changes requested):

src/transformStateWithClones.js
alexsom64: it changed the object without coping
last 2 tests failed without this line

src/transformStateWithClones.js
alexsom64: fixed

src/transformStateWithClones.js
vitaliy-martynyuk: in this case, you can delete all 3 `objects.push(newObj);` and create `objects.push({ ...newObj });` before the last `return` statement of this function

PR #1014, opened 2022-03-11
Solution

OlexandrIstomin:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,35 @@
  *
  * @return {Object[]}
  */
-function transformStateWithClones(state, actions) {
-  // write code here
+function transformStateWithClones(object, array) {
```
Bohdan-Maksymiuk, line 9: Why do you rename variables? You should think about obvious naming. Everyone should understand your code.

Bohdan-Maksymiuk (dismissed):
Good code logic, but don't use obj, arr, i, newArr and etc. names for your variables. Naming is important!

src/transformStateWithClones.js
OlexandrIstomin: Скорее всего, случайно перенес название функции без параметров

varseniuk (approved):

PR #1027, opened 2022-03-28
Solution

yaremenko-maksym:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const currentState = { ...state };
+  const stateChangesArr = [];
+
+  for (let i = 0; i < actions.length; i++) {
+    switch (true) {
```
OksanaTyshchenko, line 15: you can write switch (actions[i].type) and case 'addProperties', because expression in switch is checking on strict comparison to value in case

src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
...
+      case actions[i].type === 'addProperties':
+        const { extraData } = actions[i];
+
+        Object.assign(currentState, extraData);
+        stateChangesArr[i] = Object.assign({}, currentState);
```
OksanaTyshchenko, line 20: this expression repeats three times maybe it is better to use if? you will able to write this statement only once at the end of loop

src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
...
+        break;
+
+      case actions[i].type === 'removeProperties':
+        for (const key of actions[i].keysToRemove) {
+          delete currentState[key];
+        }
+        stateChangesArr[i] = Object.assign({}, currentState);
```
OksanaTyshchenko, line 27: or just stateChangesArr[i] = { ...currentState };

OksanaTyshchenko (changes requested):
Good job, but switch/case need to little fix

src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const currentState = { ...state };
+  const stateChangesArr = [];
```
Anna-Tkachenko, line 12: Don't use data type names in variable names
```suggestion
  const stateChanges = [];
```

Anna-Tkachenko (approved):
Good job 👍

PR #1028, opened 2022-03-29
solved

VladSkoromnyi:


vlasiuk-anatolii (approved):
Good job!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const firstState = { ...state };
+  const changeState = [];
+
+  function assigned(index, to, from) {
+    changeState[index] = Object.assign(to, from);
+  }
```
Anna-Tkachenko, line 15: This function is redundant, you can use destructuring for creating object copy and add it to array
array.push(...object)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
+      for (const key in actions[i].extraData) {
+        firstState[key] = actions[i].extraData[key];
+      }
+      assigned(i, {}, firstState);
```
Anna-Tkachenko, line 22: Don't repeat yourself: you use this functionality 3 times in your code, move it at the end of cycle

Anna-Tkachenko (changes requested):
Use switch / case if you need to do some logic depending on the value of some variable

Anna-Tkachenko (approved):
Good job! 👍

PR #1033, opened 2022-03-30
transformStateWithClones function was created

OksanaTyshchenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const clone = { ...state };
+
+  for (const item of actions) {
+    if (item.type === 'addProperties') {
```
yaremenko-maksym, line 14: `switch / case` instead of `if / else`.

yaremenko-maksym (changes requested):
Good job👍. Might use some minor changes.

src/transformStateWithClones.js
OksanaTyshchenko: Thank you for review, but I like using switch/case only with return and when there are many conditions) I don't like break , I am afraid forgetting them)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const clone = { ...state };
```
Anna-Tkachenko, line 11:
```suggestion
  const stateCopy = { ...state };
```

Anna-Tkachenko (changes requested):
Good job!
Just rewrite your solution using switch/case which is better in cases with limited amount of conditions (check checklist)

Anna-Tkachenko (approved):
Same here with default case, don't forget to add it

PR #1034, opened 2022-03-30
state was added

HelenPetrovska:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,34 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const cloneArr = [];
```
Anna-Tkachenko, line 11: Don't use name of data types in variable names

Anna-Tkachenko (approved):
Good job! 👍

PR #1035, opened 2022-03-31
Solution

Vakolyyk:


VETALEK (approved):
Works fine, i think

Anna-Tkachenko (approved):
Good job!
Just use switch / case if you need to do some logic depending on the value of some variable

PR #1037, opened 2022-03-31
Solution

strukk311:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const result = [];
+  const steteItems = { ...state };
+
+  for (const action of actions) {
+    switch (true) {
+      case action.type === 'addProperties':
```
KostiaO, line 16: destruct objects: const { type, extradata } = action;

KostiaO (changes requested):
use destruct

src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const result = [];
+  const steteItems = { ...state };
```
Anna-Tkachenko, line 12:
```suggestion
  const stateItems = { ...state };
```

Anna-Tkachenko (approved):
Good job!

PR #1038, opened 2022-03-31
Clonestate solution

vlasiuk-anatolii:


VladSkoromnyi (approved):
Good job! but maybe you can add repeat part of code in function

Anna-Tkachenko (approved):
Good job!
Just pay attention to variable names
Will be better to create more meaningful variable name instead of res

PR #1040, opened 2022-03-31
Create solution

ivan-kadykalo:


HelenPetrovska (changes requested):

Anna-Tkachenko (approved):
Good job!
Just use more meaningful variable names instead of arrContainer and tempObj

PR #1041, opened 2022-03-31
Solution

lem47:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const result = [];
```
Anastasiia-Sharkova, line 11: pick a better name for variable :)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+
+  for (const action of actions) {
+    switch (true) {
```
Anastasiia-Sharkova, line 14: [CODE STYLE]: switch/case should always have default case for error handling.

Anastasiia-Sharkova (approved):

PR #1042, opened 2022-03-31
solution

kurochka-vladyslav:


Anastasiia-Sharkova (approved):

Anna-Tkachenko (approved):
Good job!
Pay attention to variable naming, it should be meaningful

PR #1045, opened 2022-03-31
solution

OleksandrNi:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,29 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const clone = Object.assign({}, state);
+  const result = [];
```
Anastasiia-Sharkova, line 12: pick a better names for variables :)

src/transformStateWithClones.js
```diff
@@ -8,6 +8,29 @@
...
+
+  for (const key of actions) {
+    if (key.type === 'addProperties') {
```
Anastasiia-Sharkova, line 15: [CODE STYLE]: Use switch statement if you have limited amount of conditions.

Anastasiia-Sharkova (approved):

Anna-Tkachenko (changes requested):
Good job!
Just rewrite your code using switch / case statement
This way will be better when you need to do some logic depending on the value of some variable

src/transformStateWithClones.js
danylo-onishchuk: its good point

danylo-onishchuk (approved):

PR #1053, opened 2022-04-01
stateful-clones

Nikitiliok:


SeKaM0 (approved):
Good job

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const likeState = { ...state };
```
Anna-Tkachenko, line 10: What is likeState?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+  const result = [];
+
+  for (let i = 0; i < actions.length; i++) {
+    switch (actions[i].type) {
+      case 'addProperties':
+        for (const key in actions[i].extraData) {
+          likeState[key] = actions[i].extraData[key];
+        }
+        copyObj = { ...likeState };
+        result.push(copyObj);
```
Anna-Tkachenko, line 21: Don't repeat yourself: you use the same code 3 times, move it to the end of the cycle

Anna-Tkachenko (changes requested):
Don't forget to add default case for switch / case statement

PR #1066, opened 2022-04-02
Solution

EvgenBabin1:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
```
VETALEK, line 28: code would be more productive if you just return `{}` instead of deleting everything

VETALEK (approved):
Good job

danylo-onishchuk (approved):

PR #1070, opened 2022-04-04
Task done

andriyshvets:


andriyshvets:
Task*

Kateryna03 (commented):
good job! maybe by default return smth? or throw an error as  mentor in his solution )

danylo-onishchuk (approved):

PR #1083, opened 2022-04-13
Solution

oleg-zolkiewskyi:


marina-tilniak (changes requested):
add default value for switch case

ArturErmolenko (approved):
Well done 🔥

PR #1085, opened 2022-04-15
Solution

TanyaPanfyorova:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultArr = [];
+  let transState = { ...state };
+
+  for (const obj of actions) {
```
Oleksandra-Pasenchenko, line 13:
```suggestion
for (const action of actions) {
```

Oleksandra-Pasenchenko (approved):
Good job, but have a look at the comment: it's better to name variables that you iterate over as singular, like this:
action of actions
state of states

src/transformStateWithClones.js
TanyaPanfyorova: Got it! Fixed

Oleksandra-Pasenchenko (approved):

PR #1087, opened 2022-04-18
Solution

oles-partyka:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arr = [];
+  const newObj = Object.assign({}, state);
```
Bohdan-Maksymiuk, line 11: don't use such naming (arr , obj)
you can use `stateVersions` and `stateCopy` or smth else semantic

Bohdan-Maksymiuk (changes requested):
good job but naming is important

Bohdan-Maksymiuk (approved):

PR #1090, opened 2022-04-20
Task completed

andreisucman:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clone = { ...state };
+  const result = [];
+
+  for (let i = 0; i < actions.length; i++) {
```
59MrRobot, line 13: Again here I would suggest using for...in loop

59MrRobot (approved):
Good work Andrei, just left one comment

VinnieJ-2k20 (changes requested):
Same as previous task. Rewrite using `switch/case`.

VinnieJ-2k20 (approved):
Good job 👍

PR #1091, opened 2022-04-20
solve statefulClones

tsybulko-natalya:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const result = [];
```
```suggestion
  const stateVersions = [];
```
not just result

Bohdan-Maksymiuk (approved):
Good job

PR #1093, opened 2022-04-21
solution

LukaszGryziewicz:


andreisucman (changes requested):
Good work but use switch instead of the if blocks.

andreisucman (approved):
Good job.

VinnieJ-2k20 (approved):
Good job 👍

PR #1099, opened 2022-04-25
solution

huberciak:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        }
+        break;
+
+      default:
+        throw new Error('Invalid paramater');
+    }
+
+    const updatedState = { ...objectCopy };
+
+    statesArray.push(updatedState);
```
59MrRobot, line 37: I think here you can just push `{ ...objectCopy }` , there's no need for the `updatedState` variable

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const objectCopy = { ...state };
+  const statesArray = [];
+
+  for (const action of actions) {
```
59MrRobot, line 13: You should use object destructuring here according to the checklist point #5

59MrRobot (changes requested):
Good job, Hubert, just a couple suggestions

59MrRobot (approved):
Good job!👍

meforoma (approved):
yeah 😎 👍

PR #1100, opened 2022-04-26
solutionI

EnjSadman:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arr = [];
+
+  for (let i = 0; i < actions.length; i++) {
+    if (i === 0) {
+      arr[i] = { ...state };
+    } else {
+      arr[i] = { ...arr[i - 1] };
+    }
```
Denchepornyuk, line 17: you don't need these, create an empty array outside the loop and push to this array result of each loop iteration

Denchepornyuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+
+  const copy = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
+    const currentAction = actions[i];
+
+    switch (currentAction.type) {
+      case 'addProperties':
+        Object.assign(copy, currentAction.extraData);
+        arr.push({ ...copy });
```
Denchepornyuk, line 20: move arr.push outside the `switch` and remove push from each case

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1102, opened 2022-04-26
solution

alexandra-protyanova:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+      }
+    }
+
+    if (el.type === 'removeProperties') {
+      for (const del of el.keysToRemove) {
+        delete obj[del];
+      }
+    }
+    res.push({ ...obj });
+  }
```
Denchepornyuk, line 30: everything is perfect but use `switch case`

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1103, opened 2022-04-26
clones

elkhayate:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
...
+        break;
+
+      case 'clear':
+        for (const key in output) {
+          delete output[key];
+        }
+        break;
+
+      default:
+        return null;
```
meforoma, line 33: default case often used to catch errors, so it's better to provide some descriptive outcome
at least `return 'not supported action provided'`

meforoma (approved):

PR #1104, opened 2022-04-26
solution

AnastasiyaSopotnytska:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const cloneOfState = { ...state };
+  const resultArray = [];
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'clear') {
+      Object.keys(cloneOfState).forEach(key => delete cloneOfState[key]);
```
Denchepornyuk, line 16: nope, use `for in` loop instead of `Object.keys` and `forEach`

src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
...
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      for (let y = 0; y < actions[i].keysToRemove.length; y++) {
+        if (cloneOfState.hasOwnProperty(actions[i].keysToRemove[y])) {
+          delete cloneOfState[actions[i].keysToRemove[y]];
+        }
+      }
```
Denchepornyuk, line 24: use `for in` loop  and remove `if` 

If the key doesn't exist in the object and you are trying to remove this key, JS will skip it without any error
image

src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
...
+    }
+
+    if (actions[i].type === 'addProperties') {
+      for (const a in actions[i].extraData) {
+        cloneOfState[a] = actions[i].extraData[a];
+      }
```
Denchepornyuk, line 30: remove loop and use `Object.assign` instead

Denchepornyuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -8,6 +8,30 @@
...
+      for (const y in actions[i].keysToRemove) {
+        delete cloneOfState[actions[i].keysToRemove[y]];
+      }
+    }
+
+    if (actions[i].type === 'addProperties') {
+      Object.assign(cloneOfState, actions[i].extraData);
+    }
+
+    resultArray.push({ ...cloneOfState });
```
Denchepornyuk, line 31: use `switch case` instead of `if`

Denchepornyuk (approved):

Denchepornyuk (changes requested):
fix comment above

Denchepornyuk (approved):

PR #1105, opened 2022-04-26
solution

vlad-demchuk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  const result = [];
+  const newOb = { ...state };
+
+  for (const i of actions) {
+    if (i.type === 'addProperties') {
```
alexandra-protyanova, line 14: you should use here switch, bro :(

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  const result = [];
+  const newOb = { ...state };
+
+  for (const i of actions) {
```
alexandra-protyanova, line 13: const's name a bit unsuitable

alexandra-protyanova (changes requested):

PR #1108, opened 2022-04-27
solution

SavMary:


Andriyk0 (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const cloneState = { ...state };
+  const arr = [];
+
+  for (const key in actions) {
+    switch (actions[key].type) {
+      case 'addProperties':
+        Object.assign(cloneState, actions[key].extraData);
+        arr.push({ ...cloneState });
```
Denchepornyuk, line 17: move `arr.push` outside the `switch` and remove push from each case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const cloneState = { ...state };
+  const arr = [];
+
+  for (const key in actions) {
+    switch (actions[key].type) {
```
Denchepornyuk, line 14: the default should be in each `switch case`, return the default block

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1114, opened 2022-04-27
solution stateful clone

kitaminary:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,29 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key2 of obj.keysToRemove) {
+          delete newState[key2];
+        }
+        break;
+
+      default:
+        newState = {};
```
iwanttosmile4u, line 27: I think it isn`t correct write this operation in default. It will be better if you use one more case with 'clear', use loop (key in obj) and add empty object into copied object inside this loop.

Denchepornyuk (approved):

PR #1115, opened 2022-04-27
First_commit_StateWithClones

igoree88:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
...
+  const newState = { ...state };
+
+  for (const value of actions) {
+    if (value.type === 'addProperties') {
+      Object.assign(newState, value.extraData);
+    }
+
+    if (value.type === 'removeProperties') {
+      for (const value2 of value.keysToRemove) {
+        if (newState.hasOwnProperty(value2)) {
```
rostyslav-pyliak, line 21: It should work without "if"

rostyslav-pyliak (approved):
Looks like everything OK

src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
...
+          delete newState[value2];
+        }
+      }
+    }
+
+    if (value.type === 'clear') {
+      for (const key in newState) {
```
Denchepornyuk, line 28: use `switch case` instead of `if`

src/transformStateWithClones.js
Denchepornyuk, line 21: you can remove `if`

Denchepornyuk (changes requested):

src/transformStateWithClones.js
igoree88: Fixed.

igoree88: Fixed.

Denchepornyuk (approved):

PR #1116, opened 2022-04-27
Develop

PRO-GRAM-MER:


src/transformStateWithClones.js
```diff
@@ -1,13 +1,26 @@
...
- *
- * @return {Object[]}
- */
 function transformStateWithClones(state, actions) {
   // write code here
+  const duplicate = { ...state };
+  const array = [];
+
+  for (const action in actions) {
+    if (actions[action]['type'] === 'addProperties') {
```
meforoma, line 9: the logic is fine, but code can be much more graceful
use object de-structuring and a switch - case statement with default case
```

switch (type) {
  case '...':
    ...
    break;
  case '...':
    ...
```

src/transformStateWithClones.js
```diff
@@ -1,13 +1,26 @@
...
- * @param {Object[]} actions
- *
- * @return {Object[]}
- */
 function transformStateWithClones(state, actions) {
   // write code here
+  const duplicate = { ...state };
+  const array = [];
+
+  for (const action in actions) {
```
meforoma, line 8: and wait, actions - is it an object?
i guess it's array - then you should use `for ( ... of ...)`

meforoma (changes requested):

PR #1117, opened 2022-04-27
solution

valerii-derkach:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clone = { ...state };
+  const result = [];
+
+  for (const object of actions) {
+    if (object.type === 'removeProperties') {
```
iwanttosmile4u, line 14: Use switch instead of if ;)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+      for (const values of object.keysToRemove) {
+        delete clone[values];
+      }
+    }
+
+    if (object.type === 'addProperties') {
+      for (const pair in object.extraData) {
```
iwanttosmile4u, line 21: You can use Object.assign() and do your operation in one row.

Denchepornyuk (approved):

PR #1119, opened 2022-04-27
Solution

DNK-forexample:


src/transformStateWithClones.js
```diff
@@ -1,13 +1,36 @@
...
+      for (const key of obj.keysToRemove) {
+        delete clone[key];
+      }
+    }
+
+    if (obj.type === 'clear') {
+      for (const prop in clone) {
+        delete clone[prop];
+      }
+    }
```
Denchepornyuk, line 28: use `switch case` instead of `if`

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1120, opened 2022-04-27
Solution

RomekIvantsiv:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (stateCopy.hasOwnProperty(key)) {
+            delete stateCopy[key];
+          }
+        }
+        break;
+
+      case 'clear':
+        if (action.type === 'clear') {
```
JustDan17, line 28: Remove this condition. It's excessive. You already set such condition with `case 'clear':`.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (stateCopy.hasOwnProperty(key)) {
```
JustDan17, line 21: Remove this condition. It's unnecessary. It does not change the result but makes code longer)

JustDan17 (changes requested):

Denchepornyuk (approved):

PR #1122, opened 2022-04-28
Solution

JustDan17:


RomekIvantsiv (approved):

src/transformStateWithClones.js
```diff
@@ -1,13 +1,37 @@
...
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const obj = { ...state };
+
+  for (const act of actions) {
+    switch (true) {
+      case act.type === 'addProperties':
```
Denchepornyuk, line 15: use `act.type` instead of `true` for switch and remove `===` from cases

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1125, opened 2022-04-28
with clones - is solved

enrikadej:


dellachaise (approved):
Good job!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      }
+    }
+
+    if (object.type === 'clear') {
+      for (const char in newState) {
+        delete newState[char];
+      }
+    }
+
+    result.push({ ...newState });
```
Denchepornyuk, line 30: use `switch case` instead of `if`

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1127, opened 2022-04-28
Inception?

proznerka:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete copiedState[key];
+        }
+        break;
+
+      case 'clear':
+        copiedState = {};
+        break;
```
efdalyalcin, line 27: you just need to add the default case for the switch.

efdalyalcin (changes requested):
Good work 👍  you just need the `default: break;` which was Roman's comment for me :)

PR #1130, opened 2022-04-28
 js_stateful-clones

anastasiia-nikita:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+
+      case 'clear':
+        for (const clear in copyObj) {
+          delete copyObj[clear];
+        }
+        break;
+
+      default:
+        break;
```
rostyslav-pyliak, line 32: Default and break are redundant here

Denchepornyuk (approved):

PR #1133, opened 2022-04-29
Transform state with clones

Alyona-Pisotska:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+        for (const key of action.keysToRemove) {
+          delete newState[key];
+        }
+        break;
 
+      case 'clear':
+        for (const key in newState) {
+          delete newState[key];
+        }
+    }
```
ArturErmolenko, line 29: add default case

ArturErmolenko (changes requested):

ArturErmolenko (approved):

PR #1135, opened 2022-04-30
solution

vitaliy-kovbasinskyi:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const obj = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    switch (true) {
```
valerii-derkach, line 14: use switch (action.type) {
case 'addProperties': ...
instead of switch (true) {
case action.type === 'addProperties':...
because you repeat the code 'action.type === '  3 times

valerii-derkach (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+        for (const del of action.keysToRemove) {
+          delete obj[del];
+        }
+        break;
+
+      case 'clear':
+        for (const key in obj) {
+          delete obj[key];
+        }
+        break;
```

Anna-Tkachenko (approved):

PR #1137, opened 2022-04-30
transform state

ISRLv2021:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,25 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const res = [];
+  const newState = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
```
valerii-derkach, line 14: use switch (actions[i].type) 
case 'addProperties' ... instead of if/else 
because you repeat the same code 3 times

valerii-derkach (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      case 'removeProperties':
+        for (const key of actions[j].keysToRemove) {
+          delete newState[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key1 in newState) {
+          delete newState[key1];
+        }
```
Anna-Tkachenko, line 28: Where is break?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+    }
```

Anna-Tkachenko (changes requested):

Anna-Tkachenko (changes requested):
You didn't fix my comments

Anna-Tkachenko (approved):

PR #1138, opened 2022-04-30
Solution

AndriySimyachenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+
+  let newState = { ...state };
```
valerii-derkach, line 11: use const instead let when you declair an object

valerii-derkach (approved):
well done, just one improvement - change let to const when you assigth new object

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+  const arr = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(newState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const act of action.keysToRemove) {
```
Anna-Tkachenko, line 21: what is act?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+          delete newState[act];
+        };
+        break;
+
+      default:
```
Anna-Tkachenko, line 26: It's not default case, it's case when action.type = 'clear'

Anna-Tkachenko (changes requested):

src/transformStateWithClones.js
Anna-Tkachenko: Create understandable variable names, you didn't fix my comment
What is act? I

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+        Object.assign(newState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const act of action.keysToRemove) {
+          delete newState[act];
+        };
+        break;
+
+      case 'clear':
```
Anna-Tkachenko, line 25: Don't forget about default case
It should be in all switch / case statement
You can just add break operator here

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #1139, opened 2022-04-30
solution

Ihor-Pavlenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+  const result = [];
+
+  actions.forEach(elem => {
+    let copy = {};
+
+    if (result.length === 0) {
+      copy = { ...state };
+    } else {
+      copy = { ...result[result.length - 1] };
+    }
```
aliceblunt8, line 19: make a copy outside the loop
just {... state}, it's enough

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+    if (elem.type === 'addProperties') {
+      for (const key in elem.extraData) {
+        copy[key] = elem.extraData[key];
+      }
+    } else if (elem.type === 'removeProperties') {
+      elem.keysToRemove.forEach(item => {
+        delete copy[item];
+      });
+    } else if (elem.type === 'clear') {
```
aliceblunt8, line 29: use switch case

aliceblunt8 (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let copy = { ...state };
+
+  actions.forEach(elem => {
+    if (result.length !== 0) {
+      copy = { ...result[result.length - 1] };
+    }
```
aliceblunt8, line 16: why do you need this? remove

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        break;
+
+      case 'clear':
+        for (const key in copy) {
+          delete copy[key];
+        }
+        break;
+    }
+
+    result.push(copy);
```
aliceblunt8, line 37:
```suggestion
    result.push({...copy});
```

aliceblunt8 (changes requested):

src/transformStateWithClones.js
Ihor-Pavlenko: Each element of the resulting array has to represent the state produced by the next operation. Otherwise it will fail.

ArturErmolenko (approved):

PR #1140, opened 2022-04-30
My Solution

Yaroslav-Kononenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,25 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const res = [];
+  const copy = Object.assign({}, state);
+
+  for (const partAction of actions) {
+    if (partAction.type === 'addProperties') {
```
valerii-derkach, line 14: use switch/case instead if/else because you repeat the same code 3 times

valerii-derkach (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const res = [];
+  const copy = Object.assign({}, state);
```
Anna-Tkachenko, line 11: Will be better to name it as stateCopy

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        for (const key of partAction.keysToRemove) {
+          delete copy[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in copy) {
+          delete copy[key];
+        }
+        break;
```

Anna-Tkachenko (approved):

PR #1141, opened 2022-05-02
solution

Sardorbek1122:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const stateClone = { ...state };
+  const results = [];
+
+  for (const action of actions) {
```
59MrRobot, line 14: You should destructure the object here

59MrRobot (changes requested):
Good job, just one comment

59MrRobot (approved):
Good job 👍

Sardorbek1122:
> Good job 👍

thank you for your suggestion.

src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateClone) {
+          delete stateClone[key];
+        }
+        break;
+
+      default:
```
ihor-jpeg, line 32: there needs to be a defined action for the default case. here it can be simply returning the same state

ihor-jpeg (changes requested):

VinnieJ-2k20 (approved):

PR #1142, opened 2022-05-02
solution

Uktam19980416:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
...
+      }
+    }
+
+    if (type === 'clear') {
+      for (const key in stateObj) {
+        delete stateObj[key];
+      }
+    }
+    result.push({ ...stateObj });
+  }
```
ihor-jpeg, line 37: it seems like a good opportunity for a switch case

ihor-jpeg (changes requested):

ihor-jpeg (approved):

VinnieJ-2k20 (approved):

PR #1147, opened 2022-05-09
sol

vinnibea:


src/transformStateWithClones.js
```diff
@@ -1,13 +1,31 @@
...
- * @param {Object} state
- * @param {Object[]} actions
- *
- * @return {Object[]}
- */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const current = { ...state };
+  const resultingArray = [];
+  let newr;
```
mgrinko, line 6: I don't understand this name

src/transformStateWithClones.js
```diff
@@ -1,13 +1,31 @@
...
+
+  for (let i = 0; i < actions.length; i++) {
```
mgrinko, line 8: Do you really need an index?

src/transformStateWithClones.js
```diff
@@ -1,13 +1,31 @@
...
+      }
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      for (const word in actions[i].keysToRemove) {
+        delete current[actions[i].keysToRemove[word]];
+      }
+    }
+
+    if (actions[i].type === 'clear') {
```
mgrinko, line 21: You have repeated `actions[i].type ===` 3 times
Do you really need it?

src/transformStateWithClones.js
```diff
@@ -1,13 +1,31 @@
...
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
+      for (const prop in actions[i].extraData) {
+        current[prop] = actions[i].extraData[prop];
+      }
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      for (const word in actions[i].keysToRemove) {
```
mgrinko, line 16: Why do you use 3 different words for the same thing `prop`, `word` and `key`?

src/transformStateWithClones.js
```diff
@@ -1,13 +1,31 @@
...
+        delete current[actions[i].keysToRemove[word]];
+      }
+    }
+
+    if (actions[i].type === 'clear') {
+      for (const key in current) {
+        delete current[key];
+      }
+    } newr = { ...current }; resultingArray[i] = newr;
```
mgrinko, line 25: Never do like this!!!
You put 2 command in the line with a closing `}`
Why?!!!!

mgrinko: Why do you need `newr`?

mgrinko (changes requested):

src/transformStateWithClones.js
vinnibea: I may be mistaken, but I though that since we have array of few objects then we should go through this array with loop, to get access to every element, and then compare values of the 'type' key with the commands, we were give

src/transformStateWithClones.js
vinnibea: Yep, it was awful decision to give that name to this, I will change it

src/transformStateWithClones.js
vinnibea: When I was trying to use only const current, my solution gave three three identical lines, which were the result of all the performed manipulations with the input . Second object has fixed this, so I have got right solving for all tests. 
I`ll try do to that in another way

src/transformStateWithClones.js
vinnibea: I don't know. I didn't think it was important at this point, since those variables were used only in those Individual loops...
I`ll change it

src/transformStateWithClones.js
jstmpelowycz: use for-of loop instead

jstmpelowycz (changes requested):
review the checklist one more time, p.2 + and don't forget to create vars where needed

vinnibea:
> review the checklist one more time, p.2 + and don't forget to create vars where needed
ніби пушнув вранці інший варіант, зараз закину

src/transformStateWithClones.js
```diff
@@ -1,13 +1,39 @@
...
-
-/**
- * @param {Object} state
- * @param {Object[]} actions
- *
- * @return {Object[]}
- */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const current = { ...state };
```
jstmpelowycz, line 3:
```suggestion
  const stateClone = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -1,13 +1,39 @@
...
+        }
+        break;
+
+      default:
+
+        return state;
+    }
+
+    resultingArray[i] = { ...current };
+    i++;
```
jstmpelowycz, line 32: try using `push` method, you can learn more [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push)

PR #1151, opened 2022-05-15
solution

pushkalov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = { ...state };
+  const resultArr = [];
```
```suggestion
  const states = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+
+  for (const key of actions) {
```
Bohdan-Maksymiuk, line 13:
```suggestion
  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = { ...state };
```
Bohdan-Maksymiuk, line 10:
```suggestion
  const currentState = { ...state };
```

Bohdan-Maksymiuk (changes requested):
Good job, but you should use clear names for variables

PR #1152, opened 2022-05-15
State with Clones Solution

berko228:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const dataArr = [];
```
Bohdan-Maksymiuk, line 10:
```suggestion
  const states = [];
```
?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+  const fullData = { ...state };
```
```suggestion
  const primaryState = { ...state };
```
?

Bohdan-Maksymiuk: or better `currentState`

Bohdan-Maksymiuk (changes requested):
Good job 👍 
You should use clear naming

PR #1161, opened 2022-05-25
Solution

G0odvin:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,31 @@
...
+          delete newState[key];
+        }
+        break;
+
+      case 'clear':
+        for (const del in newState) {
+          delete newState[del];
+        }
+        break;
+    }
```
ArturErmolenko, line 31: default case must be here

ArturErmolenko (approved):
Good job 👍 but don't forget default case, its important!

PR #1167, opened 2022-06-02
add solution

Galyopa:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+  const cloneState = { ...state };
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
+        for (const data in extraData) {
+          cloneState[data] = extraData[data];
+        }
```
illiarerun, line 20: You have a method to do this without loop https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        for (const key of keysToRemove) {
+          delete cloneState[key];
+        }
+
+        break;
+
+      case 'clear':
+        for (const key in cloneState) {
+          delete cloneState[key];
+        }
```
illiarerun, line 34: Try to find a way to do it without loop

illiarerun (approved):
GJ, left few comments:

PR #1169, opened 2022-06-02

VolodymyrShymanskyi:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let result = { ...state };
+  const resArray = [];
+
+  for (const act of actions) {
+    if (act.type === 'addProperties') {
+      Object.assign(result, act.extraData);
+      resArray.push({ ...result });
```
aliceblunt8, line 16: remove all push expect the last one, let's push to result at the end of each iteration

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      continue;
```
aliceblunt8, line 17: remove continue, you've already have if block, if the condition hasn;t met the code won't other if-blocks

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let result = { ...state };
+  const resArray = [];
+
+  for (const act of actions) {
+    if (act.type === 'addProperties') {
```
aliceblunt8, line 14: I think it's better to use switch case in this task, but no strong opinion

aliceblunt8 (changes requested):

VolodymyrShymanskyi:
Changes were made.
Recommendations were accepted.
Thanks.

aliceblunt8 (approved):

PR #1172, opened 2022-06-04
Solution

KateRiaz:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+        for (const keyToDelete of action.keysToRemove) {
+          delete objOfresult[keyToDelete];
+        }
+        break;
+
+      case 'clear':
+        for (const key in objOfresult) {
+          delete objOfresult[key];
+        }
+        break;
```
ArturErmolenko, line 29: add default case

ArturErmolenko (changes requested):
Good job 👍 
Don't forget about default case 
also for best practice  you can destructuring action like
`const { type, extraData, keysToRemove } = action;`

src/transformStateWithClones.js
KateRiaz: > add default case

what should the default case do?

KateRiaz: @ArturErmolenko what should the default case do?

aliceblunt8 (approved):

PR #1186, opened 2022-06-07
Solution

AntonShaga:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const result = [];
```
jstmpelowycz, line 11: giving your vars abstract names is awesome, but not this much, it may confuse other reviewers
generally, avoid giving names containing `result`, `res`, `arr`, `st`, instead do something like this – `arrayOfStates`, etc.

jstmpelowycz (changes requested):

illiarerun (approved):

PR #1206, opened 2022-06-07
Solution

LesyaSvk:


GavrilyukYura (approved):
Отличная робота!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arr = [];
+  const copy = { ...state };
+
+  for (const key of actions) {
```
jstmpelowycz, line 13:
```suggestion
  for (const action of actions) {
```

jstmpelowycz: as far as you iterate over array, you shouldn't name it `key`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+    switch (key.type) {
+      case 'addProperties':
+        Object.assign(copy, key.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const remove of key.keysToRemove) {
```
jstmpelowycz, line 20:
```suggestion
        for (const key of key.keysToRemove) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+          delete copy[remove];
+        }
+        break;
+
+      case 'clear':
+        for (const name in copy) {
```
jstmpelowycz, line 26:
```suggestion
        for (const key in copy) {
```

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #1212, opened 2022-06-08
Solution

Bohdan-Pyschchyk:


yanaost (approved):
nice var naming)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+
+  for (const { type, extraData, keysToRemove } of actions) {
+    switch (type) {
+      case 'addProperties':
+        Object.assign(stateClone, extraData);
+        break;
+
+      case 'removeProperties':
+        for (const keyRemove of keysToRemove) {
+          if (stateClone.hasOwnProperty(keyRemove)) {
```
jstmpelowycz, line 21: there is no need to verify if object contains it
when we try to delete non-existing key-value pair, it won't cause anything bad

jstmpelowycz (approved):

PR #1216, opened 2022-06-08
Solution

ostapfriend:


LesyaSvk (approved):
Wow, nice work ❗❗❗

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [];
+
+  for (let index = 0; index < actions.length; index++) {
+    const action = actions[index];
+    let newState;
+    const prewState = states[index - 1] || state;
```
jstmpelowycz, line 15:
```suggestion
    const prevState = states[index - 1] || state;
```

jstmpelowycz: :harold:

jstmpelowycz (commented):
I've never seen such a solution, looks interesting :fire

jstmpelowycz (approved):

PR #1217, opened 2022-06-08
solution

Kate-Khon:


AKomashchuk (approved):
👍

src/transformStateWithClones.js
```diff
@@ -1,13 +1,49 @@
 'use strict';
 
+function transform(state, action) {
+  switch (action.type) {
```
illiarerun, line 4: make use of object destructuring to avoid writing `action.` everytime

PR #1220, opened 2022-06-08
solution

viktorlyt:


illiarerun (changes requested):
GJ, one little fix left:

src/transformStateWithClones.js
```diff
@@ -1,13 +1,51 @@
...
+      return {
+        ...prevState,
+        ...action.extraData,
+      };
+
+    case 'removeProperties': {
+      const newState = { ...prevState };
+
+      for (const key of action.keysToRemove) {
+        if (Object.keys(newState).includes(key)) {
```
jstmpelowycz, line 15: if you try to access a non-existing property and then delete it, nothing will happen, thus this validation is simply redundant

jstmpelowycz (approved):
wow! nice decomposition 🔥 
love your solution

PR #1224, opened 2022-06-08
Solution

aedricson:


IvanGrabovsky (approved):
in my opinion - it's a good job!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+          delete newStateItem[key];
+        }
+        break;
+
+      case 'clear':
+        newStateItem = {};
+        break;
+
+      default:
+        return newStateItem;
```
jstmpelowycz, line 30: if you look at the description of your method, you'll see return value type `Object[]`, thus here you shouldn't return anything

jstmpelowycz, line 30:
```suggestion
```

jstmpelowycz (approved):

PR #1226, opened 2022-06-08
Solution

DirectX14:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+      };
+
+    case 'removeProperties':
+      const newState = { ...prevState };
+
+      for (const key of action.keysToRemove) {
+        delete newState[key];
+      }
+
+      return newState;
```
illiarerun, line 41: you should always have `default` case

illiarerun (approved):
GJ, left one comment:

illiarerun (changes requested):
Actually, you should definitely fix it

mgrinko (approved):

PR #1228, opened 2022-06-08
finished transform function

omelika:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let prevState = state;
+
+  for (const action of actions) {
+    let newState = transform (prevState, action);
```
illiarerun, line 14: why let here? you don't change it anywhere
```suggestion
    const newState = transform (prevState, action);
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+    case 'addProperties':
+      return {
+        ...prevState,
+        ...action.extraData,
+      };
+
+    case 'removeProperties': {
+      const newState = { ...prevState };
+
+      for (let keys of action.keysToRemove) {
```
illiarerun, line 36: why let and why plural? you don't change it's value and it's one key
```suggestion
      for (const key of action.keysToRemove) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+        delete newState[keys];
+      }
+
+      return newState;
+    }
+
+    default:
+      return prevState;
```
illiarerun, line 44: you return the same object here, which means that if you'd change it, the previous value in array would change too, you should return the copy of it
```suggestion
      return { ...prevState };
```

Olexandr6 (changes requested):
I can't see your commit with changes

Olexandr6 (approved):
Woww, good solution🔥

PR #1233, opened 2022-06-09
solution was added

Bielichenko:


vsupruniuk:
Interesting solution🤔

src/transformStateWithClones.js
```diff
@@ -7,7 +7,45 @@
...
+  const arr = [];
+  let stateCopy = { ...state };
+
+  for (const action of actions) {
+    arr.push(addAction(action));
+  }
+
+  return arr;
+
+  function addAction(action) {
```
jstmpelowycz, line 19: you could move this one outside of the `transformStateWithClones` function

jstmpelowycz (approved):

PR #1236, opened 2022-06-13
solution

DatsOlena:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
Bohdan-Maksymiuk, line 10: use clear naming - better `stateVersions` or smth else semantic

Bohdan-Maksymiuk (approved):
Good job 🥇

PR #1273, opened 2022-07-05
added transform states with clones ffunction

Oleksandr-Pustovit:


dashasm (changes requested):
По checklist варто використати switch (і не забудь про default)

Oleksandr-Pustovit:
Actually I think switch is more of a recomendation. It seems quite overloaded with loops inside it.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+
+      case 'clear':
+        for (const key in prototype) {
+          delete prototype[key];
+        }
+        break;
+
+      default:
+        break;
```
VETALEK, line 32: you may throw some error here

VETALEK (approved):
🔥

src/transformStateWithClones.js
Oleksandr-Pustovit: good point :)
As soon as I learn how to handle errors I will :)

PR #1276, opened 2022-07-05
Solution

LeonidPustovit:


DashaPoliazz (approved):
Nice solution, same story with `obj['keyName']`.

You are using obj['keyName'] syntaxis to get object value. I know that you know you can do it easier by using obj.keyName. I heard that second variant is best practise. Your variant we must use when we need to get so type of keys like 'key Name' or another specific key.

Also you can create copy of object with much shorter way `{ ...obj }`

OlgaStakhova (approved):
Як на мене гарна робота.

src/transformStateWithClones.js
```diff
@@ -5,9 +5,43 @@
...
+ *
+ * transformStateWithClones(state, [
+ * {type: 'addProperties', extraData: {name: 'Jim', hello: 'world'}},
+ * {type: 'removeProperties', keysToRemove: ['bar', 'hello']},
+ * {type: 'addProperties', extraData: {another: 'one'}}
+ * ])
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arrayOfClones = [];
```
solaryasha, line 16: https://mateacademy.slack.com/archives/C03H0MLSFUJ/p1656940534648619

src/transformStateWithClones.js
```diff
@@ -5,9 +5,43 @@
...
+  const newState = Object.assign({}, state);
+
+  for (const action of actions) {
+    switch (action['type']) {
```
solaryasha, line 20: You don't need to use square brackets for accessing static key

solaryasha (changes requested):

src/transformStateWithClones.js
```diff
@@ -5,9 +5,43 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in newState) {
+          delete newState[key];
+        }
+        break;
+
+      default:
```
VETALEK, line 37: throw some error here

VETALEK (changes requested):
GJ 👌

src/transformStateWithClones.js
LeonidPustovit: Vetalek, I don't understand this comment, please cpecify, what should I do.

Ihor-Karpyn (approved):

PR #1277, opened 2022-07-05
Solution

VolodymyrDolishnii:


DashaPoliazz (changes requested):
Nice work, perhaps You forgot about default value of switch case constuction

VETALEK (changes requested):
Good start 👍

VETALEK (approved):
👍

PR #1283, opened 2022-07-05
Solution

budanov20:


AndrewBessik (approved):

solaryasha (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+        break;
+
+      case 'clear':
+        for (const key in currentState) {
+          delete currentState[key];
+        }
+        break;
+
+      default:
+        break;
```
VETALEK, line 33: throw some error here, so user can know, that he gave an unknown action type
(like `throw new Error('error message')`)

VETALEK (changes requested):
Nice 👌 
commented

src/transformStateWithClones.js
budanov20: oK

solaryasha (approved):
Good job !

PR #1290, opened 2022-07-05
Solution

PavloMarkov:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const prop in stateCopy) {
+          delete stateCopy[prop];
+        }
+        break;
+
+      default:
```
VETALEK, line 34: mind throwing some error here, so user can understand, that he gave an unknown action type

VETALEK (changes requested):
GJ 🔥 
commented

src/transformStateWithClones.js
VETALEK: (in `default` case)

VETALEK (approved):
🚀

PR #1299, opened 2022-07-05
Solution

rodiontseva:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        break;
+
+      case 'clear':
+        for (const key in clonedState) {
+          delete clonedState[key];
+        }
+        break;
+
+      default:
+        return 'this doesn\'t run';
```
VETALEK, line 32: throw an error here

VETALEK (changes requested):
GJ 👍 
commented

PR #1301, opened 2022-07-06
Solution

Lentyaeev:


Vova-Balyshev (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,50 @@
...
+
+  let stateMod = {};
+
+  let i = 0;
+
+  Object.assign(stateMod, state);
+
+  stateArray.push(stateMod);
+
+  actions.forEach(e => {
```
VETALEK, line 20:
```suggestion
  actions.forEach(action => {
```

why `e`?

VETALEK, line 20:
```suggestion
  Object.assign(stateMod, state);

  stateArray.push(stateMod);

  actions.forEach((e, i) => {
```

you can get element's index inside `forEach`

VETALEK (changes requested):
Good start 👌

src/transformStateWithClones.js
Lentyaeev: e for me stands for element and looks fancy😊

PR #1309, opened 2022-07-07
Solution

Kulchevych:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        delete newState[key];
+      }
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      for (const key in newState) {
+        if (actions[i].keysToRemove.includes(key)) {
+          delete newState[key];
+        }
+      }
```
sergii-nosachenko, line 29: You can freely skip this check and just loop through all `actions[i].keysToRemove` values. `delete` operator will skip all unexistable keys [without throwing exception](https://stackoverflow.com/a/16406935).

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
-  // write code here
+  const newState = { ...state };
+  const resultStates = [];
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === `addProperties`) {
+      Object.assign(newState, actions[i].extraData);
+    }
+
+    if (actions[i].type === `clear`) {
```
sergii-nosachenko, line 18:
```suggestion
    if (actions[i].type === 'clear') {
```

Maybe better to use single quotes here and up.

sergii-nosachenko (approved):
Good job! 👍
As always - some comments for your decision. Not critical.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const resultStates = [];
```
dellachaise, line 11: Better create more meaningful variable name, for example `clones`.

dellachaise (changes requested):
Everything works great 👍 , but better rewrite to `switch case` instead of `if`.

dellachaise (approved):
Well done 👍

PR #1311, opened 2022-07-08
Solution

aleksandra-prognimak:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,29 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
-}
+  const arr = [];
```
Kulchevych, line 11: Напевно краще задавати зміннам більш смислового значення

src/transformStateWithClones.js
```diff
@@ -8,6 +8,29 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateNew[key];
+        }
+        break;
 
+      case 'clear':
+        for (const key in stateNew) {
+          delete stateNew[key];
+        }
```
AndriiRut, line 29: or you could rewrite the stateNew because you don't mutate incoming object and need to return the new obj.
by rewritten you don't do all that iteration in this loop)

AndriiRut (approved):

PR #1315, opened 2022-07-11
solution

hserpukhovitina:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+      Object.assign(changearray, action.extraData);
+    }
+
+    if (typeAction === 'clear') {
+      for (const key in changearray) {
+        delete changearray[key];
+      }
+    }
+
+    if (typeAction === 'removeProperties') {
```
alexandra-protyanova, line 31: everything is very good, but use `switch case`

alexandra-protyanova (changes requested):
Fix comments

alexandra-protyanova (approved):

PR #1322, opened 2022-07-20
Solution

Serhii-Vovk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        actionSteps.push(Clone(copyState));
+        break;
+
+      case 'addProperties':
+        Object.assign(copyState, data.extraData);
+
+        actionSteps.push(Clone(copyState));
+    }
+  }
+
```
valerii-derkach, line 39:
```suggestion
        break;
        
      case 'removeProperties':
        for (const remove of data.keysToRemove) {
          delete copyState[remove];
        }
        break;

      case 'addProperties':
        Object.assign(copyState, data.extraData);
    }
            actionSteps.push(Clone(copyState));
  }

```

move the same logic outside, for one iteration it will be only one case, so it will work every time, no need to repeat yourself

and add default case : break

valerii-derkach (changes requested):

valerii-derkach (approved):

PR #1326, opened 2022-07-25
First commit

dm-sheremetiev:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        }
+
+        cloneArray.push({ ...cloneState });
+        break;
+
+      case 'clear':
+        cloneState = {};
+        cloneArray.push({ ...cloneState });
+        break;
+    }
```
Bohdan-Maksymiuk, line 35:
```suggestion
  for (const item of actions) {
    switch (item.type) {
      case 'addProperties':
        for (const key in item.extraData) {
          cloneState[key] = item.extraData[key];
        }

        break;

      case 'removeProperties':
        for (const key of item.keysToRemove) {
          delete cloneState[key];
        }
        
        break;

      case 'clear':
        cloneState = {};
        break;
    }
    
    cloneArray.push({ ...cloneState });
```
your repeated this code this 3 time instead of write after switch once

Bohdan-Maksymiuk (changes requested):

src/transformStateWithClones.js
Bohdan-Maksymiuk:
```
    cloneArray.push({ ...cloneState });
```

Bohdan-Maksymiuk (approved):

PR #1334, opened 2022-07-31
solution

AndreyJigijalkin:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+
+  let stateful = { ...state };
+  let copy = { ...state };
+  let copyClear = { ...state };
```
AlexanderKolomiiets, line 14: you can leave only one variable and push result of it at the end of the loop

AlexanderKolomiiets (approved):
great job, but still you could make your code simpler )

MaksimusCRS (changes requested):
Use switch/case

MaksimusCRS (changes requested):
Add default case

MaksimusCRS (approved):

PR #1338, opened 2022-08-01
Task_transform-state-with-clones

MarkShapka:


alexandra-protyanova (changes requested):
Very good, but it'll be better to use `switch` here

src/transformStateWithClones.js
```diff
@@ -8,6 +8,42 @@
...
+  const obj = {};
+  const arr = [];
+
+  for (const key in state) {
+    obj[key] = state[key];
+  }
+
+  for (let i = 0; i < actions.length; i++) {
+    const array = Object.values(actions[i]);
+    const data = array[1];
```
Bohdan-Maksymiuk, line 20: looks like hardcode(((
```suggestion
  for (const action of actions) {
    const { type, extraData, keysToRemove } = action;
```
you can use [Destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) here

Bohdan-Maksymiuk: and switch by `type`

Bohdan-Maksymiuk (changes requested):

valerii-derkach (approved):

PR #1341, opened 2022-08-02
Solution

Dmytro-Matsehora:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+      Object.assign(clone1, actions[key].extraData);
+    } else if (actions[key].type === 'removeProperties') {
+      for (let i = 0; i < actions[key].keysToRemove.length; i++) {
+        delete clone1[`${actions[key].keysToRemove[i]}`];
+      }
+    } else if (actions[key].type === 'clear') {
+      for (const j in clone1) {
+        delete clone1[j];
+      }
+    }
```
tiserett, line 24:
```suggestion
  for (let key = 0; key < actions.length; key++) {
    if (actions[key].type === 'addProperties') {
      Object.assign(clone1, actions[key].extraData);
    } else if (actions[key].type === 'removeProperties') {
      for (let i = 0; i < actions[key].keysToRemove.length; i++) {
        delete clone1[`${actions[key].keysToRemove[i]}`];
      }
    } else if (actions[key].type === 'clear') {
      for (const j in clone1) {
        delete clone1[j];
      }
    }
```
You can iterate through array of actions with FOR OF, code will look simpler.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+
+    const clone2 = { ...clone1 };
+
+    clonesArray.push(clone2);
```
tiserett, line 28:
```suggestion
    clonesArray.push({ ...clone1 });
```
works same way, but without additional variable

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clone1 = { ...state };
```
tiserett, line 10:
```suggestion
  const clone1 = { ...state };
```
we don't have clone2 anymore, so you can change name clone1 to clone ;)

tiserett (changes requested):
just a few small fixes

src/transformStateWithClones.js
Dmytro-Matsehora: hmm, yes indeed, thanks)
but not FOR OFF, but FOR IN =)

src/transformStateWithClones.js
Dmytro-Matsehora: thanks for the helpful tips, i fixed it 👍

src/transformStateWithClones.js
tiserett: actions is an array, if I'm not mistaken, with arrays we should use for of, but if it works with in, then it's ok ;))

tiserett (approved):
good job!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,26 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clone = { ...state };
+  const clonesArray = [];
+
+  for (const key in actions) {
```
dellachaise, line 13: Better use `for of` cycle, this way you won't need to write `actions[key]`, only `action` for example.

dellachaise (changes requested):
Great job 👍 

But better rewrite with `for of` cycle.

danylo-onishchuk (approved):

PR #1342, opened 2022-08-02

AnastasiaTitarenko:


ArturErmolenko (changes requested):
use switch case try to avoid if else

src/transformStateWithClones.js
```diff
@@ -7,7 +7,46 @@
...
+
+            for (const keyToRemove of actions[i].keysToRemove) {
+              delete removePropertiesClone[keyToRemove];
+            }
+            resultArray.push(removePropertiesClone);
+            break;
+          default:
+            removePropertiesClone = Object.assign({}, resultArray[i - 1]);
+
+            for (const keyToRemove of actions[i].keysToRemove) {
```
Bohdan-Maksymiuk, line 38: why do you need such complicated structure?

Bohdan-Maksymiuk: you should iterate through actions and then switch through `action.type` and describe logic into cases

Bohdan-Maksymiuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+          removePropertiesClone = Object.assign({}, resultArray[i - 1]);
+
+          for (const keyToRemove of actions[i].keysToRemove) {
+            delete removePropertiesClone[keyToRemove];
+          }
+          resultArray.push(removePropertiesClone);
+        }
+        break;
+      default:
+        resultArray.push({});
```
Bohdan-Maksymiuk, line 41: get rid of this nesting `if ... else`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultArray = [];
+  let removePropertiesClone = {};
```
```suggestion
  let propertiesClone = {...state};
```
this is clone only for remove? lets make it as a clone for current state

Bohdan-Maksymiuk (changes requested):
if smth remains unclear 0 ask in fe_chat and tag mentors

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultArray = [state];
+  let removePropertiesClone = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
```
Bohdan-Maksymiuk, line 13: you can use here `for of` loop
it will help you to avoid writing index every time

Bohdan-Maksymiuk (approved):

PR #1355, opened 2022-08-07
Solution

zhenija:


alexandra-protyanova (changes requested):
Let us use `switch` for this task

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clone = [];
+  const cloneObject = { ...state };
+
+  for (const key of actions) {
+    switch (key.type) {
+      case 'addProperties':
+        Object.assign(cloneObject, key.extraData);
+        clone.push({ ...cloneObject });
```
ArturErmolenko, line 17: you can push only once

ArturErmolenko (changes requested):
No need to push in each case. Write it once in the end of iteration(after switch case).

marina-tilniak (approved):

PR #1356, opened 2022-08-07
Solution

nikitaovch99:


artem-musii (commented):
"something" isn't the best name for variable and also better to use switch

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+  const arr = [];
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i]['type'] === 'addProperties') {
+      Object.assign(something, actions[i]['extraData']);
+    } else if (actions[i]['type'] === 'removeProperties') {
+      const remove = Object.keys(something);
+
+      for (let g = 0; g < remove.length; g++) {
+        for (let k = 0; k < actions[i].keysToRemove.length; k++) {
```
naziamoff, line 20: You have to avoid nested loops where it is possible

naziamoff (changes requested):
You should look through the checklist again

naziamoff (approved):
Well done 🔥

PR #1364, opened 2022-08-08
Solution

smiyka:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const allVersionsOfState = [];
+  const versionOfState = { ...state };
+
+  for (const objOfActions of actions) {
+    if (objOfActions['type'] === 'addProperties') {
```
alexandra-protyanova, line 14:
```suggestion
    if (objOfActions.type === 'addProperties') {
```

alexandra-protyanova (changes requested):
let us use switch case in this task

ArturErmolenko (approved):
Well done 👍

PR #1365, opened 2022-08-08

BipBobBoop:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        break;
+
+      case action.type === 'removeProperties':
+        for (const keyToRemove of action.keysToRemove) {
+          if (copy.hasOwnProperty(keyToRemove)) {
+            delete copy[keyToRemove];
+          }
+        }
+
+        res.push({ ...copy });
```
Bohdan-Maksymiuk, line 27: you could push just after switch, not in every case

Bohdan-Maksymiuk (changes requested):

src/transformStateWithClones.js
BipBobBoop: > you could push just after switch, not in every case

ty, done

ArturErmolenko (approved):

PR #1366, opened 2022-08-08
Solution

Jamal-Amoian:


oksana-logos-frontend (approved):
Good)
In my opinion for iteration of array better use a loop for...of, and checklist says better use switch than if in this task.

MaksimusCRS (changes requested):
2 checklist point

MaksimusCRS (approved):

PR #1371, opened 2022-08-08
first commit

nikolay-li:


src/transformStateWithClones.js
```diff
@@ -28,6 +28,9 @@ function transformStateWithClones(state, actions) {
           delete clone[key];
         }
         break;
+
+      default:
+        break;
```
anyakrainiuk, line 33: I think it's better to write some info about Error, not just break.
for example : 'Error'

anyakrainiuk (changes requested):

src/transformStateWithClones.js
nikolay-li: fixed =)

anyakrainiuk (approved):

danylo-onishchuk (approved):

PR #1378, opened 2022-08-09
Solution

Mejeezz:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,51 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties': {
+        const { extraData } = action;
+        const newState = {
+          ...lastState, ...extraData,
+        };
+
+        lastState = newState;
+        history.push(lastState);
```
ibondrk, line 22: you perform same action in all switch cases - do it just once afterwards.

ibondrk (changes requested):
You can just remove your push row from case into the loop and it`ll become even better. In general GJ.

ibondrk (approved):
GJ🔥

MaksimusCRS (approved):
GJ

PR #1383, opened 2022-08-09
Solution

wfurs0w:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      case 'removeProperties':
+        for (const char of action.keysToRemove) {
+          delete copyState[char];
+        }
+        break;
+
+      case 'clear':
+        for (const key in copyState) {
+          delete copyState[key];
+        }
```
rboiar, line 28: Add default case

rboiar (changes requested):
Good, but you should add default case

rboiar (approved):
👍

MaksimusCRS (approved):

PR #1396, opened 2022-08-09
Solution

Anatolii-Bukoros:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        for (const key of keysToRemove) {
+          delete newState[key];
+        }
+        break;
+      case 'clear':
+        for (const key in newState) {
+          delete newState[key];
+        }
+        break;
+      default:
```
RomanYakm, line 30: I think default should have value; throw new Error(`Unknown action type: ${action.type}`);

RomanYakm (approved):
GJ. Just one suggestion.

src/transformStateWithClones.js
MaksimusCRS, line 30: At least break required

MaksimusCRS (approved):

PR #1402, opened 2022-08-10
solution

vadym-zahazei:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,55 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const history = [];
+  const cloneOfState = { ...state };
+
+  for (const action of actions) {
+    const { extraData, keysToRemove, type } = action;
+
+    switch (type) {
+      case 'addProperties': {
```
AndriiRut, line 17: the same.

AndriiRut (changes requested):

src/transformStateWithClones.js
vadym-zahazei: > the same. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

done

AndriiRut (approved):
🚀

PR #1410, opened 2022-08-11
Solution

volodymir-tymtsias:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+      case 'removeProperties':
+        for (const keyToRemove of action.keysToRemove) {
+          if (keyToRemove in resultState) {
+            delete resultState[keyToRemove];
+          }
+        }
+        resultArray.push({ ...resultState });
+        break;
+    }
```
valerii-derkach, line 35:
```suggestion
          delete resultState[key];
        }
        break;

      case 'addProperties':
        resultState = Object.assign(resultState, action.extraData);
        break;

      case 'removeProperties':
        for (const keyToRemove of action.keysToRemove) {
          if (keyToRemove in resultState) {
            delete resultState[keyToRemove];
          }
        }
        break;
    }
    
      resultArray.push({ ...resultState });
```

If you perform same action in all switch cases - do it just once afterwards.

valerii-derkach (changes requested):

valerii-derkach (approved):

PR #1416, opened 2022-08-14
Transform state with clones was created

ArmPlay:


ArturErmolenko (changes requested):
Better use switch case ,try to avoid if use construction

stanislav-topikha (changes requested):
Screenshot 2022-08-19 at 16 35 43

PR #1421, opened 2022-08-15
Solution

TheVovchik:


src/transformStateWithClones.js
```diff
@@ -1,13 +1,55 @@
...
+      toRemove(actions[i].keysToRemove);
+      container = Object.assign({}, newState);
+      returnObjects.push(container);
+    }
+
+    if (actions[i].type === 'addProperties') {
+      toAdd(actions[i].extraData);
+      container = Object.assign({}, newState);
+      returnObjects.push(container);
+    }
```
grafit-off, line 31: 1. If you used the for loop, it's gonna be better if you set a name for actions[i], for better readability
2. You should use the switch/case statement as written in the checklist
image

src/transformStateWithClones.js
```diff
@@ -1,13 +1,55 @@
...
-  // write code here
+  const newState = { ...state };
+  const returnObjects = [];
+  let container;
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'clear') {
+      toClear();
+      container = Object.assign({}, newState);
+      returnObjects.push(container);
```
grafit-off, line 18: If you perform the same action in all conditions - do it just once afterwards.

src/transformStateWithClones.js
```diff
@@ -1,13 +1,55 @@
...
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      toRemove(actions[i].keysToRemove);
+      container = Object.assign({}, newState);
+      returnObjects.push(container);
```
grafit-off, line 24: If you perform the same action in all conditions - do it just once afterwards.

src/transformStateWithClones.js
```diff
@@ -1,13 +1,55 @@
...
+    }
+
+    if (actions[i].type === 'addProperties') {
+      toAdd(actions[i].extraData);
+      container = Object.assign({}, newState);
+      returnObjects.push(container);
```
grafit-off, line 30: If you perform the same action in all conditions - do it just once afterwards.

src/transformStateWithClones.js
```diff
@@ -1,13 +1,55 @@
...
+ * @param {Object} newState
  * @param {Object[]} actions
  *
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const returnObjects = [];
+  let container;
```
grafit-off, line 12: To be honest, that container looks unnecessary to me.

grafit-off (approved):
Keep going 👯‍♂️

src/transformStateWithClones.js
```diff
@@ -1,13 +1,41 @@
...
- * @param {Object} state
+ * @param {Object} newState
  * @param {Object[]} actions
  *
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const returnObjects = [];
```
naziamoff, line 11: this name looks like a name of a function to me.
When we name a function - its name should describe what it does, so it should be a verb
When we name a variable - it should describe **what** is stored inside, so it is better to be a noun (except for boolean variables, they start with ...should or sth like that)

```suggestion
  const objectsToReturn = [];
```
If we want to do it even better - it should be
```suggestion
  const stateVersions = [];
```

naziamoff (approved):
Well done 🔥 
but check my comment

src/transformStateWithClones.js
TheVovchik: I understand, will change it here and check other tasks. Thank's for such wide advise!

TheVovchik: named them stateClones

PR #1426, opened 2022-08-17
Solution 4

mary-kucher:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+      }
+    }
+
+    if (action.type === 'clear') {
+      cloneState = {};
+    }
+
+    newArray.push({ ...cloneState });
+  }
+
```
marina-tilniak, line 30: it's a good case for using switch case

marina-tilniak (changes requested):

alexandra-protyanova (approved):
Well done!

PR #1430, opened 2022-08-19
Clone solution

Vladus173000:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [];
+  let rollingState = { ...state };
+
+  for (const action of actions) {
+    switch (action['type']) {
```
Bohdan-Maksymiuk, line 14: do you really need this `[]` brackets ?
you can use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) before switch and then use properties

Bohdan-Maksymiuk (changes requested):

ArturErmolenko (approved):
good job 👍 
for improving your code you could use you can use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) before switch and then use properties

PR #1433, opened 2022-08-21
Solution

IhorChaikovskyi:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        delete newObjectDel[arr];
+      }
+      newArray.push(newObjectDel);
+      newObject = { ...newObjectDel };
+    } else if (oneAction.type === 'clear') {
+      const clear = {};
+
+      newArray.push(clear);
+      newObject = {};
+    }
```
Bohdan-Maksymiuk, line 34: use switch for such cases

Bohdan-Maksymiuk: you can push new state object at the end of iteration 1 time instead of writing it into each if

Bohdan-Maksymiuk (changes requested):

valerii-derkach (approved):

PR #1439, opened 2022-08-25
Solution

Oleksii-Blyzniuk:


liliia-livandovskaya (changes requested):
GJ! Let's use a switch case instead of if

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arr = [];
+  const obj = { ...state };
```
marina-tilniak, line 11:
```suggestion
  const stateVersions = [];
  const stateCopy = { ...state };
```

marina-tilniak (approved):

PR #1451, opened 2022-08-30
Solution Transform state with clones

Kalynets:


ArturErmolenko (changes requested):
Let's use switch case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let correctState = Object.assign({}, state);
+
+  for (const action of actions) {
+    // let correctState = Object.assign({}, state);
+    // console.log(correctState);
```
tetiana-ivanchenko, line 15: remove comments

src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+    switch (action['type']) {
```
tetiana-ivanchenko, line 16:
```suggestion
    switch (action.type) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+      case 'addProperties':
+        correctState = Object.assign(correctState, action['extraData']);
```
tetiana-ivanchenko, line 18:
```suggestion
        correctState = Object.assign(correctState, action.extraData);
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+
+        const cloneAdd = Object.assign({}, correctState);
+
+        result.push(cloneAdd);
+        break;
+
+      case 'removeProperties':
+        for (const keys of action['keysToRemove']) {
```
tetiana-ivanchenko, line 26:
```suggestion
        for (const keys of action.keysToRemove) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+  for (const action of actions) {
+    // let correctState = Object.assign({}, state);
+    // console.log(correctState);
+    switch (action['type']) {
+      case 'addProperties':
+        correctState = Object.assign(correctState, action['extraData']);
+
+        const cloneAdd = Object.assign({}, correctState);
+
+        result.push(cloneAdd);
```
tetiana-ivanchenko, line 22: why do you make copy inside each case? you sould do it only once after all cases.
also you can use spread operator to make copy here

tetiana-ivanchenko (changes requested):

PR #1456, opened 2022-09-02
my cod stateful clones

RusNeday:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const j in copyState) {
+          delete copyState[j];
+        }
+        break;
+
+      default:
```
alexandra-protyanova, line 33: u don't need default case

alexandra-protyanova (approved):
GJ!

PR #1461, opened 2022-09-05
done!

dmitriyn3679:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,60 @@
...
+
+function addPropertiesFunc(copyState, addedData, stateHistory) {
+  Object.assign(copyState, addedData);
+
+  stateHistory.push(
+    {
+      ...copyState,
+      ...addedData,
+    }
+  );
```
Bohdan-Maksymiuk, line 43: it is bad practice to mutate arguments in function

Bohdan-Maksymiuk: it will be hard to debug in future

Bohdan-Maksymiuk (changes requested):

src/transformStateWithClones.js
Bohdan-Maksymiuk: I meant for all function where you mutated state 🙃

Bohdan-Maksymiuk (approved):

PR #1464, opened 2022-09-05
Solution

MaxManis:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
+  // Main Loop:
+  for (const action of actions) {
+    cloneShadow = {};
+
+    switch (action['type']) {
+      // For the type of 'addProperties':
+      case 'addProperties':
+        for (const extraItem in action['extraData']) {
+          clone[extraItem] = action['extraData'][extraItem];
+        }
```
grafit-off, line 25: You can use Object.assign for this case, instead of loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let cloneShadow = {};
```
grafit-off, line 10: It's a little bit strange for me and it looks unnecessary, to create a variable for an empty object for Object.assign in the loop below. In my opinion, it seems better just to add an empty object as an argument for the method, not as a variable. We need to ask a mentor.

grafit-off (approved):
GJ 🔥. Left some comments below.

src/transformStateWithClones.js
Denchepornyuk: agree, let's use Object.assign for adding new values to the object

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
+  const clone = {};
+  const result = [];
+
+  Object.assign(clone, state);
```
Denchepornyuk, line 14: use spread operator for cloning objects

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
+
+      // For the type of 'clear':
+      case 'clear':
+        for (const delAllItem in clone) {
+          delete clone[delAllItem];
+        }
+        break;
+    }
+
+    result.push(Object.assign(cloneShadow, clone));
```
Denchepornyuk, line 43: the same here, use spread operator

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1465, opened 2022-09-06
made function transformStateWithClones

Hlazman:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      }
+      arr.push(Object.assign({}, obj));
+    }
+
+    if (iterator.type === 'clear') {
+      for (const key in obj) {
+        delete obj[key];
+      }
+      arr.push(Object.assign({}, obj));
+    }
```
liliia-livandovskaya, line 33: Lets use switch case instead of if else

liliia-livandovskaya (changes requested):
GJ 👍  Pay attention to the comments

liliia-livandovskaya (approved):
Great 👍

PR #1468, opened 2022-09-07
Solution

SergIats:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultArr = [];
+
+  resultArr.push(Object.assign({}, state));
+
+  for (let i = 0; i < actions.length; i++) {
```
alexandra-protyanova, line 14: u do not need indexes here, use for of loop, to make your code more readable

alexandra-protyanova (changes requested):
And switch would be more suitable here as well

SergIats:
Hi!) I have implemented switch, but I don't understand how I can do loop without index, because I used this index (not only indexed objects of loop) in many places.

src/transformStateWithClones.js
liliia-livandovskaya: agree, rewrite your code with for of loop

liliia-livandovskaya (changes requested):
Pay attention to the comment

SergIats:
Hm... This time I think I did it

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+          resultArr.push(Object.assign({}, previousState));
+          break;
+
+        case 'clear':
+          previousState = {};
+          resultArr.push(Object.assign({}, previousState));
+          break;
+      }
+    }
+  }
```
liliia-livandovskaya, line 37:
```suggestion
  for (const action of actions) {
    switch (action.type) {
      case 'addProperties':
        Object.assign(previousState, action.extraData);
        break;

      case 'removeProperties':
        for (const key of action.keysToRemove) {
          delete previousState[key];
        }
        break;

      case 'clear':
        for (const key in previousState) {
          delete previousState[key];
        }
        break;
    }
    resultArr.push({ ... previousState });
  }
```

liliia-livandovskaya (changes requested):
GJ! Lets work on fix

src/transformStateWithClones.js
SergIats: Oh... Yes, you're right - your solution better and more harmonic... I will study to do it in this way. Thanks!

SanyaBratashchuk (approved):

PR #1471, opened 2022-09-11
Solution js_stateful_clones

anton-martyniuk1:


ArturErmolenko (changes requested):
Use switch case
 Follow [Checklist](https://github.com/mate-academy/js_stateful-clones/blob/master/checklist.md)

PR #1475, opened 2022-09-12
solution

artem-musii:


nikitaovch99 (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const current = { ...state };
+  const result = [];
```
naziamoff, line 11: You should keep your namings maximally clear. Neither `current` nor `result` gives info about what is stored inside.
```suggestion
  const currentVersion = { ...state };
  const stateVersions = [];
```

naziamoff (changes requested):
Well done, just small naming issues

naziamoff (approved):
Well done!

PR #1477, opened 2022-09-12
Solution

vladimir-dublya:


myroslav-diiak (approved):

src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
...
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'clear') {
+      for (const key in newState) {
+        delete newState[key];
+      }
+      container = Object.assign({}, newState);
+      returnObjects.push(container);
+    }
+
+    if (actions[i].type === 'removeProperties') {
```
Denchepornyuk, line 24: it's a perfect place for switch case instead of three if

src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
...
+  const returnObjects = [];
+  let container;
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'clear') {
+      for (const key in newState) {
+        delete newState[key];
+      }
+      container = Object.assign({}, newState);
+      returnObjects.push(container);
```
Denchepornyuk, line 21: don't push object to arrayActions in each if you can do it once at the end of the loop

src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
...
+  const newState = { ...state };
+  const returnObjects = [];
+  let container;
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'clear') {
+      for (const key in newState) {
+        delete newState[key];
+      }
+      container = Object.assign({}, newState);
```
Denchepornyuk, line 20: use spread operator instead of Object.assign

Denchepornyuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
...
+  const returnObjects = [];
+  let container;
+
+  for (let i = 0; i < actions.length; i++) {
+    switch (actions[i].type) {
+      case 'clear':
+        for (const key in newState) {
+          delete newState[key];
+        }
+        container = { ...newState };
```
Denchepornyuk, line 21: you don't need container, push copy of newState to the returnObjects at the end of the loop

Denchepornyuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const newState = { ...state };
+  const returnObjects = [];
+
+  for (let i = 0; i < actions.length; i++) {
```
Denchepornyuk, line 14: use `for of` loop here, will be easier to get values from the array

src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
...
+        for (const key in newState) {
+          delete newState[key];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key in newState) {
+          if (actions[i].keysToRemove.includes(key)) {
+            delete newState[key];
+          }
```
Denchepornyuk, line 26: you don't need if here

Denchepornyuk (changes requested):

src/transformStateWithClones.js
vladimir-dublya: without if fail tests

src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
...
+          delete newState[key];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key in newState) {
+          if (val.keysToRemove.includes(key)) {
+            delete newState[key];
+          }
+        }
```
Denchepornyuk, line 27: you can use loop for keysToRemove instead of newState and remove `if` here

Denchepornyuk (approved):

PR #1483, opened 2022-09-12
Develop

grafit-off:
Reducer completed

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -8,6 +8,37 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const initialState = { ...state };
+  const stateHistory = [];
+
+  actions.forEach((action) => {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
+        Object.assign(initialState, extraData);
+        break;
+
+      case 'removeProperties':
+        keysToRemove.forEach((key) => {
+          delete initialState[key];
+        });
+        break;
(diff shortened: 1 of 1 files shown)


src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
...
+        break;
+
+      case 'clear':
+        for (const key in initialState) {
+          delete initialState[key];
+        }
+        break;
+
+      default:
+        break;
```
MaxManis, line 35: Not sure you need this `default `satement since we know its only a 3 possible value of input))

MaxManis (approved):
Great job, nice use of forEach method)

src/transformStateWithClones.js
grafit-off: @MaxManis to be honest, the `default` statement looks unnecessary, but In checklist rules, we have something like that:
[CODE STYLE]: switch/case should always have a default case for error handling.

MaxManis: Agree, but it will just return an empty array if it wrong `type` passed))

Denchepornyuk: it's better to keep default block
usually, we write some logic in default if each case didn't work for us

Denchepornyuk (approved):

src/transformStateWithClones.js
grafit-off: @Denchepornyuk you need some time to sleep! Stop checking student's tasks until midnight)

PR #1484, opened 2022-09-12
Solution

Anna-Borodaienko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let copyState = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case `addProperties`:
+        copyState = Object.assign({}, copyState, action.extraData);
```
Yevheniya-Deryka, line 16: It seems to me you can skip the assignment of an empty object as the first argument of this function, skip rewriting copyState, and just type:
in this line.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+
+        for (const removeItem of action.keysToRemove) {
+          delete copyState[removeItem];
+        }
+        result.push(copyState);
+        break;
+
+      case `clear`:
+        copyState = {};
+        result.push(copyState);
```
Yevheniya-Deryka, line 31: And you can push copyState to your array after switch-case

Yevheniya-Deryka (approved):
Good job!
Left some comments, but don't think they are very important, so approved your variant.

src/transformStateWithClones.js
Anna-Borodaienko: Fixed, thanks for good idea!

Denchepornyuk (approved):

PR #1487, opened 2022-09-12
mom i did it

ihorandrianov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+          const { [key]: _, ...p } = object;
+
+          return p;
+        }, _state);
+
+        return updatedState;
+      case 'clear':
+        return {};
+      default:
+        break;
```
korniienkokarina, line 28: in default case you can return some error message for better error handling

korniienkokarina (commented):
good job!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+        return {
+          ...reducerState,
+          ...action.extraData,
+        };
+      case 'removeProperties':
+        const updatedState = action.keysToRemove.reduce((object, key) => {
+          const { [key]: _, ...p } = object;
+
+          return p;
+        }, reducerState);
```
Denchepornyuk, line 22: you can do with one for of loop, don't complicate the code

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+
+        return updatedState;
+      case 'clear':
+        return {};
```
Denchepornyuk, line 26: it's better to remove all key in existing object instead of creating a new one
with new one you will lose link to existing object

Denchepornyuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
-  // write code here
+  const reducer = (reducerState, action) => {
+    const changedState = { ...reducerState };
+
+    switch (action.type) {
+      case 'addProperties':
+        return {
+          ...reducerState,
+          ...action.extraData,
+        };
```
OleksiiBlezniuk, line 18: Also you can use Object.assign here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const reducer = (reducerState, action) => {
```
OleksiiBlezniuk, line 10: Would be nice to move this helper function outside of `transformStateWithClones` function to make it reusable

OleksiiBlezniuk (approved):

OleksiiBlezniuk (commented):
Create more meaningful commit messages, it’s hard to understand what was done in this commit from `fix` message.

PR #1491, opened 2022-09-12

Vladosik1337:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+          delete cloneObject[deleteValue];
+        }
+        break;
+
+      case 'clear' :
+        for (const deleteAll in cloneObject) {
+          delete cloneObject[deleteAll];
+        }
+        break;
+    }
```
glugovsky, line 34: Needs default case
```suggestion

    default:
        throw new Error('Wrong type!');
    }
```

glugovsky (approved):
GJ 👍

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const cloneObject = {};
+  const arrayWithObjects = [];
```
naziamoff, line 11: The name of the variable should tell us about the information that is stored inside, not its data type
So I would do it like that:
```suggestion
  const stateVersions = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+
+  Object.assign(cloneObject, state);
+
+  for (const objects of actions) {
+    const { type, extraData, keysToRemove } = objects;
```
naziamoff, line 16: You want your code to be maximally clear for those who read it and for yourself too.
```suggestion
  for (const action of actions) {
    const { type, extraData, keysToRemove } = action;
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        break;
+
+      case 'removeProperties' :
+        for (const deleteValue of keysToRemove) {
+          delete cloneObject[deleteValue];
+        }
+        break;
+
+      case 'clear' :
+        for (const deleteAll in cloneObject) {
```
naziamoff, line 30: same here, you should think of `what is stored inside the variable` to name it properly

naziamoff (changes requested):

naziamoff (approved):
Great job!

PR #1492, opened 2022-09-12
Solution

iam-datsenko:


iPanda177 (approved):
Well done!)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+    if (action.type === 'removeProperties') {
+      copyState = { ...copyState };
+
+      for (const property of action.keysToRemove) {
+        delete copyState[property];
+      }
+      arrayActions.push(copyState);
+    }
+
+    if (action.type === 'clear') {
```
Denchepornyuk, line 28: it's a perfect place for switch case instead of three `if`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let copyState = { ...state };
+  const arrayActions = [];
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      copyState = Object.assign({}, copyState, action.extraData);
+      arrayActions.push(copyState);
```
Denchepornyuk, line 16: don't push object to arrayActions in each `if` you can do it once at the end of the loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      copyState = { ...copyState };
+
+      for (const property of action.keysToRemove) {
+        delete copyState[property];
+      }
+      arrayActions.push(copyState);
+    }
+
+    if (action.type === 'clear') {
+      copyState = {};
```
Denchepornyuk, line 29: nope, don't create new object, we will lose link to the previous object
It's better to use delete for each key in the object

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1493, opened 2022-09-12
Solution

imaksweb:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      result.push({ ...newState });
+    }
+
+    if (action.type === 'removeProperties') {
+      action.keysToRemove.forEach(key => {
+        delete newState[key];
+      });
+
+      result.push(Object.assign({}, newState));
+    }
```
Anna-Borodaienko, line 25:
```suggestion
result.push({ ...newState });    
```

Anna-Borodaienko: It's the same.
You can move this .push from every iteration and add just one, at the end of loop

Anna-Borodaienko (commented):
And checklist wants switch, not some if )

Anna-Borodaienko (approved):

naziamoff (approved):
well done 🔥

PR #1496, opened 2022-09-12
Solution

ArturNagaycev:


jojnajbude (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const versionsOfState = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        versionsOfState.push({ ...stateCopy });
```
naziamoff, line 17: you should push the version into the new array only once after each switch

naziamoff (changes requested):
Nicely done 🔥 
But fix what's in my comment to make it perfect

naziamoff (approved):
Well done!

PR #1497, opened 2022-09-12
Solution

anastasiia-nahaiska:


Andrew-Skoryk (approved):
Good job! 🔥

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateAfterActions = [];
```
naziamoff, line 10: It is not a bad one, but a bit overcomplicated. We can make it simpler and clearer:
```suggestion
  const states = [];
```
or even better:
```suggestion
  const stateVersions = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+      case 'addProperties':
+        Object.assign(copyState, action.extraData);
+        break;
+
+      case 'clear':
+        copyState = {};
+        break;
+
+      case 'removeProperties':
+        const arrKeysToRemove = action.keysToRemove;
```
naziamoff, line 24: you don't need this variable

naziamoff (changes requested):
Well done, but there are some minor improvements to your code

naziamoff (approved):
Nice 🚀

Anna-Tkachenko (approved):

PR #1500, opened 2022-09-12
Solution

Rina07:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,42 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const prevState = Object.assign({}, state);
+  const nextActions = [];
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
Oleksandra-Shevchenko, line 15: same as previous task (use switch) :)

Oleksandra-Shevchenko (changes requested):

src/transformStateWithClones.js
Rina07: Agree:)

Denchepornyuk (approved):

PR #1502, opened 2022-09-13
Solution

rstarovoit:


MAKSOti (changes requested):
GJ but read cheklist

BAD EXAMPLE:

const copy = { ...state }

GOOD EXAMPLE:

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete copyState[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in copyState) {
+          delete copyState[key];
+        }
```
silchyn, line 28:
```suggestion
        copyState = {};
```

silchyn (changes requested):

src/transformStateWithClones.js
rstarovoit: Thanks! Very simple solution)

silchyn (approved):
Cool!

PR #1505, opened 2022-09-13
Task was solved

DenysKleimenov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  let stateCopy = { ...state };
+  const statesOfCopy = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy = { ...stateCopy }, action.extraData);
```
Yuriy-Komolov, line 16: You have unnecessary assignment here.
`Object.assign(stateCopy, action.extraData)`  would be better in my opinion, less code and operations. 
And with that you can declare your `stateCopy` variable with `const`

Yuriy-Komolov (changes requested):

naziamoff (approved):
Good job!

PR #1506, opened 2022-09-13
Transform state with clones

Yuriy-Komolov:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const clone = [];
+  let help = { ...state };
```
DenysKleimenov, line 12: use proper name for object copy (see checklist)

DenysKleimenov (changes requested):

src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const states = [];
+  let copyStates = { ...state };
```
naziamoff, line 12: I think would be better to rename it.
It is actually but one state version, so it doesn't have to be in plural in order to be clear to understand
I would come up with sth like this:

```suggestion
  let stateVersion = { ...state };
```

naziamoff (approved):
Well done! But check my comment. Naming matters

PR #1507, opened 2022-09-13
Solution

OleksandrGlulhovskiy:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      for (const key in newState) {
+        if (actions[i].keysToRemove.includes(key)) {
+          delete newState[key];
+        }
+      }
+      b = Object.assign({}, newState);
+      arr.push(b);
```
Andrew-Skoryk, line 30: 4. [DONT REPEAT YOURSELF] - If you perform same action in all switch cases - do it just once afterwards.
You can `arr.push(b)` at the end of loop. And SWITCH would be better...

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const arr = [];
+  let b;
```
Andrew-Skoryk, line 12: Not sure about naming

Andrew-Skoryk (changes requested):
Well done 🔥 , but there are some minor mistakes

PR #1508, opened 2022-09-13
Solution

korniienkokarina:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const statesArray = [];
+  const copyState = { ...state };
```
ihorandrianov, line 11: It may be better to create new object in every case and do not mutate copied state. But I am not sure this is a mistake. Everything else looks fine!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+      case 'removeProperties':
+        for (const keyToRemove of keysToRemove) {
+          delete copyState[keyToRemove];
+        }
+        break;
+
+      case 'clear':
+        const keysToClear = Object.keys(copyState);
+
+        for (const keyToClear of keysToClear) {
```
naziamoff, line 30: I believe `for...in` loop will do better here, for making it more simple. And we won't need Object.values in that case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const statesArray = [];
```
naziamoff, line 10: that is not necessary to explicitly write what datatype it is
You can do it just like that:
```suggestion
  const states = [];
```
or 
```suggestion
  const stateVersions = [];
```

naziamoff (changes requested):
Well done 🔥 
There are some comments though

naziamoff (approved):
gj

PR #1509, opened 2022-09-13
add js code who decides task

SerhiiRomaniukWD:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,47 @@
...
+      case 'addProperties':
+        copyState = {
+          ...copyState,
+          ...action.extraData,
+        };
+
+        totalPropertys.push({
+          ...copyState,
+          ...action.extraData,
+        });
```
Anna-Borodaienko, line 24: You can push your copyState to result once at the end, after switch loop, .push({ ...copyState })
And you first change copyState, then repeat these added items for push? You can use changed copyState for this

src/transformStateWithClones.js
```diff
@@ -7,7 +7,47 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const property of action.keysToRemove) {
+          if (Object.keys(copyState).filter(elem => elem === property)) {
+            delete copyState[property];
+          }
```
Anna-Borodaienko, line 31:
```suggestion
        for (const property of action.keysToRemove) {
           delete copyState[property];
          
```

src/transformStateWithClones.js
SerhiiRomaniukWD: yes, you are right

src/transformStateWithClones.js
SerhiiRomaniukWD: i did it again, i will fix it
thanks

Andr1yK (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete copyState[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in copyState) {
+          delete copyState[key];
+        }
```
silchyn, line 28:
```suggestion
        copyState = {};
```

silchyn (changes requested):

silchyn (approved):
Cool!

PR #1511, opened 2022-09-13
Solution

baranovskazoya:


src/transformStateWithClones.js
```diff
@@ -1,13 +1,35 @@
...
- * @param {Object} state
- * @param {Object[]} actions
- *
- * @return {Object[]}
- */
 function transformStateWithClones(state, actions) {
   // write code here
-}
+  const copy = { ...state };
+  const mainArray = [];
```
imaksweb, line 6: The name of the variable should tell us about the information that is stored inside
```suggestion
  const copyState = { ...state };
  const states = [];
```

imaksweb (changes requested):

imaksweb (approved):
Perfect 👍

Denchepornyuk (approved):

PR #1515, opened 2022-09-13
Solution

avkovlit:


xiimiik (commented):
All is good, but fix the mistakes from past task

src/transformStateWithClones.js
```diff
@@ -6,8 +6,39 @@
...
+  const copyState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(copyState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        const arrKeysToRemove = action.keysToRemove;
```
naziamoff, line 21: Figure out how to remove this variable

src/transformStateWithClones.js
```diff
@@ -6,8 +6,39 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultState = [];
```
naziamoff, line 11: I believe in good naming, so we should make it maximally simple to read. What we return is not actually resultState, it is an array of all state versions. So we may do it like that:
```suggestion
  const states = [];
```
or if we want even better then:
```suggestion
  const stateVersions = [];
```

naziamoff (changes requested):

naziamoff (approved):
Well done! 🔥

PR #1518, opened 2022-09-13
Solution

SanyaKovalenko:


MAKSOti (commented):
You need to read the checklist 

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let newObject = Object.assign({}, state);
```
Anna-Tkachenko, line 11: let's create more meaningful variable name instead of newObject

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+
+  for (const object of actions) {
+    switch (object.type) {
+      case 'addProperties':
+        newObject = Object.assign(newObject, object.extraData);
+        result.push(Object.assign({}, newObject));
```
Anna-Tkachenko, line 17: you do the same thing (pushing result object to some array) in each case

try to move this logic outside of the case functionality (do this by default after all cases)

Anna-Tkachenko (changes requested):

PR #1525, opened 2022-09-14
Solution

NickBrodyuk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,65 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let stateCopy = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
+    let actionsType;
+
+    switch (true) {
+      case actions[i].type === 'addProperties':
```
Denchepornyuk, line 17: add actions[i].type to the switch instead of true and remove === from cases

Denchepornyuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let stateCopy = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
+    let actionsType;
```
OleksiiBlezniuk, line 14: Same as in previous task

Better to use this object property instead of variable
(you didn't use it outside switch/case, you store different values in it)
It's harder to read your code with it

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+
+        result.push(stateCopy);
+        break;
+    }
```
OleksiiBlezniuk, line 40: What if type will have some other value?

Switch/case construction should have `default` case

OleksiiBlezniuk (changes requested):
Create more meaningful commit messages, it’s hard to understand what was done in this commit from `Solution` message.

PR #1533, opened 2022-09-14
States__solution

VovaDzhuryk:


vallerii:
Good job 👍

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [];
+  const copyState = { ...state };
```
OleksiiBlezniuk, line 11:
```suggestion
  const stateCopy = { ...state };
```

OleksiiBlezniuk: Just FYI - copyState - is an action, stateCopy - is a result of this action (copied state)

OleksiiBlezniuk (approved):
Create more meaningful commit messages, it’s hard to understand what was done in this commit from `States__solution` message.

PR #1535, opened 2022-09-14
Solution

ArtemHulkoDev:


barsuchellos:
GJ)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateObj = { ...state };
+  const actionsArr = [];
```
naziamoff, line 11: You should think of how you name your variables for making them maximally clear to understand. You should not add its data type to the variable, variable should describe what is stored inside
```suggestion
  const currentState = { ...state };
  const states = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
naziamoff, line 14: you should take a look at the checklist again

naziamoff (changes requested):

src/transformStateWithClones.js
ArtemHulkoDev: done)

src/transformStateWithClones.js
ArtemHulkoDev: done)

naziamoff (approved):
Great job 🚀

PR #1539, opened 2022-09-15
First commit

Bibicutu:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+
+      case 'clear':
+        for (const key in clone) {
+          delete clone[key];
+        }
+        history.push({ ...clone });
+        break;
+
+      default:
```
kamal-tufekcic, line 34: You can return or thrown an error here

kamal-tufekcic (approved):

ihor-jpeg (approved):

PR #1541, opened 2022-09-16
solution

WebsterManyanga:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        break;
+      case 'removeProperties':
+        for (let j = 0; j < (action.keysToRemove).length; j++) {
+          delete condition[action.keysToRemove[j]];
+        }
+        break;
+      default: for (const property in condition) {
+        delete condition[property];
+      }
+        break;
```
kamal-tufekcic, line 31: Not the default case, checklist # 3

kamal-tufekcic (changes requested):

ihor-jpeg (approved):

PR #1547, opened 2022-09-17
Solution

oleksusov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = {};
+
+  Object.assign(copyState, state);
```
SanyaBratashchuk, line 12:
```suggestion
  const copyState = {...state};
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+  const arrayOfStates = [];
+
+  for (const action of actions) {
+    switch (action['type']) {
```
SanyaBratashchuk, line 17:
```suggestion
    switch (action.type) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+      case 'removeProperties':
+        for (const property of action['keysToRemove']) {
+          delete copyState[property];
+        }
+        break;
+
+      case 'clear':
+        for (const key in copyState) {
+          delete copyState[key];
+        }
```
SanyaBratashchuk, line 31:
```suggestion
       copyState = {}
```

SanyaBratashchuk (changes requested):

src/transformStateWithClones.js
liliia-livandovskaya: you don't need for there, you can just set copyState = {}

liliia-livandovskaya (changes requested):
GJ! Lets work on changes

alexandra-protyanova (approved):

PR #1548, opened 2022-09-17
Solution Fixed

omerfarukbalik:


Bibicutu (approved):
OK from my point of view, good job!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+    if (action.type === 'clear') {
+      for (const key in tranformedObject) {
+        delete tranformedObject[key];
+      }
+    }
+
+    transformedState.push({ ...tranformedObject });
+  }
+
+  return transformedState;
```
ihor-jpeg, line 35: use switch

ihor-jpeg (changes requested):

PR #1549, opened 2022-09-17
solution

La-La-Op:


MAKSOti (changes requested):
GJ but read cheklist

BAD EXAMPLE:

const copy = { ...state }

GOOD EXAMPLE:

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const arr = [];
```
OleksiiBlezniuk, line 11: What is arr? Create more meaningful variables names

OleksiiBlezniuk: Also usage of data type in variable name - is a bad practice. It's a list of state copies, isn't it? stateList will be better here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+          delete stateCopy[key];
+        }
+        break;
+
+      case 'clear':
+        for (const property in stateCopy) {
+          delete stateCopy[property];
+        }
+        break;
+    }
```
OleksiiBlezniuk, line 32: switch/case should have default case, you can use it for handling cases when action type is different from these three

OleksiiBlezniuk (changes requested):

src/transformStateWithClones.js
La-La-Op: done

src/transformStateWithClones.js
La-La-Op: done

OleksiiBlezniuk (approved):
GJ!

PR #1555, opened 2022-09-19
Solution for Stateful clones

sapnachoudhary06:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,24 @@
...
+        delete currentState[keyToRemove];
+      }
+    } else if (action.type === 'clear') {
+      currentState = {};
+    }
+
+    allStates.push({ ...currentState });
+  }
+
+  return allStates;
```
kamal-tufekcic, line 28: Checklist # 2 & 3

kamal-tufekcic (changes requested):

src/transformStateWithClones.js
sapnachoudhary06: fixed.

ihor-jpeg (approved):

PR #1556, opened 2022-09-19
Solution

Deniso4ka23:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+        delete initialState[action.keysToRemove[i]];
+      }
+    }
+
+    if (action.type === 'clear') {
+      initialState = {};
+    }
+
+    allState.push({ ...initialState });
+  }
```
liliia-livandovskaya, line 29: rewrite your code using switch case instead of if else

PR #1557, opened 2022-09-19
Implement transformStateWithClones function

anastasiia-tilikina:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,55 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const versions = [];
+  const stateCopy = { ...state };
+  let version = {};
+
+  for (const action of actions) {
+    switch (true) {
```
alexandra-protyanova, line 15: it is not a good practice to use switch with true condition

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1560, opened 2022-09-20
Develop

Gintako666:


alexandra-protyanova (changes requested):
let us use switch case in this task

alexandra-protyanova (approved):

PR #1565, opened 2022-09-22
solution

vaadyak:


MaxManis (approved):
GJ

src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
...
   // write code here
+  const copy = { ...state };
+  const changesState = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          copy[key] = action.extraData[key];
+        }
```
naziamoff, line 19: It's better to use Object.assign here

src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
...
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete copy[key];
+        }
+        break;
+      case 'clear':
+        for (const key in copy) {
+          delete copy[key];
+        }
```
naziamoff, line 29: you can just set empty object for the copy in that case

naziamoff (approved):
Well done! But there are minor comments

PR #1574, opened 2022-09-29
commit

OleksChernikov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arr = [];
+  const copyState = { ...state };
+
+  for (const middleActions of actions) {
```
SanyaBratashchuk, line 13:
```suggestion
  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arr = [];
```
SanyaBratashchuk, line 10:
```suggestion
  const states = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      case 'removeProperties':
+        for (const elem of middleActions.keysToRemove) {
+          delete copyState[elem];
+        }
+        break;
+
+      case 'clear':
+        for (const key in copyState) {
+          delete copyState[key];
+        }
```
SanyaBratashchuk, line 28:
```suggestion
     copyState = {}
```

SanyaBratashchuk (changes requested):

src/transformStateWithClones.js
Bohdan-Maksymiuk: why didn't you fix this?
it will use less resources in this case

Bohdan-Maksymiuk (approved):

PR #1582, opened 2022-10-05
Solution

Julia2063:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,26 @@
...
+      for (const property of action.keysToRemove) {
+        delete copy[property];
+      } result.push({ ...copy });
+    } else {
+      for (const key in copy) {
+        delete copy[key];
+      }
+      result.push({ ...copy });
+    }
+  }
```
liliia-livandovskaya, line 27: let's use switch case there

liliia-livandovskaya (changes requested):
Lets work on changes

liliia-livandovskaya (approved):

PR #1588, opened 2022-10-07
commit

devWizHub:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (stateCopy[key]) {
```
amitpatiljc, line 23: no need to have if condition you can delete directly, if key is not present it wont delete anything and return object as it is

src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
...
+            delete stateCopy[key];
+          }
+        }
+        break;
+
```
amitpatiljc, line 28: we have 3rd case also `clear` condition

src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
...
+      default:
+        if (stateCopy) {
+          for (const key in stateCopy) {
+            delete stateCopy[key];
```
amitpatiljc, line 32: why if condition in default
need to return error in default

amitpatiljc (changes requested):

src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
...
+          if (stateCopy[key]) {
+            delete stateCopy[key];
+          }
+        }
+        break;
+
+      default:
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
```
amitpatiljc, line 32: there should be third case` clear` and then have `default` case , do not include case` clear` in default

amitpatiljc (changes requested):

src/transformStateWithClones.js
devWizHub: thanks

PR #1589, opened 2022-10-10
Task done

Vlad-Fedorishchev:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in copyMainArray) {
+        delete copyMainArray[key];
+      }
+    }
+    resultArr.push({ ...copyMainArray });
+  }
```
stanislav-topikha, line 30: switch/case

Vlad-Fedorishchev:
@stanislav-topikha fixes ready, can u check please?)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        for (const key in copyState) {
+          delete copyState[key];
+        }
+    }
+
+    resultArr.push({ ...copyState });
+  }
+
+  return resultArr;
 }
```
Anastasiia-Svintsova, line 41: You don't need to write a switch every time. Just write it once and iterate over the cases. You can see the example [here](https://www.w3schools.com/js/js_switch.asp)
```
switch(expression) {
  case x:
    // code block
    break;
  case y:
    // code block
    break;
  default:
    // code block
}
```

Anastasiia-Svintsova (changes requested):

Anastasiia-Svintsova (approved):
Good job

PR #1596, opened 2022-10-15
SOLUTION FINALLY;))

MarinaT-N:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,57 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const NewArray = [];
+  let massive = { ...state };
```
Oleksandra-Pasenchenko, line 11: Better use more informative names for variables, 'massive' means 'large' in English, you probably meant 'array' or something like that🙃

src/transformStateWithClones.js
```diff
@@ -7,7 +7,57 @@
...
+
+  for (let numArr = 0; numArr < actions.length; numArr++) {
```
Oleksandra-Pasenchenko, line 13: I recommend using `for...of` loop if you iterate over all elements of the array - it's faster

src/transformStateWithClones.js
```diff
@@ -7,7 +7,57 @@
...
+    const action = Object.assign({}, actions[numArr]);
+
+    for (const key in action) { // loop all types in array 'action'
+      if (action[key] === 'addProperties') {
+        const state1 = { ...massive };
```
Oleksandra-Pasenchenko, line 18: Why do you create this variable in each 'if'? You already have it at the beginning of the function body, no need to duplicate

src/transformStateWithClones.js
```diff
@@ -7,7 +7,57 @@
...
+        for (const x in state2) {
+          delete state2[x];
+        };
+
+        NewArray.push(state2);
+
+        for (const b in massive) {
+          delete massive[b];
+        }
+        massive = { ...state2 };
```
Oleksandra-Pasenchenko, line 39: Why is it needed? You need only to add properties in this 'if'. Please, revise your solution and remove unnecessary code from other 'if's - leave only what is needed for this specific 'if'

src/transformStateWithClones.js
```diff
@@ -7,7 +7,57 @@
...
+
+  for (let numArr = 0; numArr < actions.length; numArr++) {
+    const action = Object.assign({}, actions[numArr]);
+
+    for (const key in action) { // loop all types in array 'action'
+      if (action[key] === 'addProperties') {
+        const state1 = { ...massive };
+
+        Object.assign(state1, action.extraData);
+        NewArray.push(state1);
```
Oleksandra-Pasenchenko, line 21: You can push it once at the loop (after all 'if's)

Oleksandra-Pasenchenko (changes requested):

Oleksandra-Pasenchenko (changes requested):
Not all comments were fixed, I will write in your chat and provide more detailed feedback

PR #1598, opened 2022-10-16
first commit

haduigon:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
-  // write code here
+  const state2 = { ...state };
+  const log = [];
+
+  for (const action in actions) {
+    switch (actions[action].type) {
+      case 'addProperties':
+        Object.assign(state2, actions[action].extraData);
+
+        const log2 = { ...state2 };
```
TarasHoliuk, line 18: taking into account that you're anyway doing copy using `spread` operator, you can add here extra data as well, and don't do object.assign in the 16 row
in this case, you won't mutate the state object, so the 10 row will be unnecessarily too

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const state2 = { ...state };
+  const log = [];
+
+  for (const action in actions) {
+    switch (actions[action].type) {
```
TarasHoliuk, line 14: would be nice to create a variable for `actions[action]` - you're using it often. It will increase readability (especially in such cases as in the 25 row)

TarasHoliuk (changes requested):

src/transformStateWithClones.js
haduigon: done )

src/transformStateWithClones.js
haduigon: fixed mistakes but can`t avoid copying the state at the beginning it can lead to the mistakes in other scenarios

src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+    switch (action2.type) {
+      case 'addProperties':
+
+        logCopy = {
+          ...logCopy, ...action2.extraData,
+        };
+
+        const addLog = { ...logCopy };
+
+        log.push(addLog);
```
AlisaVlasova, line 25:
```suggestion
        logCopy = {
          ...logCopy, ...action2.extraData,
        };

        log.push(logCopy);
```

AlisaVlasova (changes requested):

liliia-livandovskaya (approved):

PR #1606, opened 2022-10-19
Solution

iryna-lelyk:


Illia-Skladnik (approved):
Looks perfect.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        break;
+
+      case 'clear':
+        Object.keys(stateCopy).forEach(key => delete stateCopy[key]);
+        break;
+
+      default:
+        break;
+    }
+    arrWithCopies.push({ ...stateCopy });
```
mykola-koval, line 32: you have already copied it before, no need to do it here i guess
```suggestion
    arrWithCopies.push(stateCopy);
```

mykola-koval (approved):

src/transformStateWithClones.js
iryna-lelyk: If I remove here spread operator, I have errors in 2 tests that check on multiple properties (removeProperties, then addProperties, etc)

PR #1610, opened 2022-10-20
stateClonesDK

Kukharuk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+        break;
+
+      case 'clear':
+        resultObj = {};
+
+        const resultObj4 = { ...resultObj };
+
+        result.push(resultObj4);
+
+        break;
```
polosanya, line 41: Don't forget to add default case for your switch

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let resultObj = { ...state };
```
polosanya, line 11: It is better not to use `result` everywhere in your variable's name until it's really result

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+
+        const resultObj3 = { ...resultObj };
+
+        result.push(resultObj3);
+        break;
+
+      case 'clear':
+        resultObj = {};
+
+        const resultObj4 = { ...resultObj };
```
polosanya, line 37: I think you can create better names than resultObj1, 2, 3, 4

polosanya (changes requested):
Good job! Please check my comments

polosanya (approved):
Good job!

PR #1621, opened 2022-10-20
solution

andrii-bassarab:


Svitlana-Yudina (approved):
It's a very interesting solution) GJ!

Olexandr6 (changes requested):
Add default `block` to the switch statement

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        clones.push(lastObj);
+        break;
+
+      case 'clear':
+        clones.push({});
+        break;
+    }
+  }
+
+  clones.shift();
```
Olexandr6, line 35: Just create a one more variable for the initial state and you don't need `shift()`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        const lastObj = { ...clones[clones.length - 1] };
+
+        for (const key of action.keysToRemove) {
+          delete lastObj[key];
+        }
+        clones.push(lastObj);
+        break;
+
+      case 'clear':
+        clones.push({});
```
Olexandr6, line 30: You can push changes once after `switch` statement, not in each `case`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clones = [ { ...state } ];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        const copy = { ...clones[clones.length - 1] };
```
Olexandr6, line 15: I would better to create a separate variable for the current state

Olexandr6 (changes requested):
Good job! Let's work on improvements:

denys-danyliuk (approved):

PR #1624, opened 2022-10-21
Solution

bohdan-stepanets:


Stepico (approved):

Stepico:
GJ, just maybe one piece of advice, I guess it would be better to make an empty line after for statement in your switch

mykola-koval (approved):

PR #1632, opened 2022-10-22
solution

webdevw:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      case 'removeProperties':
+        for (const remKeys of keysToRemove) {
+          delete clone[remKeys];
+        }
+        break;
+
+      case 'clear':
+        for (const remProp in clone) {
+          delete clone[remProp];
+        }
```
ivanchuk-yurii, line 30: you can do just `clone = {}`

ivanchuk-yurii, line 30: you can do just `clone = {}`

ivanchuk-yurii (approved):

PR #1634, opened 2022-10-22
Stateful clones solution

karyna-salnykova:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+
+        break;
+
+      case 'clear':
+        Object.keys(cloneState).forEach(key => delete cloneState[key]);
+        break;
+
+      default:
+        break;
+    }
```
Svitlana-Yudina, line 32: I think, you should add some act to default. `return 'Error';` or `throw new Error('actions is not correct');`

Svitlana-Yudina (approved):
GJ!))

denys-danyliuk (changes requested):

illiarerun (approved):

PR #1640, opened 2022-10-23
solution

Anatolii-Voloshyn:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
+  const result = [];
```
Largoxoxo, line 11: Bad variable name. It is never a good idea to use result. Although checklist doesn't explicitly state that 'result' is a bad variable name. It is presumed that after 'Situs inversus' we should not use such variable names. Better to change this one.

Largoxoxo (changes requested):

src/transformStateWithClones.js
Anatolii-Voloshyn: Thanks! Agree with you. It's a bad variable name)

Largoxoxo (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
```
denys-danyliuk, line 10:
```suggestion
  const stateCopy = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+  const resultState = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(copyState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const property of action.keysToRemove) {
```
denys-danyliuk, line 20:
```suggestion
        for (const key of action.keysToRemove) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
+  const resultState = [];
```
denys-danyliuk, line 11:
```suggestion
  const transformedState = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        break;
+
+      case 'clear':
+        for (const key in copyState) {
+          delete copyState[key];
+        };
+        break;
+
+      default:
+        throw new Error('Wrong type of transformState');
```
denys-danyliuk, line 32:
```suggestion
        throw new Error('Wrong type of action');
```

denys-danyliuk (approved):

denys-danyliuk (changes requested):
👆

PR #1645, opened 2022-10-23
Solution

alpopkov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+
+        result.push({ ...stateCopy });
+        break;
+
+      default:
+        break;
```
kryvets-vadym, line 38: here you need to show the error using `throw`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
-  // write code here
+  const result = [];
+  const stateCopy = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+
+        result.push({ ...stateCopy });
```
kryvets-vadym, line 18: you can take it out of the switch and do it once

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
kryvets-vadym, line 10: use more detailed variable names

kryvets-vadym (approved):
hope my comments help you, good luck ))

src/transformStateWithClones.js
denys-danyliuk: agree

src/transformStateWithClones.js
denys-danyliuk: good point, 4 rule in checklist

src/transformStateWithClones.js
denys-danyliuk: use 'throw new Error', read more here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw

denys-danyliuk (changes requested):

PR #1646, opened 2022-10-24
Solution

MikhailKucherina:


shtht24 (approved):
nice solution :)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
```
denys-danyliuk, line 10:
```suggestion
  const stateCopy = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+  const allStates = [];
```
denys-danyliuk, line 11:
```suggestion
  const transformedStates = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const property of action.keysToRemove) {
+          delete newState[property];
+        }
+        break;
+
+      case 'clear':
+        Object.keys(newState).forEach(key => delete newState[key]);
```
denys-danyliuk, line 26: it would be better to use 'for in' here in my opinion

denys-danyliuk (approved):

src/transformStateWithClones.js
denys-danyliuk: 5 point in checklist

denys-danyliuk (changes requested):

mykola-koval (approved):

PR #1653, opened 2022-10-25
Clones stateful

Oleh-Burii:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let copyState = { ...state };
+  const newState = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```
liliia-livandovskaya, line 15: lets add enum and use it in cases

liliia-livandovskaya (changes requested):
-

PR #1662, opened 2022-10-31
test ok

diolna:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      arrayObject.push({ ...updateObject });
+    } else if (action.type === 'removeProperties') {
+      for (const keyRemove of action.keysToRemove) {
+        delete updateObject[keyRemove];
+      }
+      arrayObject.push({ ...updateObject });
+    } else if (action.type === 'clear') {
+      for (const key in updateObject) {
+        delete updateObject[key];
+      }
```
Bohdan-Maksymiuk, line 25: avoid such nesting `else if` conditions
This code is hard to read and maintain

Bohdan-Maksymiuk: use `switch case` for such predefined cases

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+
+      const time = { ...updateObject };
```
Bohdan-Maksymiuk, line 27: why time? use meaningful naming

Bohdan-Maksymiuk (changes requested):

src/transformStateWithClones.js
diolna: fixed

src/transformStateWithClones.js
diolna: Fixed. Removed temporary variable.

PR #1673, opened 2022-11-05
Added the solution

zbyn3k:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,38 @@
...
+
+      cloneHistory.push(Object.assign({}, newClone));
+    }
+
+    if (action.type === 'clear') {
+      for (const key in newClone) {
+        delete newClone[key];
+      }
+
+      cloneHistory.push(Object.assign({}, newClone));
```
amitpatiljc, line 36: use of spread operator could be better here !

amitpatiljc (changes requested):
Let's use `switch` case here instead of `if` statements, as we have limited conditions to deal with

src/transformStateWithClones.js
zbyn3k: Hi Amit, swapped the Object.assign with spread operators and replaced if-statements with switch-case. Let me know if it looks good. Thanks!

amitpatiljc (changes requested):
Here also add default case and all good

zbyn3k:
Should be good! Thanks!

amitpatiljc (approved):
Nice !!

PR #1674, opened 2022-11-05
newState

Mykola-Hadupiak:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let clone = { ...state };
```
SanyaBratashchuk, line 10:
```suggestion
  let stateClone = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+  const arr = [];
```
SanyaBratashchuk, line 11:
```suggestion
  const copies = [];
```

SanyaBratashchuk (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,45 @@
...
+
+      case ACTIONS.clear:
+        deleteObjectKeys(Object.keys(stateCopy), stateCopy);
+        stateChainAdder();
+        break;
+
+      default:
+        return null;
+    }
+  }
```
etojeDenys, line 46: you can call stateChainAdder only one time after switch/case(after all cases)

etojeDenys (changes requested):
nit:

nazarmatsevych (approved):
👍

PR #1676, opened 2022-11-08
solution

Tsvigunova:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+        object[key] = obj.extraData[key];
+      }
+    } else if (obj.type === 'removeProperties') {
+      for (const key of obj.keysToRemove) {
+        delete object[key];
+      }
+    } else if (obj.type === 'clear') {
+      for (const key in object) {
+        delete object[key];
+      }
```
liliia-livandovskaya, line 27: lets use `switch case` instead of `if else`

liliia-livandovskaya (changes requested):
GJ! Pay attention to the comments

liliia-livandovskaya (approved):
Great job!

PR #1678, opened 2022-11-09
Solution 1.0

aliliolek:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,30 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const newState = { ...state };
+  const arrayOfObjects = [];
+
+  for (let i = 0; i < actions.length; i++) {
+    const oneElement = actions[i];
+
+    if (oneElement.type === 'addProperties') {
```
loralevitska, line 17: you compare `oneElement.type` in all `if` statement. It's better use here `switch/case`

loralevitska (changes requested):

src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
...
+
+        for (let m = 0; m < keysRemoving.length; m++) {
+          delete newState[`${keysRemoving[m]}`];
+        }
+        break;
+      case 'clear':
+        for (const keyA in newState) {
+          delete newState[keyA];
+        }
+        break;
```
polosanya, line 32: Don't forget to add default case

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comment

PR #1680, opened 2022-11-10
Solution

petrolozynskyi:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      }
+      newArr.push({ ...newState });
+    }
+
+    if (action.type === 'clear') {
+      for (const stat in newState) {
+        delete newState[stat];
+      }
+      newArr.push({ ...newState });
+    }
```
liliia-livandovskaya, line 33: Lets use `switch case` instead of `if`

liliia-livandovskaya (changes requested):
Great job! Pay attention to the comments

PR #1684, opened 2022-11-12
Solution

VadymKim:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const returnedArray = [];
+  const cloneState = { ...state };
+  // returnedArray.push( {...state });
+
+  for (const obj of actions) {
+    if (obj.type === 'addProperties') {
```
Oleksandra-Pasenchenko, line 15: Use 'switch case' instead of nested 'if's

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+      Object.assign(cloneState, obj.extraData);
+    } else if (obj.type === 'removeProperties') {
+      for (const i of obj.keysToRemove) {
```
Oleksandra-Pasenchenko, line 18:
```suggestion
      for (const key of obj.keysToRemove) {
```
Make the variable names more informative (also think of a better name for `returnedArray`)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const returnedArray = [];
+  const cloneState = { ...state };
+  // returnedArray.push( {...state });
```
Oleksandra-Pasenchenko, line 12: Remove the comment

Oleksandra-Pasenchenko (changes requested):

VadymKim:
Hi, could you check out my solution after fixes, please.

PR #1688, opened 2022-11-14
Solution

OksanaBaloh:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,27 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  let stateCopy = { ...state };
+  const result = [];
+
+  for (const obj of actions) {
```
Bohdan-Maksymiuk, line 14:
```suggestion
  for (const action of actions) {
```
use descriptive names

Bohdan-Maksymiuk: also use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) for usage of type, extraData, keysToRemove

Bohdan-Maksymiuk (changes requested):
Good job, lets improve our code style

SanyaBratashchuk (approved):

PR #1690, opened 2022-11-14
solution

Maks-Tsarenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        stateArray.push({ ...stateCopy });
+
+        break;
+
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        stateArray.push({ ...stateCopy });
```
Bohdan-Maksymiuk, line 25: you pushed at the end of each case
mb we can write push ones after the switch?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const stateArray = [];
+
+  for (const action of actions) {
+    switch (action.type) {
```
Bohdan-Maksymiuk, line 14: try to use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) from action

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        stateArray.push({ ...stateCopy });
+
+        break;
+
+      case 'removeProperties':
+
+        for (const keys of action.keysToRemove) {
```
Bohdan-Maksymiuk, line 31:
```suggestion
      case 'removeProperties':
        for (const keys of action.keysToRemove) {
```

Bohdan-Maksymiuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+
+        break;
+
+      case 'removeProperties':
+        for (const keys of action.keysToRemove) {
+          delete stateCopy[keys];
+        }
```
polosanya, line 32: Don't forget to add a default case in case your type doesn't match any case

polosanya (changes requested):
Good job! I'm approving it, but please pay attention to the comment

src/transformStateWithClones.js
Maks-Tsarenko: thanks, but you forget to approve) according to the terms of the task, we have only 3 types of case ( add, remove and clear)

SanyaBratashchuk (approved):

PR #1691, opened 2022-11-15
Initial commit

artem-hirzhev:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      case 'removeProperties':
+        for (const key in keysToRemove) {
+          delete nextState[keysToRemove[key]];
+        }
+        break;
+
+      case 'clear':
+        for (const clear in nextState) {
+          delete nextState[clear];
+        }
```
polosanya, line 30: What if type will not match any of cases?
Don't forget to add default case.

polosanya (changes requested):
Good job! Please check my comment and fix it

SanyaBratashchuk (approved):

PR #1693, opened 2022-11-16
solution

Fynkzia:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
+      default:
+        break;
+    }
+
+    stateCopyArray.push(stateCopy);
+    stateCopy = { ...stateCopyArray[stateCopyArray.length - 1] };
```
Oleksandr-Filo, line 33: you can write it simpler
```suggestion
    stateCopyArray.push({ ...stateCopy });
```

Oleksandr-Filo (approved):
Looks good, I left one suggestion

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const stateCopyArray = [];
```
solaryasha, line 11: Avoid using data structure names in variable names, it's really confusing

```suggestion
  const stateCopies = [];
```

solaryasha (approved):

PR #1694, opened 2022-11-16
solution

kateryzhen:


tetiana-ivanchenko (changes requested):
use 'default' in the switch for error handling

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let state2 = { ...state };
+  const stateArr = [];
+
+  for (const actionObj of actions) {
```
tetiana-ivanchenko, line 13:
```suggestion
  for (const action of actions) {
```

PR #1695, opened 2022-11-16
Job Done!

IoghiStudio:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const currentState = { ...state };
+
+  for (const obj of actions) {
+    if (obj.type === 'addProperties') {
+      Object.assign(currentState, obj.extraData);
```
mgrinko, line 15: you mutate one initial cloned object all the time
It is better to create a copy on each iteration

mgrinko (approved):
Please finish your landing page BEFORE proceeding with Javascript Advanced!!!

mgrinko (changes requested):
Please fix my previous comment

src/transformStateWithClones.js
IoghiStudio: Thanks for advice!

IoghiStudio: I don't understand.

mgrinko: You created a copy before the loop
and modify it on every action
But it is better to create a copy for each action 
I'll show you tomorrow

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        }
+      }
+    }
+
+    if (obj.type === 'clear') {
+      if (Object.keys(currentState).length !== 0) {
+        for (const key in currentState) {
+          delete currentState[key];
+        }
+      }
```
mgrinko, line 31: If you need to get an empty object
Why do you copy the props?

mgrinko (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+  const currentState = { ...state };
+
+  for (const obj of actions) {
+    if (obj.type === 'addProperties') {
+      Object.assign(currentState, obj.extraData);
+    }
+
+    if (obj.type === 'removeProperties') {
+      for (const key of obj.keysToRemove) {
+        if (currentState.hasOwnProperty(key)) {
```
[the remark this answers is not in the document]
IoghiStudio: That's right!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+      Object.assign(currentState, action.extraData);
+    }
+
+    if (type === 'removeProperties') {
+      action.keysToRemove.forEach(key => delete currentState[key]);
+    }
+
+    if (type === 'clear') {
+      currentState = {};
+    }
```
meforoma, line 26: better use a `switch..case` construct that provides you with a `default` case, where you can provide an explanatory fallback, like
```
```

meforoma (changes requested):
see the comment about `switch..case`

PR #1697, opened 2022-11-17
Solution

IrynaK-a:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const newState = { ...state };
+
+  for (const obj of actions) {
```
loralevitska, line 13: try to use more meaningful naming so that you can understand what is meant `obj.extraData` (action.extraData) or `obj.type` (action.type). 
```suggestion
  for (const action of actions) {
```

loralevitska (changes requested):
great job! let's improve it

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const newState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 14: You can do destructuring before and take type directly

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        action.keysToRemove.every((element) => delete newState[element]);
+        result.push({ ...newState });
+        break;
+
+      case 'clear':
+        for (const key in newState) {
+          delete newState[key];
+        }
+
+        result.push({ ...newState });
```
polosanya, line 30: You do push for each case. Just move it out of the switch...case and do it after all iterations

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        break;
```
polosanya, line 31: Add a default case to avoid situations when type doesn't match any of the cases

polosanya (changes requested):
Almost perfect! Please check the comments

loralevitska (approved):
good job!

PR #1700, opened 2022-11-19
done

Aleksey-Aleksey:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const copyState = {};
+
+  for (const key in state) {
+    copyState[key] = state[key];
+  }
```
TarasHoliuk, line 15: If you want to copy the whole object it's better to use `rest` operator or `Object.assign`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
TarasHoliuk, line 18: the same comments as for the previous task: use destructuring and `switch`

TarasHoliuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const remove of action.keysToRemove) {
+          delete copyState[remove];
+        }
+        break;
 
+      case 'clear':
+        for (const el in copyState) {
```
tetiana-ivanchenko, line 26: work on your variables naming, next time I will reject work with this naming

tetiana-ivanchenko (approved):

PR #1705, opened 2022-11-21
solution

OlggaJul:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+      case 'removeProperties':
+        for (const toRemove of propertiesToDelete) {
+          delete stateCopy[toRemove];
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
```
official-artem, line 32: Maybe you can just write stateCopy = {}; In this case, you don't need to use loop and use more memory.

official-artem (approved):
Good job

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let stateCopy = { ...state };
```
solaryasha, line 10: Avoid reassigning variables, it decreases code [readability](https://blog.sapegin.me/all/avoid-reassigning-variables/)

solaryasha (changes requested):

PR #1707, opened 2022-11-21
solution

ysrckr:


nurdanokcu (approved):
Superb.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateClone, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateClone[key];
```
mgrinko, line 21: You mutate one initial object all the time
Much better is to create a copy and mutate it on each action

mgrinko (changes requested):

src/transformStateWithClones.js
ysrckr: I don't modify the initial state. When I create a new copy on each action. 
`Should handle multiple types 
Should handle a long list of types `
These two tests don't pass.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+  const result = [];
+  let stateClone = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        stateClone
+        = {
+            ...stateClone, ...action.extraData,
+          };
```
meforoma, line 19: in this approach - formatting:
```suggestion
        stateClone = {
          ...stateClone,
          ...action.extraData,
        };
```

but i'd suggest to not re-assign stateClone, but rather use 
`Object.assign(stateClone, action.extraData)`

and then you would declare it with `const`: `const stateClone = { ...state };`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          stateClone = { ...stateClone };
```
meforoma, line 24: hot sure it's needed

src/transformStateWithClones.js
meforoma: but yeah, for `clear` it's much easier to keep it as `let` to simply assign an empty object

meforoma (approved):

PR #1708, opened 2022-11-21
Solution

nurdanokcu:


ysrckr (approved):
Same as the previous one. All of the if statements are checks equality of action.type, so as suggested in checklist you could have used switch statement. Also clone variable name is understandable in this task because it is very short. But in general it would be good to use something like stateClone.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+  const stateVersions = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateClone, action.extraData);
+        break;
+
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => delete stateClone[key]);
```
mgrinko, line 20: It is better to copy the state each time
than mutate one initial cloned object

mgrinko (changes requested):

meforoma (approved):

PR #1711, opened 2022-11-21
Solution

romanblyashuk94:


superpooperxxx (approved):
Looks good to me!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,60 @@
...
+
+  function removeProperties(object, keysToDelete) {
+    const newObj = { ...object };
+
+    keysToDelete.forEach(key => {
+      delete newObj[key];
+    });
+
+    return newObj;
+  }
```
ivan-kadykalo, line 63: You can write operations inside the switch case, without helping functions.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,60 @@
...
+  let clones = [state];
+
+  actions.forEach(handleAction);
+
+  clones.shift();
+
+  return clones;
+
+  function handleAction({ type, extraData, keysToRemove }) {
+    const lastState = clones[clones.length - 1];
```
ivan-kadykalo, line 19: If you want to use helping functions, make them pure. 
The pure function should take some parameters and return the result, which you should use later. So I think there is no reason to use the helping function in this task.

ivan-kadykalo (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,47 @@
...
+
+  actions.forEach(({ type, extraData, keysToRemove }) => {
+    const lastState = clones[clones.length - 1];
+
+    switch (type) {
+      case 'addProperties':
+        clones = [...clones, {
+          ...lastState,
+          ...extraData,
+        }];
```
solaryasha, line 20: It's a lot of reassigning, why don't you just use `push` method ?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,47 @@
...
+        const newState = { ...lastState };
+
+        keysToRemove.forEach(key => {
+          delete newState[key];
+        });
+
+        clones = [
+          ...clones,
+          newState,
+        ];
```
solaryasha, line 33: Again, a lot of reassigning, we could it with the `push`
Also, you could move this code from `switch` instruction

src/transformStateWithClones.js
```diff
@@ -7,7 +7,47 @@
...
+          {},
+        ];
+        break;
+
+      default:
+        throw new Error('Wrong action type!');
+    }
+  });
+
+  clones.shift();
```
solaryasha, line 48: Really, really strange. Why do we need to create an array with one item and then remove it?

solaryasha (changes requested):

valerii-derkach (approved):

PR #1712, opened 2022-11-22
Solution

UmizDemud:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
-  // write code here
+  const result = [];
+  const backupObj = Object.assign({}, state);
+
+  actions.forEach(action => {
+    switch (action.type) {
+      case 'addProperties':
+        for (const element of Object.entries(action.extraData)) {
+          backupObj[element[0]] = element[1];
+        }
```
mgrinko, line 18: Why don't you use `Object.assign` here?
It will simplify the code

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (key in backupObj) {
```
mgrinko, line 22: you don't need this check

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+            delete backupObj[key];
+          }
+        }
+        break;
+
+      case 'clear':
+        for (const key in backupObj) {
+          delete backupObj[key];
+        }
```
mgrinko, line 31: why not just set `{}` instead of deleting props?

mgrinko (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let backupObj = Object.assign({}, state);
```
meforoma, line 11: spread operator would make it even more concise 🙂 
```suggestion
  let backupObj = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+
+  actions.forEach(action => {
+    switch (action.type) {
+      case 'addProperties':
+        backupObj = { ...backupObj };
```
meforoma, line 16: do you really need to assign it on each case?
i don't think so - you just mutate it on each `case`, but then the only place you need a copy of it - is on `.push` - see the next comment

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+        for (const key of action.keysToRemove) {
+          delete backupObj[key];
+        }
+        break;
+
+      case 'clear':
+        backupObj = {};
+    }
+
+    result.push(backupObj);
```
meforoma, line 31: so you don't need `backupObj = { ...backupObj };` on each case, but here you do this:
```suggestion
    result.push({ ...backupObj });
```
-> create a copy of the obj on-fly and push it

meforoma (changes requested):
see few comments that will improve the code

src/transformStateWithClones.js
UmizDemud: Very true

src/transformStateWithClones.js
UmizDemud: Might as well :D applying

src/transformStateWithClones.js
UmizDemud: Yes definitelly

naziamoff (approved):
Great job! 🔥

PR #1715, opened 2022-11-23
Solution

halynarydasova:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
...
+        action.keysToRemove.forEach((key) => delete stateCopy[key]);
+
+        break;
+      }
+
+      case 'clear': {
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        };
+        break;
```
qoa11a, line 31:
```suggestion
        Object.assign(stateCopy, action.extraData);
        break;

      case 'removeProperties': {
        action.keysToRemove.forEach((key) => delete stateCopy[key]);
        break;
      }

      case 'clear': {
        for (const key in stateCopy) {
          delete stateCopy[key];
        }
        break;
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const resultState = [];
```
qoa11a, line 11: It's not exactly state. Better name it stateHistory or something like this.

qoa11a (changes requested):
Good job!
Few comments.

jstmpelowycz (approved):
LGTM

PR #1717, opened 2022-11-23
Solution

Artem193:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+          delete stateClone[value];
+        }
+      }
+      result.push({ ...stateClone });
+    }
 
+    if (actions[i].type === 'clear') {
+      result.push({});
+      stateClone = {};
+    }
```
liliia-livandovskaya, line 31: use switch case instead of ifs

liliia-livandovskaya (changes requested):
Pay attention to the comments

src/transformStateWithClones.js
Anastasiia-Svintsova: Not fixed

Anastasiia-Svintsova (changes requested):
Please, fix the previous comment before requesting a review again. If you have any questions feel free to ask them in the fe_chat

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  const result = [];
+  let stateClone = { ...state };
+
+  for (let i = 0; i <= actions.length - 1; i++) {
```
ivanchuk-yurii, line 13: it's better to use `for of` loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+    switch (actions[i].type) {
+      case 'addProperties':
+        Object.assign(stateClone, actions[i].extraData);
+        result.push({ ...stateClone });
```
ivanchuk-yurii, line 17: this line is the same for every `case`, you can move it from `switch`

ivanchuk-yurii (approved):

PR #1718, opened 2022-11-23
solution

lefteris-L:


CharAnagn (approved):
GG 👍

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+      case 'clear':
+        for (const key in stateClone) {
+          delete stateClone[key];
+        }
+        break;
+
+      default: break;
+    }
+    modifiedStates.push({ ...stateClone });
```
mgrinko, line 35: You could copy before modifying 
instead of mutating the copy created at the beginning

mgrinko (approved):

src/transformStateWithClones.js
lefteris-L: @mgrinko You mean leave the object empty and copy inside the loop?

PR #1723, opened 2022-11-24
Solution

Iksich:


UmizDemud (approved):
Looks correct!

src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const copyState = { ...state };
+  const resultArr = [];
+
+  for (const action of actions) {
```
mgrinko, line 14: it is better to create a copy on each iteration than modify all the time an object copied at the beginning

mgrinko (approved):

PR #1727, opened 2022-11-24
Solution

Rijo-J:


sercansimsek (approved):
Nicely done!!!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+        break;
+      case 'removeProperties':
+        stateCopy = { ...stateCopy };
+        obj.keysToRemove.forEach(key => delete stateCopy[key]);
+        break;
+      case 'clear':
+        stateCopy = { };
+        break;
+      default:
+        break;
```
meforoma, line 27: don't leave `default:` blank - provide an explanatory fallback, like
```
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateHistory = [];
+  let stateCopy = { ...state };
+
+  for (const obj of actions) {
```
meforoma, line 13: just naming-wise : 
```suggestion
  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+        stateCopy = { ...stateCopy };
+        obj.keysToRemove.forEach(key => delete stateCopy[key]);
+        break;
+      case 'clear':
+        stateCopy = { };
+        break;
+      default:
+        break;
+    }
+    stateHistory.push(stateCopy);
```
meforoma, line 29: you can create a copy when pushing it:
```suggestion
    stateHistory.push({ ...stateCopy });
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+  let stateCopy = { ...state };
+
+  for (const obj of actions) {
+    switch (obj.type) {
+      case 'addProperties':
+        stateCopy = { ...stateCopy };
+        Object.assign(stateCopy, obj.extraData);
+        break;
+      case 'removeProperties':
+        stateCopy = { ...stateCopy };
```
meforoma, line 20: if making a copy on `.push` - no need to copy on each `case`

meforoma (changes requested):
see few comments

src/transformStateWithClones.js
Rijo-J: ✅

src/transformStateWithClones.js
Rijo-J: ✅

src/transformStateWithClones.js
Rijo-J: ✅

src/transformStateWithClones.js
Rijo-J: ✅

meforoma (approved):
🎉

PR #1728, opened 2022-11-24
Solution

sercansimsek:


Rijo-J (approved):
Looks good to me! 🚀

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+          delete stateCopy[key];
+        }
+        break;
+
+      case 'clear':
+        stateCopy = {};
+        break;
+
+      default:
+        break;
```
meforoma, line 33: don't leave `default:` blank - provide an explanatory fallback, like
```
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+    }
+    result.push(stateCopy);
```
meforoma, line 35: you can create a static state right here instead of `stateCopy = { ...stateCopy };` on each case:

```suggestion
    result.push({ ...stateCopy });
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+  for (const impact of actions) {
+    switch (impact.type) {
+      case 'addProperties':
+        stateCopy = { ...stateCopy };
+        Object.assign(stateCopy, impact['extraData']);
+        break;
+
+      case 'removeProperties':
+        stateCopy = { ...stateCopy };
```
meforoma, line 21: you don't really need it if you make a copy on `.push` stage
```suggestion
```

meforoma (changes requested):
please see the optimisation comments

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+  const stateCopy = { ...state };
+
+  for (const impact of actions) {
+    switch (impact.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, impact.extraData);
+        break;
+
+      case 'removeProperties':
+
```
meforoma, line 20:
```suggestion
```

meforoma (approved):
nnice 👍

PR #1729, opened 2022-11-24
solution

lukacsja:


mmahfuzi (approved):
Nice job!

src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
...
+        break;
+
+      case 'clear':
+        stateCopy = { ...stateCopy };
+
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
+    }
```
meforoma, line 36: switch requires a `default:` case as a fallback option for cases not covered
usually it's `throw new Error('error explanation')`

meforoma (approved):
good one.
just get used to provide a `default:` case for a switch construct

src/transformStateWithClones.js
lukacsja: ill keep in mind, thank you

PR #1730, opened 2022-11-24
solution

mmahfuzi:


Mohameddahi (approved):
well done !

mmahfuzi:
> well done !

Thanks :)

src/transformStateWithClones.js
```diff
@@ -8,6 +8,43 @@
...
+        for (const key in newState) {
+          delete newState[key];
+        }
+        resArray.push(newState);
+        break;
+
+      default:
+        throw new Error('action is not supported');
+    }
+  }
```
meforoma, line 45: you could make it shorter
```suggestion
  for (const action of actions) {
    switch (action['type']) {
      case 'addProperties':
        Object.assign(newState, action['extraData']);
        break;

      case 'removeProperties':
        for (const key of action['keysToRemove']) {
          delete newState[key];
        }
        break;

      case 'clear':
        for (const key in newState) {
          delete newState[key];
        }
        break;

      default:
        throw new Error('action is not supported');
    }

    resArray.push({ ...newState }); // pushing a copy of a current state after mutations above
  }
```

meforoma (approved):
it's a working solution 👍
, yet can be more concise - see the comment

PR #1733, opened 2022-11-25
Solution

Grayfox147:


mmahfuzi (approved):
Well done!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+  const clon = [];
+  let newState = {
+    ...state,
+  };
+
+  for (const action of actions) {
+    newState = {
+      ...newState,
+    };
+
```
meforoma, line 19: you could make this copy after the switch - during `.push` - see below

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+      case 'clear':
+        for (const key in newState) {
+          delete newState[key];
+        }
+        break;
+
+      default:
+        throw new Error('action is not supported');
+    }
+    clon.push(newState);
```
meforoma, line 40:
```suggestion
    clon.push({ ...newState });
```

src/transformStateWithClones.js
meforoma, line 19: you could make this copy after the switch - during `.push` - see below
if you, you won't need this here
```suggestion
```

meforoma (approved):
good.
check out the comment i left you.

PR #1735, opened 2022-11-25
Solution

Mohameddahi:


mmahfuzi (approved):
Good job! just one small comment, it's better not to keep default empty - you can add this: throw new Error('action is not supported')

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        stateClone = { ...stateClone };
+        action.keysToRemove.forEach(key => delete stateClone[key]);
+        break;
+
+      case 'clear':
+        stateClone = {};
+        break;
+
+      default:
+        break;
```
meforoma, line 30: @mmahfuzi is right, better provide an explicit fallback for `default` case

meforoma (approved):

PR #1736, opened 2022-11-25
Created Clone Objects, Transformed States

Kavster92:


mmahfuzi (approved):
Well done! just a small suggestion,  it's better not to keep the default, empty. you can add: throw new Error('action is not supported') to default.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateVersions = [];
+  let stClone = { ...state };
+
+  for (const order of actions) {
+    switch (order.type) {
+      case 'addProperties':
+        stClone = { ...stClone };
```
naziamoff, line 16: why do you need to copy the copy?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateVersions = [];
+  let stClone = { ...state };
```
naziamoff, line 11: better to come up with more descriptive naming, it is now unclear what `stClone` means

naziamoff (approved):
Gj!
I have but two small comments

src/transformStateWithClones.js
Kavster92: When removed, the tests which apply to multiple objects fail repeatedly:
image

I am under the assumption that the obj needs to be recreated inside the for loop each time for a very specific reason that I don't yet understand. I spent about an hour looking for what it might be... this is the log:

 FAIL  src/transformStateWithClones.test.js
  √ Should create a new object with a single added property (3ms)
  √ Should create a new object with multiple added properties
  √ Should combine old properties with added ones (1ms)
  √ Should use the latest value when adding an existent property
  √ Should create an empty object when removing the last property
  √ Should create an object without removed properties (5ms)
  √ Should create the same state when removing no properties (1ms)
  √ Should not fail when removing not existing property (1ms)
  √ Should create an empty object after clear (1ms)
  √ Should not fails when calling clear for an empty state
  × Should handle multiple types (6ms)
  × Should handle a long list of types (5ms)

  ● Should handle multiple types

    expect(received).toEqual(expected) // deep equality

    - Expected
    + Received

    @@ -1,13 +1,13 @@
      Array [
        Object {
    -     "bar": "foo",
    +     "another": "one",
          "foo": "bar",
    -     "hello": "world",
          "name": "Jim",
        },
        Object {
    +     "another": "one",
          "foo": "bar",
          "name": "Jim",
        },
        Object {
          "another": "one",

      214 |     },
      215 |   ]))
    > 216 |     .toEqual([
          |      ^
      217 |       {
      218 |         foo: 'bar', bar: 'foo', name: 'Jim', hello: 'world',
      219 |       },

      at Object.<anonymous> (src/transformStateWithClones.test.js:216:6)

  ● Should handle a long list of types

    expect(received).toEqual(expected) // deep equality

    - Expected
    + Received

    @@ -3,18 +3,21 @@
          "foo": "bar",
          "name": "Jim",
        },
        Object {},
        Object {},
    -   Object {},
    +   Object {
    +     "yet": "another property",
    +   },
        Object {
          "yet": "another property",
        },
    -   Object {},
    +   Object {
    +     "foo": "bar",
    +   },
        Object {
          "foo": "bar",
    -     "name": "Jim",
        },
        Object {
          "foo": "bar",
        },
      ]

      258 |     },
      259 |   ]))
    > 260 |     .toEqual([
          |      ^
      261 |       {
      262 |         foo: 'bar', name: 'Jim',
      263 |       },

      at Object.<anonymous> (src/transformStateWithClones.test.js:260:6)

Test Suites: 1 failed, 1 total
Tests:       2 failed, 10 passed, 12 total
Snapshots:   0 total
Time:        3.076s

PR #1737, opened 2022-11-25
solution

DimaKolesnyk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,45 @@
...
-  // write code here
+  const arrayOfClones = [];
+  const stateClone = { ...state };
+
+  for (const action of actions) {
+    switch (true) {
+      case (action.type === 'addProperties'):
+        for (const key in action.extraData) {
+          stateClone[key] = action.extraData[key];
+        }
```
VladimirEdelmann, line 18: This for loop is redundant. Use `Object.assign(stateClone, action.extraData);` instead

src/transformStateWithClones.js
```diff
@@ -7,7 +7,45 @@
...
+        const addPropertiesClone = Object.assign({}, stateClone);
+
+        arrayOfClones.push(addPropertiesClone);
+
+        break;
+
+      case (action.type === 'removeProperties'):
+        for (const key in action.keysToRemove) {
+          delete stateClone[action.keysToRemove[key]];
+        }
```
VladimirEdelmann, line 29: Use for of here
```
    delete stateClone[key];
}
```

VladimirEdelmann (changes requested):
1. Add new object state after switch statement in the end of loop

valerii-derkach (approved):

PR #1738, opened 2022-11-25
solution

dolnaro:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateVersions = [];
+  let stClone = { ...state };
+
+  for (const order of actions) {
+    switch (order.type) {
+      case 'addProperties':
+        stClone = { ...stClone };
```
Alireza-Mohammadhossein, line 16: I don't think this reassign is necessary

Alireza-Mohammadhossein (commented):
Good job

src/transformStateWithClones.js
naziamoff: I agree with @Alireza-Mohammadhossein

naziamoff (approved):
GJ overall. But take a look at the comment of @Alireza-Mohammadhossein

PR #1739, opened 2022-11-25
solution

kenvastolord:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,40 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  let stateClone = { ...state };
+  const stateTransform = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        stateClone = { ...stateClone };
```
Alireza-Mohammadhossein, line 17: I don't think this reassign is necessary

Alireza-Mohammadhossein (commented):
Good job

src/transformStateWithClones.js
naziamoff: I agree with @Alireza-Mohammadhossein

naziamoff (approved):
GJ 🔥

PR #1742, opened 2022-11-25
Solution

NgineZZ:


Alireza-Mohammadhossein (commented):
Well done!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let cloneObj = { ...state };
```
naziamoff, line 11: better to avoid calling variables like `result` or explicitly write the type of data we store, like `Obj`.
```suggestion
  const states = [];
  let clone = { ...state };
```

naziamoff (approved):
Good job, but consider my comments on variable naming

PR #1744, opened 2022-11-25
solution

OlexandraKorol:
Stateful clones

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arr = [];
+  const clone = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(clone, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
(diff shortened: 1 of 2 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const transformState = [];
```
VladimirEdelmann, line 10: stansformedStates is clearer
And also stateList, stateVarsions is clear to me.
Maybe allStates.
Transform is a verb. Using verbs is pretty clear for function names when we change something. But when we talking about array we need to describe what type of values we store there

VladimirEdelmann (approved):

valerii-derkach (changes requested):
fix comments

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const transformedState = [];
```
valerii-derkach, line 10:
```suggestion
  const transformedStates = [];
```
this is plural, all states, so use meaningful names

valerii-derkach (approved):

PR #1747, opened 2022-11-25
solution

IoghiStudio:


mmahfuzi (approved):
Well done!

naziamoff (approved):
🔥 
But let's not name our variables like `result`

PR #1748, opened 2022-11-26
Finished

JzSol:


dahimohamed (approved):
Good job !

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      case "clear":
+        copy = {};
+        break;
+
+      default:
+        throw new Error(
+          action.type
+            ? `${action.type} is not supported`
+            : `Action is ${action.type}`
+        );
```
naziamoff, line 32: you refer to action type anyway here, even if it is undefined. So basically you do the same thing in both cases

naziamoff (approved):
Well done, but the error logic should be improved

PR #1750, opened 2022-11-26
solution

kreisberg:


mr-valentos (approved):
All good

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const stateArrCopy = [];
```
valerii-derkach, line 11: let's use something more meanfull in variable names
this array is the history of states, also avoid data types in names

valerii-derkach (changes requested):

src/transformStateWithClones.js
kreisberg: ✔️

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const stateVersion = [];
```
valerii-derkach, line 11:
```suggestion
  const stateVersions = [];
```

valerii-derkach (approved):

PR #1754, opened 2022-11-26
Solution for transform state with clones

katerinalex:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const versionOfState = [];
+  let newState = { ...state };
+
+  for (const i of actions) {
+    if (i.type === 'addProperties') {
+      Object.assign(newState, i.extraData);
+      versionOfState.push({ ...newState });
+    } else if (i.type === 'removeProperties') {
```
polosanya, line 17: Avoid nesting if...else. Your code looks a bit complicated. 
Try [switch...case](https://www.w3schools.com/js/js_switch.asp) instead

polosanya (changes requested):
Good job! Please check my comments to make your code cleaner

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const versionOfState = [];
+  let newState = { ...state };
+
+  for (const i of actions) {
```
polosanya, line 13:
```suggestion
  for (const action of actions) {
```
Let's use more informative names

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+    switch (i.type) {
+      case 'addProperties':
+        Object.assign(newState, i.extraData);
+        versionOfState.push({ ...newState });
+        break;
+
+      case 'removeProperties':
+        for (const j of i.keysToRemove) {
```
polosanya, line 21: Same here. What is `j`?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+          if (j in newState) {
+            delete newState[j];
+          }
+        }
+        versionOfState.push({ ...newState });
+        break;
+
+      case 'clear' :
+        newState = {};
+        versionOfState.push({ ...newState });
```
polosanya, line 31: It would be great to add a default case to avoid potential errors in a situation where the type does not match any of the cases.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#syntax

polosanya (changes requested):
Almost perfect! Let's polish it up

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (key in newState) {
+            delete newState[key];
+          }
+        }
+        versionOfState.push({ ...newState });
```
tetiana-ivanchenko, line 26: do it once after all cases

tetiana-ivanchenko (changes requested):

polosanya (approved):
Well done!

PR #1763, opened 2022-11-27
Solution Transform State

Sasha-Green:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateArray = [];
+  const copy = { ...state };
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
```
qoa11a, line 14: extraData and keysToRemove might be undefined. It might be a source of bugs, so it's better not to desctructure properties that are not 100% in an object.

qoa11a (changes requested):

PR #1764, opened 2022-11-27
Solution

RomeoKs:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  let cloneState = Object.assign({}, state);
```
UmizDemud, line 11: You can change this with, but I think this code does exactly the same thing. Just noting for the sake of it.
```js
let cloneState = { ...state };
```

UmizDemud (approved):
Code seems perfect 👍

src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
...
+  const result = [];
```
naziamoff, line 12: better to change the name for the more meaningful one

naziamoff (approved):
Great job!

src/transformStateWithClones.js
RomeoKs: And for the sake of it I will also do it.

Thank you Ummut!

PR #1765, opened 2022-11-27
solution

NadiiaKotliar:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+
+      case 'clear':
+        for (const key in clonesState) {
+          delete clonesState[key];
+        }
+        break;
+
+      default:
+        break;
+    }
```
tetiana-ivanchenko, line 33: default should handling errors

tetiana-ivanchenko (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+
+      case 'removeProperties':
+        for (const n of action.keysToRemove) {
+          delete clonesState[n];
+        }
+        break;
+
+      case 'clear':
+        clonesState = {};
+        break;
```
tetiana-ivanchenko, line 27: I mentioned before, that default case should handling error, but you just removed it. add default, and add error handling into it
image

tetiana-ivanchenko (changes requested):

PR #1766, opened 2022-11-27
finished

jackelashkar:


alaaashkar (approved):
good job !

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newArray = [];
+  const newState = { ...state };
```
naziamoff, line 11: variable name should describe what is stored inside . Not the purpose you use it for, not it’s data type. So let’s avoid naming our variables like result , newArray or obj

naziamoff (approved):
Well done! 🔥 
But consider the naming comments. And would be good to add some empty lines between cases

PR #1767, opened 2022-11-27
completed

alaaashkar:


jackelashkar (approved):
goodjob!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newArray = [];
+  const newState = { ...state };
```
naziamoff, line 11: Let's make those variable names more descriptive and clear
```suggestion
  const states = [];
  const clone = { ...state };
```

naziamoff (approved):
GJ 🔥 
But consider my comments on naming

PR #1770, opened 2022-11-27
solution

anastasiia-roiek:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+  const stateVersions = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(clone, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const keysToRemove of action.keysToRemove) {
```
Sergii-Osedchuk, line 20: I think you have to write variable name in loop in singular form instead of keysToRemove.

Sergii-Osedchuk (approved):
Change variable name in loop.
Everything is good.

Ihor-Karpyn (approved):

PR #1780, opened 2022-11-28
solution

vladimir-arkanov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+    ...state,
+  };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
+        for (const key in newState) {
+          delete newState[key];
+        }
+        resArray.push({ ...newState });
```
oleh-yablunovskyi, line 21: p4 of checklist. 
You don't need to push your newState in every case. Just do it once in the end of cycle iteration.

oleh-yablunovskyi (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resArray = [];
```
valerii-derkach, line 10:
```suggestion
  const resArray = [];
```

Please choose more meaningful names, it is the array of history, also avoid using data types in names

valerii-derkach (changes requested):

valerii-derkach (approved):

PR #1785, opened 2022-11-29
Solution

Anastasia4sik:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+
+      case 'removeProperties':
+        for (const keyToRemove of action.keysToRemove) {
+          delete actionState[keyToRemove];
+        }
+        break;
+
+      case 'clear':
+        for (const key in actionState) {
+          delete actionState[key];
```
VladimirEdelmann, line 27: Keep one way of naming

VladimirEdelmann (approved):

src/transformStateWithClones.js
Anastasia4sik: I`m sorry, but I don`t think this is necessary

valerii-derkach (approved):

PR #1787, opened 2022-11-29
first commit by Edriss Miakhel

Edriss572:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const stateVersions = [];
+  let stateClone = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        stateClone = { ...stateClone };
```
naziamoff, line 17: why do you need to copy the state before each method?

PR #1798, opened 2022-12-07
Initial solution

Dynamizx:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const _state = { ...state };
```
polosanya, line 10: image

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+  const states = [];
```
polosanya, line 11: Same here. Would be great to use more descriptive name for this variable

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+
+  for (const action in actions) {
+    switch (actions[action].type) {
```
polosanya, line 14: Can't you just write like this?
```suggestion
    switch (action.type) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+      case 'addProperties':
+        Object.assign(_state, actions[action].extraData);
```
polosanya, line 16: Same here and below
```suggestion
        Object.assign(_state, action.extraData);
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+        break;
+      case 'removeProperties':
+        for (const key of actions[action].keysToRemove) {
+          delete _state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in _state) {
+          delete _state[key];
+        }
```
polosanya, line 26: image

polosanya (changes requested):
Almost perfect! Let's provide some fixes

polosanya (approved):
Well done!

PR #1802, opened 2022-12-09
Solution

AnnyKoval:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
-  // write code here
+  const stateClone = { ...state };
+  const stateArray = [];
+
+  for (const action of actions) {
+    const { type } = action;
+
+    switch (type) {
+      case 'clear':
+        for (const key in stateClone) {
```
witflash, line 18: you can declare `let stateClone = { ...state }` and for `case 'clear'` just do `stateClone = {}` (create an empty object).

Bu your solution is workable too.

witflash (commented):
Great! 👍🏼

witflash (approved):

PR #1807, opened 2022-12-13
first attempt tests passed but feel lik etoo much code

spencerspe:


qoa11a (changes requested):
This is really a lot of redundant code. See my comments to previous task (js_stateful-object) and apply them here as well. Also you don't need so much loops and conditions. Try to solve it using only one loop (for (const action of actions)) and one switch case.

spencerspe:
Thank you. I know it is too much code but my first goal was to get things
working. I used a for loop with i as I wanted to compare i with i -1 can I
do that with (const action of actions)?

I found creating the first object off one object and then creating every
further object off the new object difficult.

spencerspe:
>Got it, thanks.

jstmpelowycz (changes requested):
great job! one thing to fix

spencerspe:
Thank you. What do you suggest as default? Does it make sense to just have
newState = newState, to keep things flowing?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let newState = { ...state };
+  const clones = [];
+
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 14: You can make a destructuring to get all the properties you need of action. 
It can help to make your code cleaner

polosanya (approved):
You can just return a state without any changes on default case, or some error

PR #1809, opened 2022-12-16
Solution

yana-kotsulym:


ViktorHavryliuk (changes requested):
Please use for ... of loop to iterate through actions:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of
And switch case operator to check action types: 

This will make your code much cleaner and easier to understand

src/transformStateWithClones.js
```diff
@@ -7,7 +7,81 @@
...
+    const actionValues = Object.values(part);
+    const actionKeys = Object.keys(part);
+
+    for (let n = 0; n < actionKeys.length; n++) {
+      switch (actionKeys[n]) {
+        case 'type':
+          switch (actionValues[n]) {
+            case 'addProperties':
+              if (result.length === 0) {
+                result.push(Object.assign({}, state, actionValues[n + 1]));
```
tetiana-ivanchenko, line 22: prev commens was about ude for of loop, also your algorithm is very difficult 
why do you need nested switches?
you can just create ,cope of state, and operate with it 
f.e.:

const copeState = { ...state };

for (const action of  actions) { 
  switch (action.type) {
    case 'addProperties':
        break;
        ......
        that's all what you need for addProperties case, also you can simplify all your cases

tetiana-ivanchenko (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(copyState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if ([key] in copyState) {
```
ivanchuk-yurii, line 22: you don't need this

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+            delete copyState[key];
+          }
+        }
+        break;
+
+      case 'clear':
+        for (const item in copyState) {
+          delete copyState[item];
+        }
```
ivanchuk-yurii, line 31:
```suggestion
        result.push({});
        continue;
```

ivanchuk-yurii (approved):

PR #1813, opened 2022-12-18
solution

drhtka:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+        break;
+
+      case 'removeProperties':
+
+        for (const keys of action.keysToRemove) {
+          delete stateCopy[keys];
+        }
+
+        stateArray.push({ ...stateCopy });
+    }
```
DarkMistyRoom, line 36: switch/case should always have default case for error handling.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+  const stateCopy = { ...state };
+  const stateArray = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        stateArray.push({ ...stateCopy });
```
DarkMistyRoom, line 19: you can move push after switch. you make the same operation regardless of the action type

DarkMistyRoom (changes requested):

witflash (approved):
GJ!

PR #1819, opened 2022-12-21
Solution

anna-inozemets:
Function transformStateWithClones was realised

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultObject = { ...state };
+  const resultArr = [];
+
+  for (const item of actions) {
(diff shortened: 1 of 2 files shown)


andrieieva-yuliia (approved):
Good job 🌷

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+  const stateClone = { ...state };
+  const resultArr = [];
+
+  for (const item of actions) {
+    switch (item.type) {
+      case 'clear':
+        for (const key in stateClone) {
+          delete stateClone[key];
+        }
+        resultArr.push({ ...stateClone });
```
Denchepornyuk, line 19: don't push in each case, you can do it only once at the end of switch case

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1821, opened 2022-12-23
Solution

ruslan-yarosh:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
ViktorHavryliuk, line 14: Would be better to use switch statement instead of multiple ifs

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        delete stateCopy[key];
+      }
+
+      result.push({ ...stateCopy });
+    }
+
+    if (action.type === 'clear') {
+      for (const key in stateCopy) {
+        delete stateCopy[key];
+      }
```
ViktorHavryliuk, line 31: Can do stateCopy = {} here instead

ViktorHavryliuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+          delete stateCopy[key];
+        }
+
+        result.push({ ...stateCopy });
+
+        break;
+
+      case 'clear':
+        stateCopy = {};
+        result.push({ ...stateCopy });
```
TarasHoliuk, line 32: since you've written a new object literal no need to copy it via the spread operator - it is already an entirely new object, so you can do it in this way: `push(stateCopy);`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+        break;
```
TarasHoliuk, line 34: always add the `default` case

TarasHoliuk (changes requested):

src/transformStateWithClones.js
ruslan-yarosh: If I don't use the spread operator in this case the last test will fail. Because in case 'addProperties ' after Object.assign the last object, which was pushed after case 'clear' will be changed too.
image
image
image

ViktorHavryliuk (approved):

PR #1824, opened 2022-12-23
Transform state with clones solution

andrieieva-yuliia:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let stateCopy = { ...state };
+  const stateCopies = [];
```
OlhaYevstifieieva, line 11: I think you should not name variables with a difference of several letters. Because of this, there may be errors in the future.

OlhaYevstifieieva (approved):
Perfect job! 🤗

src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => delete stateCopy[key]);
+        break;
+      case 'clear':
+        stateCopy = {};
```
Denchepornyuk, line 22: let's remove all properties from stateCopy object instead of assigning an empty object

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1825, opened 2022-12-24
Solution

undepend:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+      for (const prop of obj.keysToRemove) {
+        delete resultObject[prop];
+      }
+      result.push({ ...resultObject });
+    }
+
+    if (obj.type === 'clear') {
+      resultObject = {};
+      result.push({ ...resultObject });
+    }
```
SanyaBratashchuk, line 29: use `switch case`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+    if (obj.type === 'addProperties') {
+      Object.assign(resultObject, obj.extraData);
+      result.push({ ...resultObject });
+    }
+
+    if (obj.type === 'removeProperties') {
+      for (const prop of obj.keysToRemove) {
+        delete resultObject[prop];
+      }
+      result.push({ ...resultObject });
```
SanyaBratashchuk, line 23: You can push just once in the end of each iteration

SanyaBratashchuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+        for (const prop of obj.keysToRemove) {
+          delete resultObject[prop];
+        }
+        break;
+
+      default:
+        resultObject = {};
+        break;
+    }
+    result.push({ ...resultObject });
```
witflash, line 29:
```suggestion
    }
    
    result.push({ ...resultObject });
```
I prefer to use an empty line between different block codes.

witflash (approved):
Good job!

PR #1829, opened 2022-12-25
commit

jaxxxon78:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+      case 'removeProperties' :
+        for (const prop of action.keysToRemove) {
+          delete copy[prop];
+        }
 
+        break;
+
+      case 'clear' :
+        copy = {};
+        break;
```
Viktor-Kost, line 28: It is a good practice to add `default` case for `switch` statement
If you don't have a default behavior - just throw an Error with text "incorrect action type"

Viktor-Kost (approved):
GJ!
Add a recommendation for you

PR #1836, opened 2022-12-26
Solution

sv-dubov:


nzrii-kysel (approved):
looks good
checklist recommends using a `switch` statement

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      stateArr.push({ ...stateCopy });
+      continue;
+    }
+
+    if (action.type === 'clear') {
+      for (const key in stateCopy) {
+        delete stateCopy[key];
+      }
+      stateArr.push({ ...stateCopy });
+    }
```
Denchepornyuk, line 33: check checklist, it's better to use switch case here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const stateArr = [];
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(stateCopy, action.extraData);
+      stateArr.push({ ...stateCopy });
```
Denchepornyuk, line 16: don't push in each case, you can do it only once at the end of switch case

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1842, opened 2022-12-27
Solution

chermak-oleh:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clone = { ...state };
+  const copyArr = [];
+
+  for (const obj of actions) {
+    if (obj.type === 'addProperties') {
```
andrieieva-yuliia, line 14: According to the checklist it's better to use switch statement 🙂

andrieieva-yuliia (approved):
Nice work 👍

src/transformStateWithClones.js
Denchepornyuk: agree, let's use switch case here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      Object.assign(clone, obj.extraData);
+      copyArr.push({ ...clone });
```
Denchepornyuk, line 16: don't `push` in each case, you can do it only once at the end of switch case

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1846, opened 2022-12-28
Solution

Coyoote:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopies = [];
+  const stateCopy = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        stateCopies.push({ ...stateCopy });
```
andrieieva-yuliia, line 17: You don't need to repeat this line, just write it below once 🙂

andrieieva-yuliia (changes requested):

Coyoote:
Agreed, fixed it accordingly, thanks for the review 😃

src/transformStateWithClones.js
Coyoote: Agreed, fixed it accordingly, thanks for the review 😃

src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopies = [];
+  const stateCopy = { ...state };
```
Denchepornyuk, line 11: it can be an empty object
```suggestion
  const stateCopy = {};
```

Denchepornyuk, line 11: it can be an empty object
```suggestion
  const stateCopy = {};
```

Denchepornyuk (approved):

src/transformStateWithClones.js
Coyoote: I've tried with an empty object, but the tests don't go through as the copy doesn't have starting values of the state object.

PR #1847, opened 2022-12-28
solution

Dashaverz:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+
+      case `removeProperties`:
+        for (const key of action.keysToRemove) {
+          if (stateCopy.hasOwnProperty(key)) {
+            delete stateCopy[key];
+          }
```
nzrii-kysel, line 24: no need for this `if` statement, nothing happens if you try to `delete` a property that isn't in the object

nzrii-kysel (changes requested):

src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const res = [];
```
alexandra-protyanova, line 11: Not the best name for variable

alexandra-protyanova (approved):

PR #1851, opened 2022-12-28
solution

vladyslav78292:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = Object.assign({}, state);
+  const result = [];
```
chermak-oleh, line 11: It is might be better name than just 'result'. Anyway, good job!)

chermak-oleh (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        break;
+
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+
+      default:
+        throw Error('error');
+    }
+    result.push(Object.assign({}, stateCopy));
```
Denchepornyuk, line 33: It's better to use spread operator to make a copy instead of Object.assign

Denchepornyuk (approved):

PR #1855, opened 2022-12-28
finished Stateful clone task

andrii-kurash:
Have a good time !🔥

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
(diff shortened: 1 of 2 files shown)


OlhaYevstifieieva (changes requested):
Good job, but checklist for what to do this with switch. Don`t forget about default condition

andrii-kurash:
> Good job, but checklist for what to do this with switch. Don`t forget about default condition

Thanks, I didn`t see it!👍

alexandra-protyanova (approved):
Well done 👍

PR #1856, opened 2022-12-28
Added solution

burovetskyi:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+      case 'addProperties':
+        Object.assign(newState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => delete newState[key]);
+        break;
+      case 'clear':
+        newState = {};
+        break;
```
chermak-oleh, line 24: It is better to remove all properties from newState object instead of assigning an empty object

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClone = [];
+  let newState = { ...state };
```
chermak-oleh, line 11: use 'const' instead of 'let'

chermak-oleh (changes requested):
Good job, need some fixes

src/transformStateWithClones.js
```diff
@@ -7,7 +7,26 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClone = [];
+  const newState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```
Denchepornyuk, line 14: where is default block?

Denchepornyuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => delete newState[key]);
+        break;
+      case 'clear':
+        Object.keys(newState).forEach(key => delete newState[key]);
+        break;
+
+      default:
+        break;
```
Denchepornyuk, line 27: add some logic to default block

Denchepornyuk (approved):

PR #1857, opened 2022-12-28
Solution

MFrolova-art:


victorionov (approved):
Good Job!!

src/transformStateWithClones.js
```diff
@@ -8,6 +8,28 @@
...
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+      case 'removeProperties':
+        for (const keyToRemove of action.keysToRemove) {
+          delete stateCopy[keyToRemove];
+        }
+        break;
+      case 'clear':
+        stateCopy = {};
```
Denchepornyuk, line 25: don't create new empty object, it's better to clear existing object instead of creating a new one

Denchepornyuk (changes requested):

PR #1858, opened 2022-12-28
solution

OlehSharha:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+      case 'clear' :
+        Object.keys(stateClone).forEach(key => delete stateClone[key]);
+        break;
+      case 'addProperties' :
+        Object.assign(stateClone, action.extraData);
+        break;
+      case 'removeProperties' :
+        for (const removleKey of action.keysToRemove) {
+          delete stateClone[removleKey];
+        }
```
Avichez, line 24: You can use here also method `forEach()`,  but I don't think that this is a mistake.

Avichez (approved):
Good solution, especially throwing error by default, good idea) 👍

alexandra-protyanova (approved):
Well done 👍

PR #1863, opened 2022-12-28
Cloning and transforming state - completed

GarryRocksman:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const cloneState = { ...state };
+
+  for (const action of actions) {
+    switch (action['type']) {
```
Anastasiia-Khudych, line 14: I guess, it's better to use dot notation here, because `type` is not a variable. ( `action.type`).

Anastasiia-Khudych (approved):
Good job!

src/transformStateWithClones.js
Denchepornyuk, line 14:
```suggestion
    switch (action.type) {
```

Denchepornyuk (approved):

src/transformStateWithClones.js
GarryRocksman: fixed

PR #1866, opened 2022-12-28
solution

victorionov:


GarryRocksman (approved):
GJ. But as Denny's says - code in one line is not always better code. Because it's harder to read and understand it. But it's just my thoughts. Good Job!!!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+  const result = [];
+
+  for (const transform of actions) {
+    switch (transform.type) {
+      case 'addProperties' :
+        Object.assign(copyState, transform.extraData);
+        break;
+      case `removeProperties` :
+        transform.keysToRemove.forEach(keyToRemove =>
+          delete copyState[keyToRemove]);
```
Denchepornyuk, line 20: not the best code formatting

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
+  const result = [];
```
Denchepornyuk, line 11: `result` isn't the best name for variable

Denchepornyuk (approved):

PR #1867, opened 2022-12-28
solution

Oleksandr-Maly:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const propAfterAction = [];
+
+  for (const action of actions) {
+    switch (true) {
+      case action.type === 'addProperties' :
```
imariash86, line 15: the same here, you can move 'action.type' right to switch

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        Object.assign(stateCopy, action.extraData);
+        break;
+
+      case action.type === 'removeProperties' :
+        for (const keys of action.keysToRemove) {
+          if (stateCopy.hasOwnProperty(keys)) {
+            delete stateCopy[keys];
+          };
```
imariash86, line 23: I think there is no need to check 'hasOwnProperty', it won't be deleted if it isn't there:)

imariash86 (approved):
good job:)

ArturErmolenko (approved):

PR #1868, opened 2022-12-28
solution

ann-yurkevych:


nzrii-kysel (approved):
nice work

src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
...
+    switch (action.type) {
+      case 'addProperties': {
+        Object.assign(currentState, action.extraData);
+        result.push({ ...currentState });
+        break;
+      }
+
+      case 'removeProperties': {
+        action.keysToRemove.forEach((key) => {
+          if (key in currentState) {
```
alexandra-protyanova, line 24: It is overkill here, if you delete property which is not one of the object's, it is just doing nothing, so it is ok

src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
...
+        });
+        result.push({ ...currentState });
+        break;
+      }
+
+      case 'clear': {
+        for (const key in currentState) {
+          delete currentState[key];
+        }
+        result.push({ ...currentState });
```
alexandra-protyanova, line 36: Let us find a way to prevent duplication of this line for three times

alexandra-protyanova (changes requested):

ArturErmolenko (approved):

PR #1874, opened 2022-12-29
first commit

karaedd:


marina-tilniak (changes requested):
also would  be good to use switch case here also

Viktor-Kost (approved):
Nice!

For future - it's a good practice to use `default` case at switch statement. If you don't have default behavior - just throw an error

PR #1876, opened 2022-12-29
task stateful_clones was implemented

OlyaZaiets:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete copyState[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in copyState) {
+          delete copyState[key];
+        }
```
nzrii-kysel, line 28:
```suggestion
        copyState = {};
```

nzrii-kysel (approved):
well done

src/transformStateWithClones.js
alexandra-protyanova: Good idea to avoid extra loop

alexandra-protyanova (approved):

PR #1884, opened 2023-01-01
Solution

VadymTsyndra:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,67 @@
...
+        arr.push(obj);
+      } else {
+        const obj = Object.assign({}, arr[arr.length - 1]);
+
+        for (const prop in obj) {
+          delete obj[prop];
+        }
+
+        arr.push(obj);
+      }
```
DarkMistyRoom, line 30: why do you need this` if else` construction? just make a copy of the `state` and push the current state in the end of every interation

DarkMistyRoom (changes requested):
Use `switch case`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      case ('addProperties'):
+        Object.assign(copyOfState, action.extraData);
+        break;
+      case ('removeProperties'):
+        for (const prop of action.keysToRemove) {
+          delete copyOfState[prop];
+        }
+        break;
+      default:
+        action = null;
```
SanyaBratashchuk, line 29:
```suggestion
        throw new Error('unknown action type')
```

SanyaBratashchuk (approved):

PR #1885, opened 2023-01-03
Solution

Alessandranitti15:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const allActions = []; // final array
```
jstmpelowycz, line 10: this one should not be called `allActions`, it's more like a history of states, name it accordingly

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+  const newState = { ...state };
+
+  for (const keys in actions) {
+    if (actions[keys].type === 'clear') {
```
jstmpelowycz, line 14: let's use switch-case statements

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      }
+    }
+
+    if (actions[keys].type === 'addProperties') {
+      Object.assign(newState, actions[keys].extraData);
+    }
+
+    if (actions[keys].type === 'removeProperties') {
+      for (const action in actions[keys].keysToRemove) {
+        delete newState[actions[keys].keysToRemove[action]];
```
jstmpelowycz, line 26: split this single-line action into multi-line actions so that this code construction becomes more readable

jstmpelowycz (changes requested):
nicely done, it works! let's take care of our code

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const stateHistory = [];
+
+  for (const action in actions) {
```
qoa11a, line 13: It's better to use `for .. of` loop for arrays. Also then you won't need to get actionToDo, as your every action would already be what you need instead of index.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+      case 'addProperties':
+        Object.assign(newState, actionToDo.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key in actionToDo.keysToRemove) {
+          const toDelete = actionToDo.keysToRemove[key];
+
+          delete newState[toDelete];
+        }
```
qoa11a, line 26: Same here. See how easier it becomes with `for .. of` loop.
```suggestion
        for (const key of actionToDo.keysToRemove) {
          delete newState[key];
        }
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        break;
+
+      case 'clear':
+        for (const property in newState) {
```
qoa11a, line 30: It's actually a key, not a property.

qoa11a (changes requested):
Good job! 👍 
Few small fixes.

jstmpelowycz (approved):

PR #1892, opened 2023-01-06
Solution

ihor-halkin:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  let clone = { ...state };
+  const result = [];
+
+  for (const action of actions) {
```
loralevitska, line 13: Destructure action

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+
+      case 'removeProperties':
+        for (const prop of action.keysToRemove) {
+          delete clone[prop];
+        };
+        break;
+
+      default:
+        // eslint-disable-next-line no-console
+        console.log('Action not found');
```
loralevitska, line 31: When none of the values fit, you just need to move on through the loop instead of outputting anything to the console, fix that

loralevitska (changes requested):
Almost done)

Anastasiia-Svintsova (approved):

PR #1893, opened 2023-01-08
solution of transformStateWithClones

RomanTsvindin:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const transformedState = Object.assign({}, state);
+  const transformedArray = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(transformedState, action.extraData);
```
polosanya, line 16: You can make destructuring to get all the properties of action that you need.
It will make your code cleaner

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+        for (const key of action.keysToRemove) {
+          delete transformedState[key];
+        };
+        break;
+      case 'clear':
+        for (const key in transformedState) {
+          delete transformedState[key];
+        }
+        break;
+    }
```
polosanya, line 28: It is better to add a default case to handle situations when action type didn't match any of cases.
image

polosanya (changes requested):
Let's polish it up!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const transformedState = Object.assign({}, state);
+  const transformedArray = [];
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
```
```suggestion
    const { 
      type, 
      extraData, 
      keysToRemove 
    } = action;
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+          delete transformedState[key];
+        };
+        break;
+
+      case 'clear':
+        for (const key in transformedState) {
+          delete transformedState[key];
+        }
+        break;
+      default: return undefined;
```
polosanya, line 31: I think it's better to throw some error or just break

polosanya (approved):
Good job!

PR #1894, opened 2023-01-08
added solution

constantweb:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let clone = { ...state };
+  const result = [];
+
+  for (const prop of actions) {
```
tetiana-ivanchenko, line 13:
```suggestion
  for (const action of actions) {
```

tetiana-ivanchenko (changes requested):

Bohdan-Maksymiuk (changes requested):
[checklist](https://github.com/mate-academy/js_stateful-clones/blob/master/checklist.md)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+          delete clone[key];
+        }
+        break;
+
+      case 'clear':
+        clone = {};
+        break;
+
+      default:
+        break;
```
tetiana-ivanchenko, line 30: image

tetiana-ivanchenko (changes requested):
fix prev comments, please

src/transformStateWithClones.js
constantweb: Hi, I'm not quite sure what I'm supposed to fix

tetiana-ivanchenko: default case should handling error, and you just make it empty
use `throw new Error('your error text')`

tetiana-ivanchenko (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+          delete clone[key];
+        }
+        break;
+
+      case 'clear':
+        clone = {};
+        break;
+
+      default:
+        throw new Error('your error text');
```
nazarmatsevych, line 30:
```suggestion
        throw new Error('your error text');
```

It's better to use smth like 'Unknown action type' instead of 'your error text'

nazarmatsevych (approved):
gj, just small suggestion

PR #1895, opened 2023-01-08
Solution

ia-lerimund:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,27 @@
...
+      for (const key of actions[prop].keysToRemove) {
+        delete newObj[key];
+      }
+      result.push({ ...newObj });
+    } else if ((actions[prop].type === 'clear')) {
+      for (const key in newObj) {
+        delete newObj[key];
+      }
+      result.push({});
+    }
```
Bohdan-Maksymiuk, line 28: avoid such nesting `else if` conditions
This code is hard to read and maintain
use `switch` instead

src/transformStateWithClones.js
```diff
@@ -8,6 +8,27 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const newObj = Object.assign({}, state);
+  const result = [];
+
+  for (let prop = 0; prop < actions.length; prop++) {
```
Bohdan-Maksymiuk, line 14: use `for of` loop
you don't need indexes here

Bohdan-Maksymiuk (changes requested):

src/transformStateWithClones.js
ia-lerimund: Fixed it, thanks for the advice!

src/transformStateWithClones.js
```diff
@@ -8,6 +8,30 @@
...
+          delete newObj[key];
+        }
+        result.push({ ...newObj });
+        break;
+      case 'clear':
+        for (const key in newObj) {
+          delete newObj[key];
+        }
+        result.push({});
+    }
```
witflash, line 31: You need to add `default` statement for `switch / case` block. 
You can handle errors there, for example

witflash (changes requested):
Good job!
But check my last comment, please.

PR #1897, opened 2023-01-11
Develop

Spec0s:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,46 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const stateClonesArr = [];
+  let stateClone = { ...state };
+
+  function removeProps(propsToRemove) {
+    for (const key of propsToRemove) {
+      stateClone = { ...stateClone };
```
qoa11a, line 16: Why do you do it?

qoa11a (approved):
Good job! 👍 
I don't see a reason for re-declaring stateClone in removeProps and clearAllProps functions though.

src/transformStateWithClones.js
Spec0s: i am still a bit confused with this cloning and reference thing with clones have maybe different material i can read to understand it better ?  i  placed stateClone to different spot is that what you meant ?

polosanya (approved):

PR #1900, opened 2023-01-14
solution

44shaggy:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const currentObject = { ...state };
+  const clonesArray = [];
+
+  for (const i in actions) {
```
polosanya, line 13:
```suggestion
  for (const action of actions) {
```
Don't iterate array using `in` operator.
Also, use descriptive names

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+    switch (actions[i].type) {
```
```suggestion
    switch (action.type) {
```
You can also use desctructuring to get all the properties of action

polosanya (changes requested):

src/transformStateWithClones.js
44shaggy: yep. that looks better

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const currentObject = { ...state };
```
polosanya, line 10:
```suggestion
  const stateCopy = { ...state };
```
Actually, it's not a currentObject

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          delete currentObject[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in currentObject) {
+          delete currentObject[key];
+        }
```
polosanya, line 30: You can also just make it an empty object:
`currentObject = {}`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        break;
+    }
```

polosanya (changes requested):
Let's polish it up)

polosanya (approved):
Well done!

PR #1901, opened 2023-01-16
Solution

OlyaNikora:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
+
+      default:
```
Viktor-Kost, line 31: for default you can use either some behavior or throwing an error

Viktor-Kost (approved):

PR #1912, opened 2023-01-19
Solution

pushkarskiyrodion:


DarkMistyRoom (approved):
good job, but switch/case should always have default case for error handling

PR #1916, opened 2023-01-21
My solution

andriy-litkovets:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let copyOfState = { ...state };
+  const answerArray = [];
```
polosanya, line 11: I think it's better to specify what array it is

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
```
```suggestion
    const { 
      type, 
      extraData,
      keysToRemove,
    } = action;
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          delete copyOfState[key];
+        }
+        break;
+
+      case 'clear':
+        for (const item in copyOfState) {
+          delete copyOfState[item];
+        }
```
polosanya, line 30:
```suggestion
        for (const item in copyOfState) {
          delete copyOfState[item];
        }
```
```suggestion
        copyOfState = {}
```
You can just make it an empty object

polosanya (changes requested):
Almost done! A few small fixes

polosanya (approved):
Well done!

PR #1921, opened 2023-01-23
Solution

georgiy-nezbritskiy:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,44 @@
...
+
+    if (action.type === 'addProperties') {
+      for (const key in action.extraData) {
+        newState[key] = action.extraData[key];
+      }
+
+      result.push({
+        ...newState,
+      });
+    }
```
vkryvytskyy, line 45: It's better to use `switch`

src/transformStateWithClones.js
```diff
@@ -8,6 +8,44 @@
...
+
+    if (action.type === 'addProperties') {
+      for (const key in action.extraData) {
+        newState[key] = action.extraData[key];
+      }
+
+      result.push({
+        ...newState,
+      });
```
vkryvytskyy, line 44: You can write this once after all `if` statements, or after `switch`

vkryvytskyy (approved):

PR #1923, opened 2023-01-23
function was created

YevheniiHura:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        delete clonedState[arr[k]];
+      }
+
+      result.push(Object.assign({}, clonedState));
+    } else if (actions[i].type === 'clear') {
+      for (const key in clonedState) {
+        if (clonedState.hasOwnProperty(key)) {
+          delete clonedState[key];
+        }
+      }
```
IvanFesenko, line 30: Use swith.
You can use for of loop `for (const action of actions)` {}

IvanFesenko (changes requested):
Please check my comments

src/transformStateWithClones.js
YevheniiHura: done

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+      case 'clear':
+        for (const key in clonedState) {
+          if (clonedState.hasOwnProperty(key)) {
+            delete clonedState[key];
+          }
+        }
+
+        result.push(Object.assign({}, clonedState));
+        break;
+    }
```
IvanFesenko, line 39: You have `result.push` on each case. You can push after switch statement
```suggestion
    switch (action.type) {
      case 'addProperties':
        Object.assign(clonedState, action.extraData);
        
        break;

      case 'removeProperties':
        const arr = action.keysToRemove;

        for (const element of arr) {
          delete clonedState[element];
        }

        break;

      case 'clear':
        for (const key in clonedState) {
          if (clonedState.hasOwnProperty(key)) {
            delete clonedState[key];
          }
        }

        break;
    }
    result.push({...clonedState});
```

IvanFesenko (changes requested):
Please fix small remarks

src/transformStateWithClones.js
YevheniiHura: done

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        }
+
+        break;
+
+      case 'clear':
+        for (const key in clonedState) {
+          delete clonedState[key];
+        }
+
+        break;
```
Viktor-Kost, line 33: it is a good practice when you add a default behavior for switch statement. Yet, if you don't have such one - just throw an Error with text like "unexpected action"

Viktor-Kost (approved):
Nice!

PR #1925, opened 2023-01-24
Solution for clone

Vasyl-Pavlenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const i of action.keysToRemove) {
+          delete newState[i];
+        };
+        break;
+
+      case 'clear':
+        Object.keys(newState).forEach(key => delete newState[key]);
```
witflash, line 26: `switch / case` should have `default` statement. You can handle some errors inside it, for example

witflash (changes requested):
Good job!
But check my comment, please

Viktor-Kost (approved):

PR #1926, opened 2023-01-24
Solution

AnastasiiaEm:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const statesArray = [];
+  const modifiedStates = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
+    const type = actions[i].type;
```
OlEzhA0, line 14: try to use `for ... of`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+          delete modifiedStates[actions[i].keysToRemove[x]];
+        };
+        statesArray.push({ ...modifiedStates });
+        break;
+      case 'clear':
+        for (const key in modifiedStates) {
+          delete modifiedStates[key];
+        };
+        statesArray.push({ ...modifiedStates });
+        break;
```
OlEzhA0, line 32:
```suggestion
        statesArray.push({});
        break;
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+  const statesArray = [];
+  const modifiedStates = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
+    const type = actions[i].type;
+
+    switch (type) {
+      case 'addProperties':
+        Object.assign(modifiedStates, actions[i].extraData);
+        statesArray.push({ ...modifiedStates });
```
OlEzhA0, line 19:
```suggestion
        statesArray.push(Object.assign(modifiedStates, actions[i].extraData));
```
or
```suggestion
        statesArray.push({ ...modifiedStates, ...actions[i].extraData });
```

OlEzhA0 (changes requested):

src/transformStateWithClones.js
AnastasiiaEm: In these ways I add extraData only to a copy of an object that I push to an array. But according to the task I also need to change the object modifiedStates for the next iterations.

src/transformStateWithClones.js
AnastasiiaEm: If I push an empty Array, modifiedStates object is not empty for the next iterations.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+    switch (type) {
+      case 'addProperties':
+        Object.assign(modifiedStates, key.extraData);
+        statesArray.push({ ...modifiedStates });
+        break;
+      case 'removeProperties':
+        for (let x = 0; x < key.keysToRemove.length; x++) {
+          delete modifiedStates[key.keysToRemove[x]];
+        };
+        statesArray.push({ ...modifiedStates });
```
Viktor-Kost, line 25: No needs to use statesArray.push({ ...modifiedStates }); at each switch case. 
Just use it once (after! switch statement at the end of loop)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const statesArray = [];
+  const modifiedStates = { ...state };
+
+  for (const key of actions) {
+    const type = key.type;
```
Viktor-Kost, line 14: it is an extra variable. you can use key.type at switch condition

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+
+    switch (type) {
```
Viktor-Kost, line 16:
```suggestion
    switch (key.type) {
```

Viktor-Kost (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const statesArray = [];
+  const modifiedStates = { ...state };
+
+  for (const key of actions) {
```
mykhalenych, line 13:
```suggestion
  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+    switch (key.type) {
```
mykhalenych, line 14: destruct const {key, extraData, keysToRemove} = action

mykhalenych (changes requested):

IvanFesenko (approved):

PR #1929, opened 2023-01-25
Solution

dianaavem:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+  let cloneState = {...state};
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(cloneState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const rem of action.keysToRemove) {
```
ViktorHavryliuk, line 20:
```suggestion
        for (const key of action.keysToRemove) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+          delete cloneState[rem];
+        }
+        break;
+
+      case 'clear':
+        cloneState = {};
+        break;
+
+      default:
+        return arr;
```
ViktorHavryliuk, line 30: It is better to throw error in this case
```suggestion
        throw new Error("Unsupported action type");
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let arr = [];
```
ViktorHavryliuk, line 10: Use some more meaningful name, so that it's clear what this array is for without need to look what you push there somewhere lower in the code

ViktorHavryliuk (changes requested):

SanyaBratashchuk (approved):

PR #1931, opened 2023-01-26
Develop

silcki:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
...
 function transformStateWithClones(state, actions) {
   // write code here
+  const result = [];
+  let stateCopy = { ...state };
+
+  for (const i in actions) {
+    const { type, extraData = {} } = actions[i];
+    let { keysToRemove = [] } = actions[i];
+
+    if (type === 'addProperties') {
```
OlEzhA0, line 18: try to use switch case here

src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
...
+      Object.assign(stateCopy, extraData);
+    } else {
+      if (type === 'clear') {
+        keysToRemove = Object.keys(stateCopy);
+      }
```
OlEzhA0, line 23: why do we need it?

OlEzhA0 (changes requested):

PR #1935, opened 2023-01-27
Stateful Clones

mdm-21:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let res = [];
+  const newState = { ...state };
+
+  res[0] = newState;
+
+  for (let i = 0; i < actions.length; i++) {
+    switch (actions[i].type) {
```
ViktorHavryliuk, line 16: Better use for... of loop to avoid index access
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of
```suggestion
  for (const action of actions) {
    switch (action.type) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let res = [];
```
ViktorHavryliuk, line 10: Try to give variables a more descriptive name
```suggestion
  let previousStates = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+  const newState = { ...state };
+
+  res[0] = newState;
```
ViktorHavryliuk, line 13: You can remove this, according to examples in task you don't have to push original state to array:
image

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+
+  for (let i = 0; i < actions.length; i++) {
+    switch (actions[i].type) {
+      case 'addProperties':
+        const addRes = Object.assign(res[i], actions[i].extraData);
+
+        res[i + 1] = { ...addRes };
```
ViktorHavryliuk, line 20: Try to change other cases in similar way
```suggestion
        newState = Object.assign(newState, action.extraData);

        previousVersions.push({ ...newState });
```

ViktorHavryliuk (changes requested):
Try to find solution without using accessing by index, it would reduce amount of code and make it simplier to read and understand

PR #1942, opened 2023-01-30
Solution

mdm-21:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+          delete cloneState[key];
+        }
+        break;
+
+      case 'clear':
+        cloneState = {};
+        break;
+
+      default:
+        break;
```
Volodymyr-Melnyk-Fron10der, line 30: You can throw an error if there is no such type
```suggestion
      default:
        throw new Error('There is no type like that!')
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let cloneState = { ...state };
```
Volodymyr-Melnyk-Fron10der, line 11: check this checkpoint
image

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+
+  for (const action of actions) {
```
Volodymyr-Melnyk-Fron10der, line 13: You don't need to write `action` every time to get object property, just destructurize them right away
```suggestion
  for (const { type, extraData .... } of actions) {
```

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, good work, keep moving! I left some comments you can fix

PR #1944, opened 2023-01-31
Develop

katiailchenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          clone[key] = action.extraData[key];
+        };
+
+        result.push({ ...clone });
```
ivadimko, line 20: you can move `result.push` after `switch` block to avoid repeated code

e.g.

```
switch {
  case ...
    break;
  case ...
    break;
}

push(...clone)
```

ivadimko (approved):

src/transformStateWithClones.js
katiailchenko: Yes! You are right!

PR #1950, opened 2023-02-02
Clones

Yevhen-Ivanchuk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateArray = [];
+
+  let stateCopy = {
+    ...state,
+  };
+
+  for (const part of actions) {
```
polosanya, line 16:
```suggestion
  for (const action of actions) {
```
I think action sounds better. 
Also, you can make a destructuring to get all the needed props of action (type, extraData, keysToRemove)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+    switch (part.type) {
+      case 'addProperties':
+        stateCopy = {
+          ...stateCopy,
+          ...part.extraData,
+        };
```
polosanya, line 22: You can use [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) for that

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+          delete stateCopy[removeElement];
+        }
+        stateArray.push({ ...stateCopy });
+        break;
+
+      case 'clear':
+        stateCopy = {};
+        stateArray.push({});
+        break;
+    }
```
polosanya, line 37: image
Also, you can push the stateCopy after all cases in order not to duplicate yourself

polosanya (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+
+  let stateCopy = {
+    ...state,
+  };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        stateArray.push({ ...stateCopy });
```
agniya-i, line 20: looks like `stateArray.push({ ...stateCopy });` repeat each time, so it's better to move it down, after `switch`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+        }
+        stateArray.push({ ...stateCopy });
+        break;
+
+      case 'clear':
+        stateCopy = {};
+        stateArray.push({});
+        break;
+      default:
+        throw stateCopy;
```
agniya-i, line 35:
```suggestion
        throw new Error('Unknown action type');
```
throw an error

agniya-i (changes requested):

qoa11a (approved):

PR #1956, opened 2023-02-03
Stateful clones created

yevhen1408:


tetiana-ivanchenko (changes requested):
fix your code according to checklist

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const add = 'addProperties';
+  const remove = 'removeProperties';
+  const res = [];
+  const stateCopy = Object.assign({}, state);
```
ViktorHavryliuk, line 13: you can use spread syntax here instead
```suggestion
  const stateCopy = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+          delete stateCopy[key];
+        }
+        res.push(Object.assign({}, stateCopy));
+        break;
+      default:
+        for (const prop in stateCopy) {
+          delete stateCopy[prop];
+        }
+        res.push({});
+        break;
```
ViktorHavryliuk, line 32: This is not default case, it's "clear" action type. Default case should throw error informing user that provided action type is not supported if it's neither add, remove nor clear

ViktorHavryliuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+    switch (action.type) {
+      case add:
+        Object.assign(stateCopy, action.extraData);
+        res.push({ ...stateCopy });
+        break;
+      case remove:
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
+        res.push({ ...stateCopy });
```
SanyaBratashchuk, line 26: You can push just once in the end of each iteration

SanyaBratashchuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+          delete stateCopy[key];
+        }
+        break;
+      case clear:
+        for (const prop in stateCopy) {
+          delete stateCopy[prop];
+        }
+        break;
+      default:
+        break;
```
ViktorHavryliuk, line 32: It would be better to inform user that something went wrong
```suggestion
      default:
        throw new Error("Unsupported action type");
```

ViktorHavryliuk (approved):

PR #1958, opened 2023-02-03
Solution

iryna-serniuk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const cloneArray = [];
+  const cloneObject = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(cloneObject, action.extraData);
+        cloneArray.push({ ...cloneObject });
```
Anastasiia-Svintsova, line 17: You do  `cloneArray.push({ ...cloneObject });` in any case, so move it to the end of the logic and write only once

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        break;
+
+      case 'clear':
+        for (const key in cloneObject) {
+          delete cloneObject[key];
+        }
+        cloneArray.push({ ...cloneObject });
+    }
+  }
```
Anastasiia-Svintsova, line 35: Also you should add a default case at the end of the switch/case statement

Anastasiia-Svintsova (changes requested):

TarasHoliuk (approved):

PR #1959, opened 2023-02-04
Solution

mdm-21:


ViktorHavryliuk (changes requested):
Good work, but please use switch case instead of multiple ifs. And don't forget to add default case there

PR #1964, opened 2023-02-06
stateful-clones - first try

RufatRM:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+        }
+        arrayOfStates.push({ ...copyOfState });
+        break;
+      }
+
+      case 'clear': {
+        for (const key in copyOfState) {
+          delete copyOfState[key];
+        }
+        arrayOfStates.push({ ...copyOfState });
```
polosanya, line 39: You do the same in all cases. Just push it after switching all cases.
Also, don't forget about the default case for error handling

polosanya (changes requested):
Almost done!

polosanya (approved):
Well done!

PR #1965, opened 2023-02-06
Solution

Zavalishyn-Oleksandr:
I haven't checked it before, but I never use index in code, so I can use only element in forEach

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const copy = { ...state };
+
+  actions.forEach((elem, index) => {
+    switch (elem.type) {
+      case ('addProperties'):
+        const a = Object.assign({}, copy, elem.extraData);
+
+        Object.assign(copy, elem.extraData);
+        result.push(a);
+        break;
+
+      case ('removeProperties'):
+        elem.keysToRemove.forEach(rem => {
+          delete copy[rem];
+        });
+
+        const b = Object.assign({}, copy);
+
+        result.push(b);
+        break;
+
+      case ('clear'):
+        for (const key in copy) {
+          delete copy[key];
+        }
+
+        const c = Object.assign({}, copy);
+
+        result.push(c);
+        break;
+    }
+  });
+
+  return result;
 }
 
 module.exports = transformStateWithClones;
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
     "eslint": "^5.16.0",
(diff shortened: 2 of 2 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+
+        const b = Object.assign({}, copy);
+
+        result.push(b);
+        break;
+
+      case ('clear'):
+        for (const key in copy) {
+          delete copy[key];
+        }
```
loralevitska, line 35: It's just copy so `result.push({})`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const copy = { ...state };
+
+  actions.forEach((elem, index) => {
+    switch (elem.type) {
+      case ('addProperties'):
+        const a = Object.assign({}, copy, elem.extraData);
```
loralevitska, line 16: What is `a`? Use more meaningful variable names

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+        break;
+
+      case ('clear'):
+        for (const key in copy) {
+          delete copy[key];
+        }
+
+        const c = Object.assign({}, copy);
+
+        result.push(c);
```
loralevitska, line 39: You're in a loop, so push your data at the end of the switch. You don't have to do it every time

loralevitska (changes requested):
- Add `default` case using `switch/case`

Zavalishyn-Oleksandr:
Now I don't use unnecessary variables and do result.push() correctly, but I didn't catch why do I need default case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const copy = { ...state };
+
+  actions.forEach(elem => {
```
DarkMistyRoom, line 13:
```suggestion
  actions.forEach(action => {
```

DarkMistyRoom (changes requested):
You need to use the default case to handle the wrong data. We can't be sure about the correctness of input data, so it is good practice to inform a user about mistakes.

And better ask questions in the chat, then you won't wait so long

Zavalishyn-Oleksandr:
I've got some problems with internet connection, that's why I usually don't use fe_chat, but you're right, I should try to use it more often

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        });
+        break;
+
+      case ('clear'):
+        for (const key in copy) {
+          delete copy[key];
+        }
+        break;
+
+      default: return 'Please, enter the valid data';
```
nazarmatsevych, line 31:
```suggestion
      default: return 'Unknown action type';
```

and in this case better to use [throw new Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw)

nazarmatsevych (approved):
Good work,
just small suggestion in comments

PR #1967, opened 2023-02-07
Solution

OlhaKissa:


src/transformStateWithClones.js
```diff
@@ -1,13 +1,48 @@
...
+        }
+        break;
+      case ('addProperties'):
+        Object.assign(copyOfState, action.extraData);
+        break;
+      case ('removeProperties'):
+        for (const prop of action.keysToRemove) {
+          delete copyOfState[prop];
+        }
+        break;
```
Viktor-Kost, line 40: it is a good practice when you add a default behavior for switch statement. Yet, if you don't have such one - just `throw an Error `with text like `"unexpected action"`

Viktor-Kost (approved):
Nice!

PR #1970, opened 2023-02-08
solution

MSA1991:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
-  // write code here
+  let copyState = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        copyState = {
+          ...copyState, ...action.extraData,
+        };
```
Anastasiia-Svintsova, line 18: Also, you can use Object.assign here
```

File: src/transformStateWithClones.js [hunk 2]
```diff
@@ -7,7 +7,33 @@
...
+      case 'clear':
+        copyState = {};
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (copyState[key]) {
+            delete copyState[key];
+          }
+        }
```
Anastasiia-Svintsova, line 30: Add a default case to the switch/case construction to prevent different bugs

Anastasiia-Svintsova (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(copyState, action.extraData);
+        break;
+
+      case 'clear':
+        copyState = {};
+        break;
+
+      default:
```
ivbutsykin, line 23: No no, it's a critical mistake here. For example, if I use not defined action it will try to remove my fields, it's bad)
I think you a little bit misunderstand this comment:
https://github.com/mate-academy/js_stateful-clones/pull/1970#discussion_r1101334568

Here was meaning next:
```
      default:
        throw new Error('Unexpected action');
```

Please return back your previous case and add code above ^ for default case.

ViktorHavryliuk (changes requested):
Please implement this feedback
image

ViktorHavryliuk (approved):

PR #1972, opened 2023-02-09
Develop

Kocha1991:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let newState = { ...state };
+  const newArr = [];
+
+  for (const action of actions) {
```

loralevitska (changes requested):
Almost perfect) Let's improve your styles

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          delete newState[key];
+        }
+        break;
+
+      case 'clear':
+        newState = {};
+        break;
```
DarkMistyRoom, line 29: it isn't the best option, because it will be a different object. better to `delete `all properties

DarkMistyRoom (approved):

PR #1973, opened 2023-02-09
Solution

luka4ok:
Please, check my solution.

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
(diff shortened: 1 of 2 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const result = [];
+
+  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+          delete newState[key];
+        }
+
+        break;
+
+      case 'clear':
+        for (const key in newState) {
+          delete newState[key];
+        };
+
```
loralevitska, line 30: It's only copy, so you don't need loop just `newState  = {}`

loralevitska (changes requested):
Almost done) Let's improve

ivbutsykin (approved):
Good job!

PR #1974, opened 2023-02-09
Solution

yanZibierov:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,42 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const data in action.extraData) {
+          const prop = action.extraData[data];
+
+          clone[data] = prop;
+        }
+        resultArr.push(clone);
```
vlad-demchuk, line 22: it would be clearer to use Object.assign method
also, you can push after cases

vlad-demchuk (changes requested):

src/transformStateWithClones.js
OlEzhA0, line 22: why do we need to push on each case? just push 1 time after cases

src/transformStateWithClones.js
```diff
@@ -8,6 +8,42 @@
...
+        clone = Object.assign({}, resultArr[resultArr.length - 1]);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete clone[key];
+        }
+        resultArr.push(clone);
+        clone = Object.assign({}, resultArr[resultArr.length - 1]);
```
OlEzhA0, line 31:
```suggestion
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,42 @@
...
+        break;
 
+      case 'clear':
+        for (const key in clone) {
+          delete clone[key];
+        }
+        resultArr.push(clone);
+        clone = Object.assign({}, resultArr[resultArr.length - 1]);
```
OlEzhA0, line 39:
```suggestion
       clone = {}
```

OlEzhA0 (changes requested):

etojeDenys (approved):
LGTM

etojeDenys (approved):

PR #1976, opened 2023-02-10
Solution

MarinaOlenchenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        break;
+
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        stateArray.push({ ...stateCopy });
+
+        break;
+
+      case 'removeProperties':
+
```
ViktorHavryliuk, line 31: No need to have empty line here

ViktorHavryliuk (changes requested):
Good work, but you also need to add default case to handlle error if action type is neither clear, addProperties nor removeProperties

ViktorHavryliuk (approved):

PR #1977, opened 2023-02-10
solution

Zelinskyi-Serhii:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const copy = { ...state };
+
+  for (const key in actions) {
+    if (actions[key].type === 'addProperties') {
```
kateryna-babina, line 14: Maybe it's better to move actions[key] to variable

kateryna-babina (changes requested):
good job, but I suggest one small changes

PR #1978, opened 2023-02-10
Solution

nazarkharkevych:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,61 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+
+  for (const action of actions) {
+    const lastItemInArray = result[result.length - 1];
```
Volodymyr-Melnyk-Fron10der, line 13:
```suggestion
    const currentStateObject = result[result.length - 1];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,61 @@
...
+        let addedProperties;
+
+        if (result.length === 0) {
+          addedProperties = Object.assign({}, state, action.extraData);
+        } else {
+          addedProperties
+            = Object.assign({}, lastItemInArray, action.extraData);
+        }
+
+        result.push(addedProperties);
```
Volodymyr-Melnyk-Fron10der, line 26: I suggest to short it a little bit.
If there is no elements in array your variable will be equal  `undefined` so you don't need to check result.length, just check is there a currentStateObject
```suggestion
        result.push(Object.assign({}, currentStateObject || state, action.extraData));
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,61 @@
...
+        break;
+
+      case 'removeProperties':
+        let removedProperties;
+
+        if (result.length === 0) {
+          removedProperties = Object.assign({}, state);
+        } else {
+          removedProperties = Object.assign({}, lastItemInArray);
+        }
```
Volodymyr-Melnyk-Fron10der, line 36: Same refactoring here + rename variable
```suggestion
        let object;
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,61 @@
...
+
+        if (result.length === 0) {
+          clearedProperties = Object.assign({}, state);
+        } else {
+          clearedProperties = Object.assign({}, lastItemInArray);
+        }
+
+        Object.keys(clearedProperties).forEach(key => {
+          delete clearedProperties[key];
+        });
```
Volodymyr-Melnyk-Fron10der, line 54: All same

Volodymyr-Melnyk-Fron10der (changes requested):
Hi! Let's improve your code:

ViktorHavryliuk (approved):

PR #1979, opened 2023-02-10
js_stateful-clones

JulyaPetrovskaya:


etojeDenys (changes requested):
you always need to add the default case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+
+      case 'clear':
+        for (const del in copy) {
+          delete copy[del];
+        }
+        break;
+
+      default:
+        break;
```
nazarmatsevych, line 32: Better to throw an error in default case for example:
`throw new Error('Unknown action type')`

nazarmatsevych (approved):
Good job!
just small suggestion

PR #1982, opened 2023-02-11
Solution

vshepel:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+        if (copyState[key]) {
+          delete copyState[key];
+        }
+      }
+    } else if (action.type === 'addProperties') {
+      Object.assign(copyState, action.extraData);
+    }
+
+    result.push({ ...copyState });
+  }
```
nazarmatsevych, line 29: Try to avoid nested if else statements, in this case better to use `switch/case`

nazarmatsevych (changes requested):

src/transformStateWithClones.js
vshepel: I tried to make a better solution, please check

etojeDenys (changes requested):
don`t forget to add `default case`

vshepel:
> don`t forget to add `default case`

I don't understand what I need to put in it? 

```js
default:
        console.log('No available type');
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
+  const result = [];
+
+  actions.map(({ type, keysToRemove, extraData }) => {
+    switch (type) {
+      case 'clear':
+        Object.keys(copyState).map(key => delete copyState[key]);
```
OlEzhA0, line 16:
```suggestion
        Object.keys(copyState).forEach(key => delete copyState[key]);
```

OlEzhA0: or
```suggestion
        copyState = {}
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+        break;
+
+      case 'removeProperties':
+        keysToRemove.map(key => copyState[key] && delete copyState[key]);
```
OlEzhA0, line 20:
```suggestion
        keysToRemove.forEach(key => copyState[key] && delete copyState[key]);
```

OlEzhA0 (changes requested):

src/transformStateWithClones.js
vshepel: @OlEzhA0 Can you argue why forEach and not map?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
+  const result = [];
+
+  actions.map(({ type, keysToRemove, extraData }) => {
```
Bohdan-Maksymiuk, line 13:
```suggestion
  actions.forEach(({ type, keysToRemove, extraData }) => {
```
in general here you should prefer `forEach` too, cause you don't return anything on each iteration, just do some logic

PR #1986, opened 2023-02-12
Solution

anna-akhalmosulishvili:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+      case 'addProperties':
+        const addProps = obj.extraData;
+
+        addCopy = {
+          ...stateCopy,
+          ...addProps,
+        };
+
+        clonesArr.push(addCopy);
+        stateCopy = Object.assign({}, addCopy);
```
DarkMistyRoom, line 25:
```suggestion
        Object.assign(stateCopy, action.extraData);
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clonesArr = [];
+  let stateCopy = { ...state };
+  let addCopy;
+
+  for (const obj of actions) {
```
DarkMistyRoom, line 14:
```suggestion
  for (const action of actions) {
```

src/transformStateWithClones.js
DarkMistyRoom: that is enough

src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+        break;
+
+      case 'removeProperties':
+        const removeProps = obj.keysToRemove;
+
+        for (const key of Object.keys(stateCopy)) {
+          if (removeProps.indexOf(key) !== -1) {
+            delete stateCopy[key];
+          }
+        }
```
DarkMistyRoom, line 36: ok, but you can do it shorter
```suggestion
        action.keysToRemove.forEach((key) => delete stateCopy[key]);
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+
+        clonesArr.push(stateCopy);
```
DarkMistyRoom, line 38: make the push just once in the end of each iteration

src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+        break;
+
+      default:
```
DarkMistyRoom, line 41: it is a bad idea to use default in this way. better specify the value and use the default case to handle errors

src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+        const emptyObj = {};
+
+        stateCopy = Object.assign({}, emptyObj);
```
DarkMistyRoom, line 44: why do you need to assign an empty object to another empty object?

DarkMistyRoom: better delete all properties from the stateCopy to keep the empty state for the next iterations

DarkMistyRoom (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+        Object.assign(stateCopy, action.extraData);
+        break;
+
+      case 'removeProperties':
+        action.keysToRemove.forEach((key) => delete stateCopy[key]);
+        break;
+
+      case 'clear':
+        Object.keys(stateCopy).forEach(key => delete stateCopy[key]);
+        break;
```
Viktor-Kost, line 25: it is a good practice when you add a default behavior for switch statement. Yet, if you don't have such one - just `throw an Error` with text like `"unexpected action"`

PR #1987, opened 2023-02-12
stateful clones solution

kateryna-babina:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      for (const key in stateClone) {
+        delete stateClone[key];
+      }
+    }
+
+    if (obj.type === 'removeProperties') {
+      for (const key of obj.keysToRemove) {
+        delete stateClone[key];
+      }
+    }
```
timoxa2t, line 28:
```suggestion
    switch (obj.type) {
      case 'addProperties' :
          Object.assign(stateClone, obj.extraData);
          break;
       case 'clear': 
          for (const key in stateClone) {
            delete stateClone[key];
          }
          break;
       case 'removeProperties' : 
          for (const key of obj.keysToRemove) {
            delete stateClone[key];
          }
   }  
```
You could change if construction to switch

timoxa2t (approved):
Good job

illiarerun (approved):

PR #1990, opened 2023-02-13
Solution

DaveBeetle:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      case 'removeProperties':
+        for (const keys of action.keysToRemove) {
+          delete clone[keys];
+        }
+        break;
+
+      case 'clear':
+        for (const key in clone) {
+          delete clone[key];
+        }
```
Viktor-Kost, line 28: Instead of additional loop, you can just rewrite you `clone` object
Hint: change clone variable to `let` 
```suggestion
        clone = {}
```

Viktor-Kost (approved):
Good job!
Take a look at an advise I left at comment

PR #1991, opened 2023-02-13
Solution

timoxa2t:


Andr1yK (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+      case 'addProperties':
+        Object.assign(clone, action.extraData);
+        break;
+      case 'removeProperties':
+        removeProps(clone, action.keysToRemove);
+        break;
+      case 'clear':
+        removeProps(clone, Object.keys(clone));
+    }
+    history.push({ ...clone });
```
tetiana-ivanchenko, line 24: add default case

tetiana-ivanchenko (changes requested):

src/transformStateWithClones.js
timoxa2t: and what would it be?

PR #2003, opened 2023-02-14
Stateful Clones - Solution

Yevhenii0536:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultArray = [];
+  const copyOfState = JSON.parse(JSON.stringify(state));
+
```
MarkBabich, line 12: [Checklist](https://github.com/mate-academy/js_stateful-clones/blob/master/checklist.md) p.1

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const item in action.extraData) {
+          copyOfState[item] = action.extraData[item];
+        }
+        resultArray.push({ ...copyOfState });
+        break;
```
MarkBabich, line 20: You repeat this part. [Checklist](https://github.com/mate-academy/js_stateful-clones/blob/master/checklist.md) p.4

MarkBabich (changes requested):
Works fine. But not all checklist points were included :)

src/transformStateWithClones.js
Yevhenii0536: so what? i'am not mutate object :D 🍡

src/transformStateWithClones.js
Yevhenii0536: i push element on every step. if each case doesn't work, my switch will break without any actions. if i'll delete push actions from every case and put him to the floor of loop - it will break my function

src/transformStateWithClones.js
MarkBabich: ok, up to you. I just noticed that there are two ways in the checklist to do that. I don't know this way, sorry :)

Andr1yK: Wow... why do u need `JSON parse/stringify`?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+
+      case 'clear':
+        for (const item in copyOfState) {
+          delete copyOfState[item];
+        }
+        break;
+
+      default:
+        return '';
```
Andr1yK, line 34: let's throw an error here (if we do not support a certain type of action)

Andr1yK (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
-  // write code here
+  const stateHistory = [];
+  const copyOfState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const item in action.extraData) {
+          copyOfState[item] = action.extraData[item];
+        }
```
Andr1yK, line 18: U can use Object.assign or spread operator

Andr1yK (approved):

PR #2007, opened 2023-02-14
task implemented

andriibezkorovainyi:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const tempState = {
```
DenysDubas, line 11: it is not clear what this name means

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          delete tempState[key];
+        }
+        break;
+
+      default:
+        continue;
```
DenysDubas, line 36: it’s better to add some kind of error, because the cases should have worked for us

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
DenysDubas, line 10: result of what? result it's a bad name

DenysDubas (changes requested):

illiarerun (approved):

PR #2010, opened 2023-02-14
Solution

arkushenko-anna:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newObj = { ...state };
+  const array = [];
+
```
yaroslav-vasilyev, line 12: What does array mean?
Better to use names, what others devs can understand
Maybe actionHistory

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newObj = { ...state };
```
yaroslav-vasilyev, line 10: Same as previous, newObj don't understandable

yaroslav-vasilyev (approved):
Good Job!
Look at comments and fix some fix to get better!!!)

src/transformStateWithClones.js
illiarerun: agree, fix this variable's name

src/transformStateWithClones.js
illiarerun: here too

illiarerun (changes requested):

illiarerun (approved):

PR #2012, opened 2023-02-14
Solution

ksavelyeva:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      case 'clear' :
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
+    }
+
+    const assign = { ...stateCopy };
+
+    stateHistory.push(assign);
```
MaksymMaliuk, line 33: You don't need to create a new variable. Just use `{ ...stateCopy }` in `push`
```suggestion

    stateHistory.push( { ...stateCopy });
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
+        break;
+      case 'clear' :
```
MaksymMaliuk, line 24: I'm not sure, but i think you need to add blank line between different cases.

MaksymMaliuk (changes requested):
It's nice work, but needs some fixes

MaksymMaliuk (commented):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
+        break;
+
+      case 'clear' :
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
```
Andr1yK, line 28:
```suggestion
       stateCopy = {}
```

Andr1yK (approved):

PR #2021, opened 2023-02-15
solution

Dmytro-Kretov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const newObj = Object.assign({}, state);
```
nazarmatsevych, line 11:
```suggestion
  const newObj = Object.assign({}, state);
```

Use better naming,  for example stateClone or stateCopy

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+
+  for (const object of actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+    switch (object.type) {
```
nazarmatsevych, line 14: Use destructuring of object properties

nazarmatsevych (changes requested):
Good work,
check my commets pls

nazarmatsevych (approved):
👍

PR #2030, opened 2023-02-15
Stateful clones

VadKol:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let forCloneState = { ...state };
```
Andr1yK, line 10: why do we need `for` here?)

```suggestion
  let stateCopy = { ...state };
```

Andr1yK (approved):

PR #2034, opened 2023-02-16
Solution

kostiantin-hudzenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
+        break;
+
+      default:
+        stateCopy = {};
+        break;
```
Viktor-Kost, line 27: it's not a good practice to use such actions as default. So use a separated case for this action type. 

for default you can  just` throw an Error` with text like "`unexpected action`"

Viktor-Kost (changes requested):
Good job!

Need to fix one moment

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        }
+        break;
+
+      case 'clear':
+        stateCopy = {};
+        break;
+
+      default:
+        break;
+    }
```
etojeDenys, line 31: for default, you can just throw an Error with text like "unexpected action"

etojeDenys (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+          delete stateCopy[key];
+        }
+        break;
+
+      case 'clear':
+        stateCopy = {};
+        break;
+
+      default:
+        return 'unexpected action';
```
TarasHoliuk, line 30: not critical in this task, just FYI (for the future):
you may **throw** the new Error with such message ([mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error)) in this way:
```suggestion
        throw new Error('unexpected action');
```

TarasHoliuk (approved):
Approved with one comment (just FYI - no need to re-request review)

PR #2035, opened 2023-02-17
Solution

MTS13:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+  actions.forEach(action => {
+    switch (action.type) {
+      case ('addProperties') :
+        Object.assign(copy, action.extraData);
+        break;
+
+      case ('removeProperties') :
+        Object.keys(action.keysToRemove).forEach(function(key) {
+          delete copy[action.keysToRemove[key]];
+        });
```
etojeDenys, line 22:
```suggestion
        for (const key of action.keysToRemove) {
          delete copy[key];
        }
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+        break;
+
+      case ('clear') :
+        for (const key in copy) {
+          delete copy[key];
+        }
```
```suggestion
       copy = {};
       break;
```

etojeDenys (changes requested):
also add the default case

for default you can just throw an Error with text like "unexpected action"

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let copy = { ...state };
+  const result = [];
+
+  actions.forEach(action => {
```
Volodymyr-Melnyk-Fron10der, line 13:
```suggestion
  actions.forEach(({ ... }) => {
```

Volodymyr-Melnyk-Fron10der (approved):
Good✅ But try always use decomposition where it's possible

PR #2042, opened 2023-02-20
Solution for Transform State with Clones

ctk12:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const newObj = { ...state };
+  const resArr = [];
```
polosanya, line 12: Let's use cleaner names for the variables.
Explain what resArr includes
image

src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
...
+        keysToRemove.forEach(key => {
+          delete newObj[key];
+        });
+        break;
+
+      case 'clear' :
+        Object.keys(newObj).forEach(key => {
+          delete newObj[key];
+        });
+        break;
```
polosanya, line 30: image

src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const newObj = { ...state };
+  const resArr = [];
+
+  for (const { type, extraData, keysToRemove } of actions) {
```
```suggestion
  for (action of actions) {
  const { 
    type, 
    extraData, 
    keysToRemove, 
  } = action
```
Your code should be not just short, but also understandable and descriptive

polosanya (changes requested):
Almost done!

qoa11a (approved):

PR #2043, opened 2023-02-21
js_stateful-clones - done✅

2pasha:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state2) {
+          delete state2[key];
+        }
+        result.push({ ...state2 });
+        break;
+    }
+  }
```
IvanFesenko, line 38: `result.push({ ...state2 });` repeats 3 times - do it after switch
```suggestion
    switch (type) {
      case 'addProperties':
        Object.assign(state2, action.extraData);
        break;

      case 'removeProperties':
        const { keysToRemove } = action;

        for (const key of keysToRemove) {
          delete state2[key];
        }
        break;

      case 'clear':
        for (const key in state2) {
          delete state2[key];
        }
        break;
    }
    result.push({ ...state2 });
  }
```

IvanFesenko (approved):

PR #2046, opened 2023-02-21
Clones solution v.1

annperetiatko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+        const keysToRemove = action.keysToRemove;
+
+        for (const removeItem in keysToRemove) {
+          delete newState[keysToRemove[removeItem]];
+        }
+        break;
+      case 'clear':
+        newState = {};
+        break;
+    }
```
mykhalenych, line 29: switch always must has default case

mykhalenych (approved):

PR #2049, opened 2023-02-22
created clone object

JasterHim:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const resalt = [];
```
```suggestion
  const result = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
```
sTorba24, line 10:
```suggestion
  const stateCopy= { ...state };
```

sTorba24: it's not new state, it's just a copy :)

sTorba24 (approved):
Good work! little comments i made :)

PR #2055, opened 2023-02-25
Solution

vanya-kalyenichenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const workState = { ...state };
+
+  for (const i of actions) {
+    for (const key in i) {
```
IvanFesenko, line 14: Use correct naming. Second loop is redudant. Use switch/case for `action.type`
```suggestion
  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
+        result.push({ ...workState });
+      }
+
+      if (i[key] === 'clear') {
+        for (const del in workState) {
+          delete workState[del];
+        }
+
+        result.push({ ...workState });
+      }
```
IvanFesenko, line 42: Do not repeat your self push result after all iterations
```suggestion

      }

      if (i[key] === 'removeProperties') {
        for (const ch of remove) {
          if (ch in workState) {
            delete workState[ch];
          }
        }

      }

      if (i[key] === 'clear') {
        for (const del in workState) {
          delete workState[del];
        }
     
      }
      result.push({ ...workState });
```

IvanFesenko (changes requested):
Refactor required. Check my comments

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  const result = [];
+  const workState = { ...state };
+
+  for (const action of actions) {
+    const add = action.extraData;
+    const remove = action.keysToRemove;
```
ivanchuk-yurii, line 15: imho these are redundant

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        for (const x in add) {
+          workState[x] = add[x];
+        }
+
+        break;
+
+      case 'removeProperties':
+        for (const ch of remove) {
```
ivanchuk-yurii, line 26: what does `ch` mean? it's better to use `key` for example

ivanchuk-yurii (approved):

PR #2060, opened 2023-02-27
 js_stateful-clones

Kirillets:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+      for (const key of command.keysToRemove) {
+        delete copy[key];
+      }
+    }
+
+    if (command.type === 'clear') {
+      for (const key in copy) {
+        delete copy[key];
+      }
+    }
```
Viktor-Kost, line 28: use switch case instead. It'll be cleaner

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copy = { ...state };
+  const arr = [];
+
+  for (const command of actions) {
```
Viktor-Kost, line 13: use singular noun from your plural one. It will give more sense 
```suggestion
  for (const action of actions) {
```

Viktor-Kost (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
-  // write code here
+  const copy = { ...state };
+  const arr = [];
+
+  for (const { type, extraData, keysToRemove } of actions) {
+    switch (type) {
+      case 'addProperties':
+        Object.assign(copy, extraData);
+        break;
+      case 'removeProperties':
```
SerhiiKirik, line 18:
```suggestion
        break;
        
      case 'removeProperties':
```

SerhiiKirik (approved):
Well done 🚀

PR #2062, opened 2023-02-28
Solution

AnnaLiman:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newarr = [];
```
qoa11a, line 10:
```suggestion
  const newArr = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+  const statecopy = {
```
qoa11a, line 11:
```suggestion
  const stateCopy = {
```

qoa11a (changes requested):
use camelCase for naming variables in JS

PR #2066, opened 2023-03-01
solution

YuriyLarin:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        break;
+
+      case 'clear':
+        for (const key in stateClone) {
+          delete stateClone[key];
+        }
+        break;
+
+      default:
+        break;
```
DarkMistyRoom, line 32: would be great to handle errors and show message to user that action type is invalid in default case

DarkMistyRoom (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+      case 'clear':
+        for (const key in stateClone) {
+          delete stateClone[key];
+        }
+        break;
+
+      default:
+        return 'invalid data';
```
DarkMistyRoom, line 32: better `trow new Error` or `alert`, but it is ok for now

DarkMistyRoom (approved):

PR #2069, opened 2023-03-03
Solution for Cloning and transforming state

julmak:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+      for (const key of action.keysToRemove) {
+        delete newState[key];
+      }
+
+      result.push({ ...newState });
+    } else if (action.type === 'addProperties') {
+      Object.assign(newState, action.extraData);
+
+      result.push({ ...newState });
+    }
```
qoa11a, line 32: You should use switch construction here. And you can make result.push only once at the end of all if statements (at the end of all cases in case of switch).

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete newState[key];
+        }
+        break;
+
+      case 'addProperties':
+        Object.assign(newState, action.extraData);
```
nazarmatsevych, line 26: Good job, you can also use here destructuring of object properties

nazarmatsevych (approved):

PR #2073, opened 2023-03-04
solution

vlad-fil:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClone = { ...state };
```
```suggestion
  let stateClone = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      case 'removeProperties':
+        for (const index in action.keysToRemove) {
+          delete stateClone[action.keysToRemove[index]];
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateClone) {
+          delete stateClone[key];
+        }
```
```suggestion
        stateClone = {};
```

etojeDenys (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+          delete stateClone[action.keysToRemove[index]];
+        }
+        break;
+
+      case 'clear':
+        stateClone = {};
+        break;
+
+      default:
+        result = 'action type not found';
```
etojeDenys, line 30: I think this default case will break our logic, so it`s better to just throw the error here

etojeDenys (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+          delete stateClone[action.keysToRemove[index]];
+        }
+        break;
+
+      case 'clear':
+        stateClone = {};
+        break;
+
+      default:
+        return 'action type not found';
```
etojeDenys, line 30: let`s don't do anything with the result or return here, just throw the error

etojeDenys (changes requested):

mykhalenych (approved):

PR #2076, opened 2023-03-06
Solution

92BS:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const arr = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(newState, action.extraData);
+        arr.push({ ...newState });
```
qoa11a, line 17: You should do this at the end of all cases as you do it in every case anyway.

qoa11a (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      case 'clear':
+        for (const keys in newState) {
+          delete newState[keys];
+        }
+        arr.push({ ...newState });
+        break;
+
+      case 'addProperties':
+        Object.assign(newState, action.extraData);
+        arr.push({ ...newState });
```
qoa11a, line 31: I meant that you should push into the array just once. No need in writing it in all cases.

qoa11a (changes requested):

witflash (changes requested):
Check the last comments, please

agniya-i (approved):

PR #2080, opened 2023-03-07
Solution

MaksymKos:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+        const { keysToRemove } = action;
+        const newState = {};
+
+        for (const key in currentState) {
+          if (!keysToRemove.includes(key)) {
+            newState[key] = currentState[key];
+          }
+        }
+
+        currentState = newState;
```
anstsot, line 31: this code could be simplified to work only with `keysToRemove`
```suggestion
        for (const key of keysToRemove) {
          delete currentState[key];
        }
```

anstsot (approved):
Looks good, approving
I've added comment that can improve your code

PR #2081, opened 2023-03-07
Solution

NVlasenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,67 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const properties = {
+    addProperties: 'addProperties',
+    clear: 'clear',
+    removeProperties: 'removeProperties',
+  };
+  const newObj = { ...state };
+  let obj = {};
+  const newArray = [];
```
TarasHoliuk, line 17: bad name
it should explain what is it

src/transformStateWithClones.js
```diff
@@ -7,7 +7,67 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const properties = {
+    addProperties: 'addProperties',
+    clear: 'clear',
+    removeProperties: 'removeProperties',
+  };
+  const newObj = { ...state };
+  let obj = {};
```
TarasHoliuk, line 16: you don't need this object
working only with `newObject` (bad name) will be enought

src/transformStateWithClones.js
```diff
@@ -7,7 +7,67 @@
...
+  const newArray = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case properties.addProperties:
+        assignNewDataToState(newObj, action);
+        obj = writeToArray(obj, newObj, newArray);
```
TarasHoliuk, line 23: to complicated
you can just mutate your initial copy (`newObject`) and then push his copy to array

src/transformStateWithClones.js
```diff
@@ -7,7 +7,67 @@
...
+}
+
+function writeToArray(newObject, obj, newArray) {
+  let object = { ...obj };
+
+  newArray.push(object);
+
+  object = {};
+
+  return object;
```
TarasHoliuk, line 54: redundant

src/transformStateWithClones.js
```diff
@@ -7,7 +7,67 @@
...
+        }
+        obj = writeToArray(obj, newObj, newArray);
+        break;
+    }
+  }
+
+  return newArray;
+}
+
+function writeToArray(newObject, obj, newArray) {
```
TarasHoliuk, line 47: you don't use `newObject` at all here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,67 @@
...
+
+function deleteFromState(newObj, key) {
+  delete newObj[key];
+}
+
+function getRemoveKeys(action) {
+  return action.keysToRemove;
+}
+
+function getSateKeys(myObject) {
```
TarasHoliuk, line 69: having separate functions for some actions is ok, but in this case, it is overcoding 🙂

TarasHoliuk (changes requested):

src/transformStateWithClones.js
NVlasenko: @TarasHoliuk if i do newArray.push(copyOfObject)
then the array will be overwritten. And at the end the array will be with 3 identical objects.
Or I misunderstood

NVlasenko: I changed newObject to copyOfObject

src/transformStateWithClones.js
```diff
@@ -7,7 +7,67 @@
...
+    clear: 'clear',
+    removeProperties: 'removeProperties',
+  };
+  const newObj = { ...state };
+  let obj = {};
+  const newArray = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case properties.addProperties:
```
anstsot, line 21: there is no need to create a separate object for action namings. you can use the name directly here
```suggestion
      case 'addProperties':
```

src/transformStateWithClones.js
anstsot, line 23: the code could be simplified to:
```suggestion
        obj = { ...newObj, ...action.extraData };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,67 @@
...
+        assignNewDataToState(newObj, action);
+        obj = writeToArray(obj, newObj, newArray);
+        break;
+
+      case properties.removeProperties:
+        for (const key of getRemoveKeys(action)) {
+          deleteFromState(newObj, key);
+        }
+        obj = writeToArray(obj, newObj, newArray);
```
anstsot, line 30: same here. no need to use `writeToArray`. you can always works with your `newObj` (it is copy of your `state` object)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,67 @@
...
+        break;
+
+      case properties.clear:
+        const keysState = getSateKeys(newObj);
+
+        for (const key of keysState) {
+          deleteFromState(newObj, key);
+        }
+        obj = writeToArray(obj, newObj, newArray);
```
anstsot, line 39: here you san simplify by assigning the empty object
```suggestion
        obj = {};
```

anstsot (changes requested):
Good job, but there are few fixes to improve your code.

src/transformStateWithClones.js
NVlasenko: @anstsot if you write the changed newObj to the array, then the result is an array with 3 identical objects (the array is overwritten)

IvanFesenko (changes requested):
Not fixed previous comments. If you have any questions - ask them in fe_chat

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  let copyOfState = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        assignNewDataToState(copyOfState, action);
+        result.push({ ...copyOfState });
```
sTorba24, line 17: you can do it outside of switch :)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          deleteFromState(copyOfState, key);
+        }
+        result.push({ ...copyOfState });
```
sTorba24, line 24: it can be removed then

sTorba24 (changes requested):
Good work! lets improve it :)

AlisaVlasova (approved):

PR #2082, opened 2023-03-07
My solution

hyrych-max:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [state];
```
OlEzhA0, line 10:
```suggestion
  const result = [];
```

why do we need add this to array and slice when return?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+
+  for (let i = 0; i < actions.length; i++) {
+    const prevState = result[i];
```
OlEzhA0, line 13: why is it here?
you need to keep your `currentState` outside loop

`let currentState = { ...state }`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+    let newState;
+
+    switch (actions[i].type) {
+      case 'addProperties':
+        newState = Object.assign({}, prevState, actions[i].extraData);
```
OlEzhA0, line 18: here you can use 
`currentState = { ...currentState, ...actions[i].extraData }`
please check how you can optimize your code with this hints

OlEzhA0 (changes requested):

hyrych-max:
I don't know, but i think this solution is better than previous one. Anyway, i'm looking forward to your comments.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
polosanya, line 10: Let's use more descriptive names than just "result"

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+
+  let currentState = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
```
```suggestion
  for (const action of actions) {
```
I think it would be cleaner and easier for you.
You can also destructure the action to get all the needed props of it (type, extraData, keysToRemove)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+    switch (actions[i].type) {
+      case 'addProperties':
+        currentState = {
+          ...currentState, ...actions[i].extraData,
+        };
```
polosanya, line 19: It is better to use [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        const keysToRemove = actions[i].keysToRemove;
+        const { ...newState } = currentState;
+
+        keysToRemove.forEach(key => delete newState[key]);
+        currentState = newState;
+        break;
+
+      case 'clear':
+        currentState = {};
+        break;
```

polosanya (changes requested):
Almost done!

src/transformStateWithClones.js
hyrych-max: @polosanya I'm sorry, but in the previous example, the mentor told me that it's better to use the spread operator instead of Object.assign.
There i use Object.assign like this:
case 'addProperties':
        newState = Object.assign({}, prevState, actions[i].extraData);
        
Tell me what better to use - first or second example?

src/transformStateWithClones.js
anstsot: still not fixed. please add the default case for switch/case.

src/transformStateWithClones.js
anstsot: still not fixed. let's use `for of` construction

anstsot (changes requested):
Let's fix previous comments.
if you have any issues or questions feel free to ask then in `fe_chat`

tetiana-ivanchenko (changes requested):

Anastasiia-Svintsova (approved):

PR #2086, opened 2023-03-08
add solution

allagutsul27:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+          if (clone.hasOwnProperty(prop)) {
+            delete clone[prop];
+          }
+        }
+        break;
+
+      default:
+        for (const key in clone) {
+          delete clone[key];
+        }
```
slavanosatskyi, line 30: i believe it would be okay to make clone a non-constant variable and just do here clone = {}

slavanosatskyi (approved):
there is one minor comment, but in general it looks nice

PR #2095, opened 2023-03-13
Func to change object`s copy added

DenysKrysanov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        break;
+
+      case 'clear' :
+        for (const key in copyState) {
+          delete copyState[key];
+        }
+        break;
+
+      default:
+        return Error;
```
PavloReutskiy, line 32: I'm not sure if you need this in our case

PavloReutskiy (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
+  const resultArray = [];
```
jstmpelowycz, line 11: let's give it another name

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+      case 'clear' :
+        for (const key in copyState) {
+          delete copyState[key];
+        }
+        break;
+
+      default:
+        return Error;
+    }
+    resultArray.push({ ...copyState });
```
jstmpelowycz, line 34:
```suggestion
    }

    resultArray.push({ ...copyState });
```

jstmpelowycz (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+
+      case 'clear' :
+        for (const key in copyState) {
+          delete copyState[key];
+        }
+        break;
+
+      default:
+        return Error;
```
jstmpelowycz, line 32: suggestion — better to do so
```suggestion
        return new Error(...);
```

jstmpelowycz (approved):

PR #2100, opened 2023-03-13
Solution

igorkruz:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const result = [];
```
AnyaLyaLya, line 11: It is bad example to name variables!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(newState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete newState[key];
+        }
```
AnyaLyaLya, line 22: use proper variable names in for of loop, please use something like 'property'

AnyaLyaLya (changes requested):
Great! But you can improve your code, please do that

src/transformStateWithClones.js
jstmpelowycz: agree)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+          delete stateCopy[property];
+        }
+        break;
+
+      case 'clear':
+        for (const property in stateCopy) {
+          delete stateCopy[property];
+        }
+        break;
+      default: throw new Error('invalid action type');
```
jstmpelowycz, line 30:
```suggestion
        break;

      default: 
        throw new Error('invalid action type');
```

jstmpelowycz (changes requested):

VinnieJ-2k20 (approved):
Good job 👍

PR #2101, opened 2023-03-14
transform clones function was created

Vasyl-Zhyliakov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const array = [];
+  const stateClone = { ...state };
+
+  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+          }
+        }
+        array.push({ ...stateClone });
+        break;
+
+      case 'clear':
+        for (const key in stateClone) {
+          delete stateClone[key];
+        }
+        array.push({ ...stateClone });
```
loralevitska, line 33: Remove redundant loop. In this task state is just copy
```suggestion
        stateClone = {}
        array.push({});
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+    }
```
loralevitska, line 35: Add default case

loralevitska (changes requested):
Good job! There are things that should be fixed:

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const array = [];
+  let stateClone = { ...state };
+
+  for (const { type, extraData, keysToRemove } of actions) {
+    switch (type) {
+      case 'addProperties':
+        Object.assign(stateClone, extraData);
+        array.push({ ...stateClone });
```
DarkMistyRoom, line 17: no need to place `push `in each `case`. just do it once in the end of iteration

DarkMistyRoom (changes requested):
let's make a little improvment

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const array = [];
+  let stateClone = { ...state };
+
+  for (const { type, extraData, keysToRemove } of actions) {
+    let newStateClone = { ...stateClone };
```
ivbutsykin, line 14: No need copy already copied state)
You can simple modify your `stateClone` object.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+          }
+        }
+        break;
+
+      case 'clear':
+        newStateClone = {};
+        break;
+
+      default:
+        break;
```
ivbutsykin, line 34: I think better decision will be throw error here:
```
default:
  throw new Error('Unexpected action');
```

ivbutsykin (approved):
Good job!
But rewrite without `newStateClone` object

PR #2110, opened 2023-03-15
Solution

KirillMaslov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
```
OleziO, line 7: Remove all comments from the code.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+  const result = [];
+  const newObject = { ...state };
+
+  for (const object of actions) {
+    switch (object.type) {
+      case 'addProperties':
+        const { extraData } = object;
+
+        for (const key in extraData) {
+          newObject[key] = extraData[key];
```
OleziO: Just do `Object.assign()`

[More info.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+        }
+        break;
+
+      case 'removeProperties':
+        const { keysToRemove } = object;
+
+        for (let i = 0; i < keysToRemove.length; i++) {
```
OleziO: The better way is to use `forEach`

OleziO (changes requested):

DarkMistyRoom (approved):
Great!

PR #2112, opened 2023-03-15
solution

artem-kyrylenko:


Vlad-Lischenovskiy (approved):
Great job!!!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
+  const resultingArray = [];
```
jstmpelowycz, line 11: I'd rather give it another name

jstmpelowycz (changes requested):
the comment down below + those I've left to the previous task

src/transformStateWithClones.js
artem-kyrylenko: fixed

PR #2118, opened 2023-03-15
Solve the task using spread operator

Hamuud:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
+  const multipleStateArr = [];
```
jstmpelowycz, line 11: what about it?
```suggestion
  const stateHistory = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+
+      case 'clear': {
+        for (const data in copyState) {
+          delete copyState[data];
+        }
+        break;
+      }
+
+      default:
+        return `Sorry i don't know this action`;
```
jstmpelowycz, line 35: just a suggestion — you could've thrown an error

jstmpelowycz (changes requested):

PR #2119, opened 2023-03-15
transformStateWithClones function were created

MariaShmakova2014:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      result.push(Object.assign({}, clone));
+    }
+
+    if (action.type === 'clear') {
+      Object.keys(clone).forEach(key => {
+        delete clone[key];
+      });
+      result.push(Object.assign({}, clone));
+    }
+  }
```
IvanFesenko, line 32: do not repeat your self
```suggestion
  for (const action of actions) {
    if (action.type === 'addProperties') {
      Object.assign(clone, action.extraData);
    }

    if (action.type === 'removeProperties') {
      action.keysToRemove.forEach(property => {
        delete clone[property];
      });
    }

    if (action.type === 'clear') {
      Object.keys(clone).forEach(key => {
        delete clone[key];
      });
    }
    result.push({ ...clone });
  }
```

IvanFesenko (changes requested):
Small improvements

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const clone = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
OlEzhA0, line 14: would be better to use `switch/case`

OlEzhA0 (changes requested):

VitaliyBondarenko1982 (approved):

PR #2121, opened 2023-03-15
solution

Anastasia2403:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const currentState = { ...state };
+  const copiedState = [];
```
artem-kyrylenko, line 11: I would rename the array because it is not one copy of state

artem-kyrylenko (approved):
Well done!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const keyToRemove of action.keysToRemove) {
+          delete currentState[keyToRemove];
+        }
+        break;
+
+      case 'clear':
+        Object.keys(currentState).forEach(key => delete currentState[key]);
```
jstmpelowycz, line 26:
```suggestion
        Object.keys(currentState).forEach(key => {
          delete currentState[key];
        });
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        break;
+
+      default:
+        throw new Error('Invalid type of action');
+    };
+    copiedState.push({ ...currentState });
```
jstmpelowycz, line 32:
```suggestion
    };

    copiedState.push({ ...currentState });
```

jstmpelowycz (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(currentState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const keyToRemove of action.keysToRemove) {
+          delete currentState[keyToRemove];
+        }
```
jstmpelowycz, line 22: you could've used `forEach` here and here

jstmpelowycz (approved):
good job!

PR #2129, opened 2023-03-15
Solution

Anna-Kuzmenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        Object.assign(cloneState, action.extraData);
+        break;
+
+      case ('clear'):
+        for (const key in cloneState) {
+          delete cloneState[key];
+        }
+        break;
+
+      case ('removeProperties'):
```
Kavon-Skela, line 25:
```suggestion
      case 'addProperties':
        Object.assign(cloneState, action.extraData);
        break;

      case 'clear':
        for (const key in cloneState) {
          delete cloneState[key];
        }
        break;

      case 'removeProperties':
```
I think brackets are unnecessary

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        for (const keyToRemove of action.keysToRemove) {
+          delete cloneState[keyToRemove];
+        }
+        break;
+
+      default:
+        return cloneState;
```
Kavon-Skela, line 32: Why do you need to return something in default case?

Kavon-Skela (changes requested):
Good job overall, but needs to be improved according to the checklist
Checklist p.3. You don't have an error handler

Kavon-Skela (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const cloneState = JSON.parse(JSON.stringify(state));
```
jstmpelowycz, line 10: could have been simpler
```suggestion
  const cloneState = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+      case 'removeProperties':
+        for (const keyToRemove of action.keysToRemove) {
+          delete cloneState[keyToRemove];
+        }
+        break;
+
+      default:
+        throw new Error('Invalid action');
+    }
+    clonArr.push({ ...cloneState });
```
jstmpelowycz, line 34:
```suggestion
    }

    clonArr.push({ ...cloneState });
```

jstmpelowycz (approved):

PR #2131, opened 2023-03-15
Solution

vksazonov:


package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
```
kukla1989, line 19: you should only add code that you write, delete from commit package.json

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+
+      case 'clear':
+        for (const key in stateClone) {
+          delete stateClone[key];
+        }
+        break;
+
+      default:
+        break;
```
kukla1989, line 32: its do nothing so its meaningless to write it, they probably mean that you write something like this in default: "throw new Error('invalid action.type');"

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arrayOfState = [];
```
kukla1989, line 10: rename on "arrayOfStates"

kukla1989 (changes requested):

jstmpelowycz (approved):
awesome!

PR #2139, opened 2023-03-16
stateful clones initial

ofenchuk:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,40 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
   // write code here
+  const transformed = [];
```
jstmpelowycz, line 12: transformed what?

src/transformStateWithClones.js
```diff
@@ -6,8 +6,40 @@
...
+        newState = {};
+        transformed.push({});
+        break;
+      }
+
+      case 'removeProperties': {
+        for (const remove of object.keysToRemove) {
+          delete newState[remove];
+        }
+        transformed.push({ ...newState });
```
jstmpelowycz, line 33: DRY, push outside of your switch-case

src/transformStateWithClones.js
```diff
@@ -6,8 +6,40 @@
...
+        break;
+      }
+
+      default:
+        return object.type;
```
jstmpelowycz, line 38: would be nice to throw an error

jstmpelowycz (changes requested):

src/transformStateWithClones.js
liliia-livandovskaya: not fixed

liliia-livandovskaya (approved):

PR #2141, opened 2023-03-16
switch case solution for deep copy

TolikSerhiychuk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = JSON.parse(JSON.stringify(state));
+
+  const states = [];
```
jstmpelowycz, line 12: could have been simpler + do not add extra spaces
```suggestion
  const stateCopy = { ...state };
  const states = [];
```

jstmpelowycz (approved):

PR #2143, opened 2023-03-16
transform state with clones solution

AP-Solution:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = JSON.parse(JSON.stringify(state));
+
+  const states = [];
```
jstmpelowycz, line 12: format properly + you can use spread operator
```suggestion
  const stateCopy = { ...state };
  const states = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+      case 'removeProperties':
```
jstmpelowycz, line 19:
```suggestion
        break;

      case 'removeProperties':
```

jstmpelowycz (approved):

PR #2145, opened 2023-03-16
Solution task 3

DimaDvm:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [state];
+  let currentState = state;
+
+  for (let i = 0; i < actions.length; i++) {
+    const action = actions[i];
```
jstmpelowycz, line 14: let's use for-of

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [state];
```
jstmpelowycz, line 10: why do you need to keep the initial state?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+    const action = actions[i];
+
+    switch (action.type) {
+      case 'addProperties':
+        currentState = {
+          ...currentState,
+          ...action.extraData,
+        };
+        break;
+      case 'removeProperties':
```
jstmpelowycz, line 23: add spaces
```suggestion
        break;

      case 'removeProperties':
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+        const filteredState = {};
+
+        for (const [key, value] of Object.entries(currentState)) {
+          if (!action.keysToRemove.includes(key)) {
+            filteredState[key] = value;
+          }
+        }
+        currentState = filteredState;
```
jstmpelowycz, line 31: too complicated, let's make it simple as in the previous task

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+        currentState = {};
+        break;
+      default:
+        throw new Error(`Unsupported action type: ${action.type}`);
+    }
+
+    states.push({ ...currentState });
+  }
+
+  return states.slice(1);
```
jstmpelowycz, line 43: remove the initial one, and you won't need slicing

jstmpelowycz (changes requested):

liliia-livandovskaya (approved):

PR #2149, opened 2023-03-16
solution

DanielSorokowski:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arrayOfStates = [];
```
solaryasha, line 10:
```suggestion
  const states = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+          stateCopy[key] = object.extraData[key];
+        }
+        break;
+      case 'removeProperties':
+        object.keysToRemove.forEach(key => {
+          delete stateCopy[key];
+        });
+        break;
+
+      default:
```
solaryasha, line 28: The same comment as in the previous task - https://github.com/mate-academy/js_stateful-object/pull/2187#discussion_r1138825291

solaryasha (changes requested):

PR #2156, opened 2023-03-20
solution

maksym-kostetskyi:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,34 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateVers = [state];
+
+  for (let i = 0; i < actions.length; i++) {
```
ivbutsykin, line 13: You can use `for of` instead `for` here

src/transformStateWithClones.js
```diff
@@ -6,8 +6,34 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateVers = [state];
```
ivbutsykin, line 11: So, as I understand, at first you add `state` in array and after that you remove it from array when make return `stateVers.slice(1)`.

It's seems not good. I can suggest you make sth like that:
```suggestion
  const result = [];
  const stateCopy = { ...state };
```

When you will push your data in result and have not modified state.

src/transformStateWithClones.js
```diff
@@ -6,8 +6,34 @@
...
+
+  for (let i = 0; i < actions.length; i++) {
+    let newState = { ...stateVers[i] };
+
+    switch (actions[i].type) {
+      case 'addProperties':
+        for (const [key1, value1] of Object.entries(actions[i].extraData)) {
+          newState[key1] = value1;
+        }
```
ivbutsykin, line 20: It can be simplified: use `Object.assign`

src/transformStateWithClones.js
```diff
@@ -6,8 +6,34 @@
...
+        break;
+      case 'removeProperties':
+        for (const key2 of actions[i].keysToRemove) {
+          if (newState.hasOwnProperty(key2)) {
```
ivbutsykin, line 24: I think, if you will delete undefined value it's will still work, so you can remove this check

ivbutsykin (changes requested):

slavanosatskyi (approved):
all good, but don't forget about default case in switch case statement

PR #2159, opened 2023-03-20
Solution

yarik-bachynskyi:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
...
 function transformStateWithClones(state, actions) {
   // write code here
+  const copyState = { ...state };
+  const states = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(copyState, action.extraData);
+        break;
```
DarkMistyRoom, line 18:
```suggestion
  for (const action of actions) {
  const { type, extraData, ... } = action;
  
    switch (type) {
      case 'addProperties':
        Object.assign(copyState, extraData);
        break;
```

DarkMistyRoom: It isn't a mistake, but decomposing is good practice

DarkMistyRoom (approved):
Good job

PR #2160, opened 2023-03-21
commit

bohdanmayor:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,27 @@
...
+      array.push({ ...stateCopy });
+    } else if (actions[i]['type'] === 'removeProperties') {
+      for (const key of actions[i]['keysToRemove']) {
+        delete stateCopy[key];
+      }
+      array.push({ ...stateCopy });
+    } else if (actions[i]['type'] === 'clear') {
+      stateCopy = {};
+      array.push({ ...stateCopy });
+    }
```
etojeDenys, line 28: Let's firstly rewrite it using switch\case

src/transformStateWithClones.js
```diff
@@ -8,6 +8,27 @@
...
   // write code here
+  const array = [];
+  let stateCopy = {
+    ...state,
+  };
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i]['type'] === 'addProperties') {
+      Object.assign(stateCopy, actions[i]['extraData']);
+      array.push({ ...stateCopy });
```
etojeDenys, line 19: we can just push copy after all `if` statements

etojeDenys (changes requested):

src/transformStateWithClones.js
bohdanmayor: Its have no sence, and if/else more comfortable

OlEzhA0 (approved):

PR #2162, opened 2023-03-23
solution

Fedos17:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const OriginalObj = Object.assign({}, state);
```
YegorKorenkov, line 10: Name of object should be in camelCase, please fix

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+  const res = [];
+
+  for (const action in actions) {
+    const operation = actions[action];
+
+    switch (operation.type) {
+      case 'addProperties':
+        Object.assign(OriginalObj, operation.extraData);
+        res.push(Object.assign({}, OriginalObj));
```
YegorKorenkov, line 19: you don't need to push originalObject to the result for each case. Try to move this after switch..case construction so that it is added to the result after each iteration of the loop. Please fix for all casses

YegorKorenkov (changes requested):
Good job! Please check comments and fix

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const camelCase = Object.assign({}, state);
```
YegorKorenkov, line 10: I mean use camelCase format :) your variable name started with a capital letter
OriginalObj -> originalObj

YegorKorenkov (approved):
Nice work!

PR #2163, opened 2023-03-23
solution

popeye88:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+
+      case 'clear':
+        nextState = {};
+        break;
+
+      default:
+        return;
+    }
+
+    result.push(nextState);
```
Viktor-Kost, line 37:
```suggestion
result.push({ ...stateCopy });
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+  }
+
+  return result.slice(1);
```
Viktor-Kost, line 40:
```suggestion
  return result;
```

Viktor-Kost (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [state];
```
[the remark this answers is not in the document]
popeye88: better let stateCopy = { ...state }; due to reassigned below

PR #2165, opened 2023-03-23
solution

tkalindima:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+          delete newState[value];
+        }
+        break;
+
+      case 'clear':
+        for (const prop in newState) {
+          delete newState[prop];
+        }
+        break;
+    }
```
Volodymyr-Melnyk-Fron10der, line 32: add default case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const newState = {
+    ...state,
+  };
+
+  for (const action of actions) {
```
Volodymyr-Melnyk-Fron10der, line 15: destructurize action

Volodymyr-Melnyk-Fron10der (changes requested):
almost there👍

PR #2172, opened 2023-03-28
solution

MikiOOOO:


mateusz-michalowski (approved):

michalcywka (changes requested):
Interesting approach, but I think we can simplify it a little bit :)
Take a look at checklist point 5, we can make a copy of an object first, and then just push next versions of it to `states` without tracking the last index of our array. Then we also don't need to `slice` the array at the end.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+
+      case 'clear':
+        stateCopy = {};
+        break;
+
+      default:
+        return 'Wrong input data';
+    }
+    states.push(stateCopy);
+    stateCopy = Object.assign({}, states[states.length - 1]);
```
michalcywka, line 38: To make it in one line:
```suggestion
    states.push({ ...stateCopy })
```

michalcywka (changes requested):

d-wasilewski (approved):

PR #2177, opened 2023-03-29
solution

artem-shuhailo:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copy = { ...state };
+  const newArray = [];
```
SanyaBratashchuk, line 11:
```suggestion
  const stateCopy = { ...state };
  const stateClones = [];
```

SanyaBratashchuk (changes requested):

tetiana-ivanchenko (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
+
+      default:
+        return 'incorrect action type';
```
SanyaBratashchuk, line 34:
```suggestion
        return throw new Error(''unknown action type'');
```

SanyaBratashchuk (approved):

PR #2185, opened 2023-03-30
task solution approach 30/03/2023 20:36

mwgreybeard72:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,40 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const statefulClone = {
+    ...state,
+  };
+  const copy = [];
```
d-wasilewski, line 14: use more descriptive name

d-wasilewski (changes requested):

src/transformStateWithClones.js
```diff
@@ -8,6 +8,40 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const statefulClone = {
```
d-wasilewski, line 11: `statefulCloneCloned` are not very descriptive. Consider using names that describe what the variables represent, in this case it's hard to guess. 🤔 Maybe try sth like statefulCloneHistory

src/transformStateWithClones.js
```diff
@@ -8,6 +8,40 @@
...
+        }
+        break;
+
+      case 'removeProperties':
+        for (let y = 0; y < actions[i].keysToRemove.length; y++) {
+          delete statefulClone[actions[i].keysToRemove[y]];
+        }
+        break;
+
+      case 'clear':
```
d-wasilewski, line 30: Consider using strict equality `===` instead of loose equality `==` when comparing object length. 🤔

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2186, opened 2023-03-30
solution

DanyloPasko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
vlad-demchuk, line 14: use switch case when you compare one value

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      currentState = {
+        ...currentState, ...action.extraData,
```
vlad-demchuk, line 16: use `Object.assign`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      };
+    } else if (action.type === 'removeProperties') {
+      const keysToRemove = action.keysToRemove || [];
```
vlad-demchuk, line 19: why do we need ` || []`?

vlad-demchuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+
+      for (const [key, value] of Object.entries(currentState)) {
+        if (!keysToRemove.includes(key)) {
+          newState[key] = value;
+        }
+      }
+      currentState = newState;
+    } else if (action.type === 'clear') {
+      currentState = {};
+    }
```
IvanFesenko, line 31: It shoul be inside switch/case instead of if/else

IvanFesenko (changes requested):

VitaliyBondarenko1982 (approved):

PR #2189, opened 2023-03-31
solution

NadieinOleh:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = [];
+  const newstate = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```
witflash, line 14: `switch/case` should have `default` statement. 
You can do
```
default:
  console.warn('There is no such action type')
```
It does nothing but it's a correct syntax for `switch/case` usage.

witflash (approved):
Well done 👍 
But don't forget about `default` statement if you use `switch` block/

PR #2195, opened 2023-04-02
Solution

SuaBy514:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
...
+        }
+
+        break;
+
+      case 'clear':
+        const stateKeys = [ ...Object.keys(clonedState) ];
+
+        for (let i = 0; i < stateKeys.length; i++) {
+          delete clonedState[stateKeys[i]];
+        }
```
d-wasilewski, line 32: what about `clonedState = {}` to clear it

d-wasilewski (changes requested):
checklist point 3

src/transformStateWithClones.js
SuaBy514: I... genuinely have no idea why I did not think of that, I feel really stupid now

d-wasilewski: It was not that obvious, don't worry

d-wasilewski (approved):

PR #2197, opened 2023-04-02
solve

ctgdprzyb:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
-  // write code here
+  const objectClones = [];
+  const clone = {};
+
+  for (const key in state) {
+    clone[key] = state[key];
+  }
+
+  for (const action of actions) {
+    switch (action['type']) {
```
d-wasilewski, line 18: you can just use . to access this property here, its not a variable

d-wasilewski (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+        break;
+    }
+
+    const cloneMirror = {};
+
+    for (const key in clone) {
+      cloneMirror[key] = clone[key];
+    }
+
+    objectClones.push(cloneMirror);
```
d-wasilewski, line 40: why not just use the original copy? 
```
    objectClones.push(copy);
```

d-wasilewski (changes requested):

PR #2208, opened 2023-04-10
solution

Absolution26:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let stateCopy = { ...state };
+  const objsAfterActions = [];
+
+  for (const { type, extraData, keysToRemove } of actions) {
+    if (type === 'addProperties') {
+      stateCopy = { ...stateCopy };
```
Anastasiia-Svintsova, line 15: You don't need to create a new copy of the state here, as you're already creating a copy before entering the loop on line 5.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let stateCopy = { ...state };
+  const objsAfterActions = [];
+
+  for (const { type, extraData, keysToRemove } of actions) {
+    if (type === 'addProperties') {
```
Anastasiia-Svintsova, line 14: Use switch/case instead

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
+
+    if (type === 'removeProperties') {
+      stateCopy = { ...stateCopy };
+
+      for (const propForRemove of keysToRemove) {
+        if (stateCopy.hasOwnProperty(propForRemove)) {
+          delete stateCopy[propForRemove];
+        }
+      }
+      objsAfterActions.push(stateCopy);
```
Anastasiia-Svintsova, line 33: You do this action in each case. So just move it to the end of the function to not repeat yourself

Anastasiia-Svintsova (changes requested):

anstsot (approved):
Great job, approved!

PR #2209, opened 2023-04-10
clones

Dyakovich-R:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,39 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
```
OleziO, line 10: Remove all comments from the code.

OleziO (changes requested):

Viktor-Kost (approved):

PR #2211, opened 2023-04-10
js_stateful_clones

YelyzavetaMolochynska:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
+  const arr = [];
```
VitaliyBondarenko1982, line 11:
```suggestion
  const result = [];
```
can be more meaningful naming

VitaliyBondarenko1982 (approved):

PR #2217, opened 2023-04-11
Solution

OlyaProkofieva:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const copy = { ...state };
+  const stateArr = [];
+
+  for (const action of actions) {
+    switch (true) {
+      case action.type === 'addProperties':
```
tetiana-ivanchenko, line 16: the same as in previous task

src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
...
+          delete copy[key];
+        }
+        break;
+      case action.type === 'clear':
+        for (const key in copy) {
+          delete copy[key];
+        }
+        break;
+      default:
+        break;
```
tetiana-ivanchenko, line 30: image

tetiana-ivanchenko (changes requested):

src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const copy = { ...state };
```
``` suggestion
  let copy = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
...
+        for (const key of action.keysToRemove) {
+          delete copy[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in copy) {
+          delete copy[key];
+        }
+        break;
```
```suggestion
      case 'clear':
        copy = {};
        break;
```

etojeDenys (changes requested):

vlad-demchuk (approved):

PR #2220, opened 2023-04-12
Solution

AlexRaven93:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,30 @@
...
+      case 'removeProperties':
+        action.keysToRemove.forEach(key =>
+          delete clone[key]);
+        break;
+      case 'clear':
+        Object.keys(clone).forEach(key =>
+          delete clone[key]);
+        break;
+      default:
+        break;
```
Volodymyr-Melnyk-Fron10der, line 28: I would throw an error like 'Unknown action type'

src/transformStateWithClones.js
```diff
@@ -8,6 +8,30 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const clone = { ...state };
+  const transform = [];
```
Volodymyr-Melnyk-Fron10der, line 12:
```suggestion
  const transformedStates = [];
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,30 @@
...
+
+  for (const action of actions) {
```
Volodymyr-Melnyk-Fron10der, line 14: Destructure the action
```suggestion
  for (const { type, extraData, ... } of actions) {
```

Volodymyr-Melnyk-Fron10der (changes requested):
Just a few possible improvements

src/transformStateWithClones.js
loralevitska: not fixed. Return `default` case

src/transformStateWithClones.js
```diff
@@ -8,6 +8,29 @@
...
+        break;
+      case 'removeProperties':
+        keysToRemove.forEach(key =>
+          delete clone[key]);
+        break;
+      case 'clear':
+        clone = {};
+        break;
+      default:
+        return 'Unknown action type';
```
loralevitska, line 27:
```suggestion
      case 'addProperties':
        Object.assign(clone, extraData);
        break;
        
      case 'removeProperties':
        keysToRemove.forEach(key =>
          delete clone[key]);
        break;
        
      case 'clear':
        clone = {};
        break;
        
      default:
        return 'Unknown action type';
```

loralevitska (approved):
Good job!

PR #2223, opened 2023-04-14
Solution

VladyslavKucher:


Volodymyr-Melnyk-Fron10der (changes requested):
I suggest to create a state variable next to stateClones  and mutate it. In the end just push a cop of state in the stateClones  , use spread operator

witflash (approved):
Good job 👍🏻 
Don't forget that `switch/case` should contain a `default` statement (for handle errors for example)

PR #2227, opened 2023-04-14
done

MayerLansky:


Volodymyr-Melnyk-Fron10der (approved):
well done, but try to use destructure assignment as much as possible

PR #2228, opened 2023-04-15
solution

HannieA:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = Object.assign({}, state);
```
```suggestion
  let newState = Object.assign({}, state);
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        for (const key of keysToRemove) {
+          delete newState[key];
+        }
+        result.push(Object.assign({}, newState));
+        break;
+
+      case 'clear':
+        for (const prop of Object.getOwnPropertyNames(newState)) {
+          delete newState[prop];
+        }
```
etojeDenys, line 32:
```suggestion
        newState = {};
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+  const newState = Object.assign({}, state);
+  const result = [];
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
+        Object.assign(newState, extraData);
+        result.push(Object.assign({}, newState));
```
etojeDenys, line 19: we can write it only one time at the end of the loop

etojeDenys (changes requested):

PR #2231, opened 2023-04-18
Solution

artyomwhite:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
...
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete currentState[key];
+        }
+        break;
+      case 'clear':
+        currentState = {};
+        break;
+    }
```
etojeDenys, line 26: don't forget always add the default case

etojeDenys (approved):
good job

PR #2233, opened 2023-04-18
commit

vladimirshub:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const currentState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```

OleziO (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+    },
+    removeProperties: (action) => {
+      for (const key of action.keysToRemove) {
+        delete currentState[key];
+      }
+    },
+    clear: () => {
+      for (const key in currentState) {
+        delete currentState[key];
+      }
```
loralevitska, line 25: if it is a copy, here you can simply overwrite the value with an empty object
```suggestion
      currentState = {};
```

loralevitska (approved):

PR #2242, opened 2023-04-23
Solution

vitalii-fedusov:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,31 @@
...
+      case 'removeProperties':
+        for (const j of action.keysToRemove) {
+          delete stateCopy[j];
+        }
+        break;
+
+      default:
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
```
etojeDenys, line 29: let's move it to clear case and in default just throw the error

etojeDenys (changes requested):

Viktor-Kost (approved):

PR #2243, opened 2023-04-23
Solution

anpiet:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let newState = [];
+  const finalState = [];
+
+  newState = { ...state };
```
AndrzejPec, line 13: The initialization of newState as an empty array seems unnecesary since you're reassigning it as an object straight away, maybe just discard line 10 and initialize it like in line 13

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+
+  for (let i = 0; i < actions.length; i++) {
+    for (const key in actions[i]) {
+      if (actions[i][key] === 'addProperties') {
```
AndrzejPec, line 17: it's more convenient to use switch statement, and when you do:

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        Object.assign(newState, actions[i]['extraData']);
+      };
+
+      if (actions[i][key] === 'removeProperties') {
+        for (const key1 in actions[i]['keysToRemove']) {
```
AndrzejPec, line 22: I think it needs to be forEach to avoid nesting loops inside if statement and/or make code more readable

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        }
+      }
+
+      if (actions[i][key] === 'clear') {
+        for (const key1 in newState) {
+          delete newState[key1];
+        };
+      };
+    };
+    finalState[i] = { ...newState };
```
AndrzejPec, line 33: consider using the push method instead to avoid unintended mutations

AndrzejPec (changes requested):
just a few improvements suggested :)

AndrzejPec (approved):
still unsure about Object.assing or spread but overall seems fine :)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const actualState = { ...state };
+  const statesArr = [];
+
+  for (const key of actions) {
```
choeqq, line 13: the more proper name would be action I think

choeqq (approved):
Well done ✅

PR #2247, opened 2023-04-26
Solution

generatedman:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,25 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  const result = [];
+  const copyState = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
33Orange, line 14: Use `switch` statement instead `if else` if you have limited amount of conditions.

33Orange (approved):
Try to use `switch` statement.
Everything else is fine.
Well done!

PR #2251, opened 2023-04-27
finished the task

EvhenIvchenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+
+      case 'clear':
+        for (const key in newState) {
+          delete newState[key];
+        }
+        array.push(newState);
+        break;
+
+      default:
+        // do nothing for unknown action types
```
sTorba24, line 37: remove comments, better to not write them or write if it is something super-tricky to understand(but in more cases if it's tricky - it can be rewritten simplier :)

In your case it's just not needed

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+  const array = [];
+  let oldState = { ...state };
+
+  for (const action of actions) {
+    const newState = { ...oldState };
+
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(newState, action.extraData);
+        array.push(newState);
```
sTorba24, line 19: you can move it after switch to avoid duplicating on 26, 33 lines

sTorba24 (changes requested):
GJ 🚀 Lets improve it a little :)

sTorba24 (changes requested):
Fix comments above

sTorba24 (approved):
Cool!

PR #2257, opened 2023-05-01
add transformStateWithClones solution

DidenkoDenis306:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
+
+      default:
```
33Orange, line 31: Instead of returning null, consider throwing an error with a meaningful message, e.g. throw new Error('Invalid action type');

`[CODE STYLE]: switch/case should always have default case for error handling.`

33Orange (approved):
Everything is fine. One little mistake with `default` state of `switch` statement.
Well done!

PR #2259, opened 2023-05-02
clones

diana-tuz:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const clonedState = Object.assign({}, state);
+
+  for (const key in actions) {
+    if (actions[key].type === 'addProperties') {
+      Object.assign(clonedState, actions[key].extraData);
+      result.push(Object.assign({}, clonedState));
```
vlad-demchuk, line 16:
```suggestion
      Object.assign(stateCopy, action.extraData);
```

it will be clearer in this way

and push it just once after all if/else

vlad-demchuk (changes requested):

YegorKorenkov (approved):

PR #2261, opened 2023-05-04
Solution

Oleksii-Hrytsenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,50 @@
...
+  for (let i = 0; i < actions.length; i++) {
+    const action = actions[i];
+
+    if (i === 0) {
+      lastObject = { ...stateCopy };
+    } else {
+      lastObject = { ...resultArray[i - 1] };
+    }
+
+    switch (true) {
```
PolinaRomanenko, line 24: Use 'switch (action.type)' instead of 'switch (true)' for more readability and better performance.

PolinaRomanenko (approved):
Look at the comment

src/transformStateWithClones.js
Oleksii-Hrytsenko: fixed

PR #2266, opened 2023-05-06
transform state with clones

pos1bl:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  let newState = { ...state };
+  const resultArray = [];
+
+  for (let i = 0; i < actions.length; i++) {
```
```suggestion
  for (const action of actions) {
```

We don't need indexes here so we can use for of loop instead

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        for (const key of actions[i].keysToRemove) {
+          delete newState[key];
+        }
+        break;
+      default:
+        return 'Something went wrong';
+    }
 
+    resultArray.push(newState);
+    newState = { ...newState };
```
nazarmatsevych, line 35:
```suggestion
    resultArray.push({ ...newState });
```

Should work as well

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+          newState[key] = actions[i].extraData[key];
+        }
+        break;
+      case 'clear':
+        for (const key in newState) {
+          delete newState[key];
+        }
+        break;
+      case 'removeProperties':
+        for (const key of actions[i].keysToRemove) {
```
nazarmatsevych, line 26: Use destructuring of action object

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  let newState = { ...state };
+  const resultArray = [];
```
nazarmatsevych, line 11:
```suggestion
  const result= [];
```

Don't use variable type for naming

nazarmatsevych (approved):
Good work,
pls check my comments

PR #2268, opened 2023-05-07
Solution

GeorgeSavostikov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
OleziO, line 10: You must work with the state, without creating any variables.

OleziO (changes requested):

OleziO (changes requested):
The previous comment was not resolved. If you have any questions - feel free to ask them in fe_chat

GeorgeSavostikov:
Good day! I considered your comment, but I consider your condition incorrect because the condition for this task says that "state " cannot be changed, it must remain unchanged. Perhaps you confused the task "Stateful Clones" with "Stateful Object". I apologize in advance if I'm wrong, but I've tried many options and the tests always come up with one error, it needs to return "state" as it was set in the function, without changes.

SanyaBratashchuk (approved):

PR #2270, opened 2023-05-08
solution

skinnyjazz:


serentinos (approved):
Good job! Looks more compact than mine solution :0

src/transformStateWithClones.js
```diff
@@ -6,8 +6,37 @@
...
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
```
silchyn, line 30:
```suggestion
      case 'addProperties':
        Object.assign(stateCopy, action.extraData);
        
        break;

      case 'removeProperties':
        for (const key of action.keysToRemove) {
          delete stateCopy[key];
        }
        
        break;

      case 'clear':
        for (const key in stateCopy) {
          delete stateCopy[key];
        }
        
        break;
```

silchyn (approved):

PR #2275, opened 2023-05-08
add solution

serentinos:


33Orange (approved):
Well done! 👍 

Just one moment for self improve.

PR #2277, opened 2023-05-09
Solution

amashutikov:


niaYaro (approved):
Great solution! 
Everything is clear and understandable.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,52 @@
...
+        break;
+
+      case 'removeProperties':
+        removeProperties(copyState, action.keysToRemove);
+        result[result.length] = { ...copyState };
+
+        break;
+
+      default:
+        return 'Error';
```
dinaminahi, line 51: Instead of returning 'Error' as a string, consider throwing an error with a descriptive message, for example: `throw new Error('Invalid action type')`.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,52 @@
...
+
+  function clearObject(object) {
+    for (const key in object) {
+      delete object[key];
+    }
+  }
+
+  function addProperties(object, propertiesToAdd) {
+    Object.assign(object, propertiesToAdd);
+  }
```
dinaminahi, line 26: Maybe declare these functions outside of `transformStateWithClones` because it is too big. What do you think?

cephy77 (approved):
Good one!)

src/transformStateWithClones.js
```diff
@@ -6,8 +6,53 @@
...
+}
+
+function addProperties(object, propertiesToAdd) {
+  Object.assign(object, propertiesToAdd);
+}
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
+  const result = [];
```
silchyn, line 28: names should be descriptive

src/transformStateWithClones.js
```diff
@@ -6,8 +6,53 @@
...
+  }
+}
+
+function addProperties(object, propertiesToAdd) {
+  Object.assign(object, propertiesToAdd);
+}
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
```
silchyn, line 27:
```suggestion
  const stateCopy = { ...state };
```

silchyn (changes requested):

src/transformStateWithClones.js
```diff
@@ -6,8 +6,53 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const stateClones = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        addProperties(stateCopy, action.extraData);
+        stateClones[stateClones.length] = { ...stateCopy };
```
dinaminahi, line 34: You can use `stateClones.push({ ...stateCopy })` instead of `stateClones[stateClones.length] = { ...stateCopy }`.

And you can write it only once, right after `switch case` to avoid repeating code

dinaminahi (changes requested):

silchyn (approved):

PR #2279, opened 2023-05-09
add solution

niaYaro:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clonedState = { ...state };
+  const stateArr = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(clonedState, action.extraData);
+        stateArr.push({ ...clonedState });
```
IrynaBalandiukh, line 17: Check point 4 in the checklist - [DON'T REPEAT YOURSELF] - If you perform the same action in all switch cases - do it just once afterward.
You can push the object once after the switch block.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+        for (const key of action.keysToRemove) {
+          delete clonedState[key];
+        }
+        stateArr.push({ ...clonedState });
+        break;
+
+      case 'clear':
+        for (const key in clonedState) {
+          delete clonedState[key];
+        }
```
IrynaBalandiukh, line 30: It`s a bad idea to use a loop in a loop, I think it`s better to use a separate function

IrynaBalandiukh (changes requested):
Good job!) but let`s make some changes in the code style

src/transformStateWithClones.js
```diff
@@ -6,8 +6,44 @@
  *
  * @return {Object[]}
  */
+function removeProperties(obj, toRemoveArray) {
```
Anna-Tkachenko, line 9: Let's create clear and descriptive variable names
`object` and `properties` will be great examples here
For now it's not clear what is this array, and what it should contain
Looks like it's some array that should be removed

src/transformStateWithClones.js
```diff
@@ -6,8 +6,44 @@
...
+function clearObj(obj) {
+  for (const key in obj) {
+    delete obj[key];
+  }
+}
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clonedState = { ...state };
+  const stateArr = [];
```
Anna-Tkachenko, line 23: Don't use data type names in variable names

Anna-Tkachenko (changes requested):

src/transformStateWithClones.js
```diff
@@ -6,8 +6,44 @@
  *
  * @return {Object[]}
  */
+function removeProperties(object, properties) {
+  for (const item of properties) {
+    delete object[item];
+  }
+}
+
+function clearObj(object) {
```
Anna-Tkachenko, line 15: Don't make the function name or variable name shortened. Use full names for better readability
```suggestion
function clearObject(object) {
```

Anna-Tkachenko (approved):

PR #2281, opened 2023-05-09
Add task solution

IrynaBalandiukh:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
+      case 'removeProperties':
+        removeProperty(stateClone, action.keysToRemove);
+        break;
+
+      case 'clear':
+        clearObject(stateClone);
+        break;
+
+      default:
+        return stateClone;
```
ruslanpashkov, line 28:
```suggestion
        throw new Error(`Unknown action type: ${action.type}`);
```
> [CODE STYLE]: switch/case should always have default case for error handling.

ruslanpashkov (approved):

Shevtsooov (approved):
That is awesome 🤩

src/transformStateWithClones.js
33Orange, line 28: Instead of returning `stateClone` in the default case, it would be better to send an error if action type doesn't exist.

33Orange (changes requested):
Well done! 🔥 

Just one moment:
- `Switch/case should always have default case for error handling.`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
+  }
+
+  return listOfStateVersions;
+}
+
+function removeProperties(object, properties) {
+  for (const property of properties) {
+    delete object[property];
+  }
+}
```
PanyaPrimal, line 41: good practice

PanyaPrimal (approved):

33Orange (approved):
Well done! 🔥

PR #2283, opened 2023-05-09
Solution

julshvets:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let copyState = { ...state };
+  const versionsOfState = [];
+
+  for (const object of actions) {
```
ruslanpashkov, line 13:
```suggestion
  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+    versionsOfState.push({ ...copyState });
+  }
+
+  return versionsOfState;
+
+  function removeProperties(object, propertiesToRemove) {
+    for (const property of propertiesToRemove) {
+      delete object[property];
+    }
+  }
```
ruslanpashkov, line 44: let's get rid of the nested function and move it after our `transformStateWithClones` function

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let copyState = { ...state };
```
ruslanpashkov, line 10: it would be better to declare it with `const` to avoid accidental modifications

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+        break;
+
+      case 'removeProperties':
+        const { keysToRemove } = object;
+
+        removeProperties(copyState, keysToRemove);
+        break;
+
+      case 'clear':
+        copyState = {};
```
ruslanpashkov, line 28: if we use `const` instead of `let` - we have to delete data in our object in a different way

ruslanpashkov (changes requested):

KatyaZaremska (changes requested):
make the changes that Ruslan wrote about

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+        break;
+
+      default:
+        return 'Error';
```
silchyn, line 32: try to throw an error with meaningful text

silchyn (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+
+        removeProperties(copyState, keysToRemove);
+        break;
+
+      case 'clear':
+        copyState = {};
+        break;
+
+      default:
+        throw Error('unknown actions type');
```
33Orange, line 32: Let's add to the error line the `action.type` that went there

src/transformStateWithClones.js
33Orange, line 10: Let's change the variable declaration to `const`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+        break;
+
+      case 'removeProperties':
+        const { keysToRemove } = action;
+
+        removeProperties(copyState, keysToRemove);
+        break;
+
+      case 'clear':
+        copyState = {};
```
33Orange, line 28: Since stateCopy is now a constant, we need to come up with a new way to clear the object.

33Orange (changes requested):
Good job!

A few moments:
- Let's change the copy variable declaration to `const`
- Since stateCopy is now a constant, we need to come up with a new way to clear the object. (Solution will be close to `removeProperties` function). And don't forget to take out `clean` function, just like `removeProperties`.
- Let's add to the error line the `action.type` that went there

src/transformStateWithClones.js
julshvets: Done

src/transformStateWithClones.js
julshvets: There is a condition in the task description that if the type is clear you should create an empty state object. So I did it this way. In the previous task, I used a loop to delete all the properties.

dinaminahi: In the way you are doing it, it creates new reference for `copyState` object, and the old value is still somewhere there in the memory
So here is better to loop and delete properties, however it can be slow with large objects, but usually we don`t have those

dinaminahi (changes requested):

src/transformStateWithClones.js
julshvets: I understand that it creates a new reference for copyState. And I agree with you. But there is a condition in the description of this task, and we must adhere to it in our solution of the task.

src/transformStateWithClones.js
```suggestion
  let stateCopy = { ...state };
```

silchyn (approved):

src/transformStateWithClones.js
julshvets: Thank you!
Done.

PR #2286, opened 2023-05-09
Solution

Shevtsooov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+      case 'removeProperties':
+        removeProperties(action.keysToRemove, stateCopy);
+        break;
+
+      case 'clear':
+        clear(stateCopy);
+        break;
+
+      default:
+        throw new Error('Error');
```
ruslanpashkov, line 29: it would be better to specify which error we will get

ruslanpashkov (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+    }
+
+    stateVersions.push({ ...stateCopy });
+  }
+
+  return stateVersions;
+}
+
+function removeProperties(keysToRemove, state) {
```
silchyn, line 38: what about `addProperties`?

silchyn (changes requested):
Pay attention to comment above

src/transformStateWithClones.js
```diff
@@ -7,7 +7,50 @@
...
+  return stateVersions;
+}
+
+function addProperties(state, propertiesToAdd) {
+  for (const property in propertiesToAdd) {
+    state[property] = propertiesToAdd[property];
+  }
+}
+
+function removeProperties(keysToRemove, state) {
```
33Orange, line 44: The order of parameters in the `removeProperties` function definition should be swapped. It should be `function removeProperties(state, keysToRemove) {`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,50 @@
...
+  const stateVersions = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        addProperties(stateCopy, action.extraData);
+        break;
+
+      case 'removeProperties':
+        removeProperties(action.keysToRemove, stateCopy);
```
33Orange, line 20: The order of arguments in the `removeProperties` function should be swapped. It should be `removeProperties(stateCopy, action.keysToRemove);`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,50 @@
...
+    stateVersions.push({ ...stateCopy });
+  }
+
+  return stateVersions;
+}
+
+function addProperties(state, propertiesToAdd) {
+  for (const property in propertiesToAdd) {
+    state[property] = propertiesToAdd[property];
+  }
```
33Orange, line 41: Let's change this part of the code to `Object.assign`

33Orange (changes requested):
Good job !

A few moments:
- Let's change addProperties function to code with `Object.assign` (Checklist p1).
- Some points about readability

silchyn (approved):

PR #2290, opened 2023-05-10
Solution

MykolaLopushenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+  for (const key of actions) {
+    const { type, extraData, keysToRemove } = key;
+
+    switch (type) {
+      case 'addProperties':
+        Object.assign(copyState, extraData);
+        result.push({ ...copyState });
+        break;
+
+      case 'removeProperties':
```
OleziO, line 22: Try to move all cases to JS object, it is good practice.

PR #2291, opened 2023-05-10
Solution

andrusb21:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+
+      default:
+        throw new Error('Error');
+    }
+
+    stateArr.push({ ...stateCopy });
+  }
+
+  return stateArr;
 }
```
YeMyr, line 42: 🔥 good job

YeMyr (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+
+        break;
+
+      default:
+        throw new Error('Error');
```
33Orange, line 35: It will be better to submit a bug with a minimal description of the bug.

For example:
`'Undefined action type'`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const stateArr = [];
```
Anna-Tkachenko, line 11: Don't use data type names in variable names

Anna-Tkachenko (approved):

PR #2292, opened 2023-05-10
add task solution

hekaq:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const stateCopiesArr = [];
```
Anna-Tkachenko, line 11: Don't use data type names for variable names

Anna-Tkachenko (approved):

PR #2295, opened 2023-05-10
Solution

dimoshen:


ulositska (approved):
You've already done all the hard work and done it well 💪 💪 💪 
But I would suggest optimizing some parts of code according to the checklist 🤗

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  const arrayOfStates = [];
```
silchyn, line 10: think about more descriptive name

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+  const stateCopy = { ...state };
+
+  for (const obj of actions) {
```
silchyn, line 13: what is `obj`?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+    if (obj.type === 'clear') {
```
silchyn, line 14: correct, `switch` would be better fit there

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+      for (const stateKey in stateCopy) {
+        if (stateKey) {
+          delete stateCopy[stateKey];
+        }
+      }
```
silchyn, line 19: could be separate function

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+
+      const clearedCopy = { ...stateCopy };
+
+      arrayOfStates.push(clearedCopy);
```
silchyn, line 23: this is repeated so can be moved outside the `switch`

silchyn (changes requested):

src/transformStateWithClones.js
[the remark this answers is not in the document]
dimoshen: yeah, you're right) now i get it

src/transformStateWithClones.js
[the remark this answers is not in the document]
dimoshen: yup, already switched it to switch 👍

dimoshen: fixed

src/transformStateWithClones.js
dimoshen: rename it to 'object'

src/transformStateWithClones.js
dimoshen: made switch block instead of 'if'

src/transformStateWithClones.js
dimoshen: made function removeKey(actionObject)

src/transformStateWithClones.js
dimoshen: stateHistory ?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  const stateHistory = [];
+  let stateCopy = { ...state };
+  const removeKey = function(actionObject) {
+    for (const keyToRemove of actionObject.keysToRemove) {
+      delete stateCopy[keyToRemove];
+    }
+  };
```
dinaminahi, line 16: Let's move this function outside `transformStateWithClones`, so it will be more readable and it won't be declared each time `transformStateWithClones` is called.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+
+  for (const action of actions) {
+    const { type } = action;
```
dinaminahi, line 19: Extract here all the data you may need: extraData, keysToRemove

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
 
+      case 'removeProperties':
+        removeKey(action);
+        break;
+
+      default: throw new Error("Unexpected 'type' value");
```
dinaminahi, line 34: For more descriptive error, you can insert action.type in it

dinaminahi (changes requested):

src/transformStateWithClones.js
```diff
@@ -6,8 +6,42 @@
...
+
+      case 'addProperties':
+        Object.assign(stateCopy, extraData);
+        break;
+
+      case 'removeProperties':
+        removeKey(action, stateCopy);
+        break;
+
+      default: throw new Error('Unexpected action.type value');
```
naziamoff, line 30: is `action.type` supposed to be a string?

src/transformStateWithClones.js
```diff
@@ -6,8 +6,42 @@
...
+  return stateHistory;
 }
 
+const removeKey = function(actionObject, object) {
+  const { keysToRemove } = actionObject;
+
+  for (const keyToRemove of keysToRemove) {
+    delete object[keyToRemove];
+  }
+};
```
naziamoff, line 45: it is now unclear what exactly this function does.
- It is called `removeKey`, but it basically removes key**s**
- it accepts two args, but it is totally not clear from their name what exactly they do

It is considered to be a bad practice to add the data type to the name of the variable, as it does not give us any info. That's why it is not clear which args the function accepts

naziamoff (approved):
gj! See some naming suggestions

PR #2296, opened 2023-05-10
solution

kirilshorin:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let newState = { ...state };
+  const result = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+  for (const action of actions) {
+    const keysToRemove = action.keysToRemove;
+    const extraData = action.extraData;
+
+    if (action.type === 'addProperties') {
```
silchyn, line 17: `switch` would be better fit there

silchyn (changes requested):

src/transformStateWithClones.js
kirilshorin: @silchyn what names would be good here?

naziamoff (approved):
looks good 👍 
But see some naming suggestions

PR #2297, opened 2023-05-10
[ym] stateful-clones

YeMyr:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const cloneState = { ...state };
+  const resultArray = [];
+
+  actions.forEach(action => {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(cloneState, action.extraData);
```
bohdan-mykhailenko, line 16: It's better to use spread operator for these operations.

bohdan-mykhailenko (approved):
Good job 👍

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+
+      case 'clear':
+        for (const key in cloneState) {
+          delete cloneState[key];
+        }
+        break;
+
+      default:
+        throw new Error('something went wrong...');
```
dinaminahi, line 32: Replace the generic error message 'something went wrong...' with a more specific message, like 'Invalid action type'.

src/transformStateWithClones.js
dinaminahi: @bohdan-mykhailenko not necessarily. The spread operator and Object.assign() are very similar. The main difference is that spreading defines new properties, while Object.assign() sets them.

https://2ality.com/2016/10/rest-spread-properties.html#spread-defines-properties-objectassign-sets-them

dinaminahi (changes requested):

src/transformStateWithClones.js
YeMyr: fixed 👌

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const cloneState = { ...state };
+  const resultArray = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const cloneState = { ...state };
```
```suggestion
  const stateCopy = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+  actions.forEach(action => {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(cloneState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete cloneState[key];
+        }
```
silchyn, line 22: could be separate function

silchyn (changes requested):

src/transformStateWithClones.js
YeMyr: fixed 👌

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+
+      case 'clear':
+        for (const key in cloneState) {
+          delete cloneState[key];
+        }
+        break;
+
+      default:
+        throw new Error('invalid type of actions');
```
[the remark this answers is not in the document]
YeMyr: 👌fixed

src/transformStateWithClones.js
YeMyr: fixed 👌

src/transformStateWithClones.js
YeMyr: fixed 👌

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const result = [];
```
33Orange, line 11: Let's give this variable `result` a more descriptive name.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+        break;
+
+      case 'removeProperties':
+        removeProperties(stateCopy, action.keysToRemove);
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
```
33Orange, line 26: You can move this code to a separate function just like 'removeProperties'

33Orange (changes requested):
Well done ! 🔥 

Just a few moments with readability.

src/transformStateWithClones.js
YeMyr: fixed 👌

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+        break;
+
+      default:
+        throw new Error('invalid type of actions');
```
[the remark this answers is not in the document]
YeMyr: fixed 👌

src/transformStateWithClones.js
YeMyr: fixed 👌

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
+function removeProperties(copy, listOfKeys) {
+  for (const key of listOfKeys) {
+    delete copy[key];
+  }
+}
+
+function clear(copy) {
+  for (const key in copy) {
+    delete copy[key];
+  }
```
dinaminahi, line 46: The better name for the arguments in these functions would be something like 'object' instead of `copy`
 `copy` is clear to you because you worked with `transformStateWithClones` and passed a copy of state there, but think if someone in the future wants to reuse you function `clear`, he will face the problem of not understanding which 'copy' is it going to clear. But in overall you did a great job 👍

dinaminahi (approved):

src/transformStateWithClones.js
YeMyr: @dinaminahi Well, it looks like, naming is my weak side, I'll try to consider it in the future
Thank you for the detailed explaination 🔥

PR #2298, opened 2023-05-10
transformStateWithClones solution

PANCH1811:


ivankhrsh (approved):
good job!

src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
...
+        break;
+
+      case 'clear':
+        for (const property in stateCopy) {
+          delete stateCopy[property];
+        }
+        break;
+
+      default:
+        break;
```
33Orange, line 33: You need to add error handling to `default` state

33Orange (changes requested):
Good job.

Just one moment:
- Add error handling to `default` state in `switch` (Checklist p3)

src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
...
+
+      case 'clear':
+        for (const property in stateCopy) {
+          delete stateCopy[property];
+        }
+        break;
+
+      default:
+        throw new Error(`Error`);
```
silchyn, line 33: error text should give some info on why error was thrown

src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const property of action.keysToRemove) {
+          delete stateCopy[property];
+        }
```
silchyn, line 23: could be separate function

silchyn (changes requested):

src/transformStateWithClones.js
PANCH1811: like this?
`function addProperties(extraData, state, stateVersions) {
  for (const property in extraData) {
    state[property] = extraData[property];
    }
     stateVersions.push({ ...state });
  }`
  
  I was thinking abt separate functions, but decided that it`s too much code, so I prefer shorter solution here.

silchyn: @PANCH1811 solution should be **_readable_**, not shorter

silchyn (approved):

PR #2301, opened 2023-05-10
add solution

ulositska:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let currentState = { ...state };
+
+  const versions = [];
+
+  for (const action of actions) {
+    currentState = { ...currentState };
```
growingananas, line 15: Why do we need to do that if there is line 10 of your code already?

growingananas (approved):
Well done!

src/transformStateWithClones.js
ulositska: Yeah ://// You're right
I was creating a new copy for each action, so as not to do all the changes in one initial file. But the same can be done just by pushing a copy at the end. Fixed

src/transformStateWithClones.js
```diff
@@ -6,8 +6,78 @@
...
+        currentState = {
+          ...currentState,
+          ...action.extraData,
+        };
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete currentState[key];
+        }
```
silchyn, line 64: could be separate function

silchyn (approved):
Don't forget to remove redundant comments

PR #2302, opened 2023-05-10
Solution

SofiaLiubchenko:


vchervinko (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  let copyState = { ...state };
+  const previousStateVers = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties': {
+        copyState = Object.assign(copyState, action.extraData);
+        previousStateVers.push({ ...copyState });
```
33Orange, line 17: Move the `previousStateVers.push({ ...copyState })` to the end of the loop to avoid duplicating it in each case.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+
+      case 'removeProperties': {
+        for (const keyToRemove of action.keysToRemove) {
+          delete copyState[keyToRemove];
+        }
+        previousStateVers.push({ ...copyState });
+        break;
+      }
+
+      case `clear`: {
```
33Orange, line 29: Replace the backticks with single quotes in case 'clear'.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let copyState = { ...state };
```
33Orange, line 10: Let's change this to `const`.
We don't need to rewrite copyState everytime

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+        copyState = Object.assign(copyState, action.extraData);
+        previousStateVers.push({ ...copyState });
+        break;
+      }
+
+      case 'removeProperties': {
+        for (const keyToRemove of action.keysToRemove) {
+          delete copyState[keyToRemove];
+        }
+        previousStateVers.push({ ...copyState });
```
Anna-Tkachenko, line 25: Don't repeat yourself: you've repeated the logic for adding new copy 3 times, let's move it outside of the `switch` block and write only once

Anna-Tkachenko (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let copyState = { ...state };
+  const previousStateVers = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties': {
+        copyState = Object.assign(copyState, action.extraData);
```
33Orange, line 16: You don't need to overwrite the object every time.

[The Object.assign() static method copies all enumerable own properties from one or more source objects to a target object. It returns the modified target object.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+      case `clear`: {
+        for (const key in copyState) {
+          delete copyState[key];
+        }
+        previousStateVers.push({ ...copyState });
+        break;
+      }
+
+      default: {
+        throw new Error('Unknown action');
```
33Orange, line 38: Need to add more error description. You can add an `action.type` that falls into the default state.

33Orange (changes requested):
Good job !

A few moments:
- We don't need to overwrite the copy object every time when we add properties.
- Move the `previousStateVers.push({ ...copyState })` to the end of the loop to avoid duplicating it in each case.
- Change backticks with single quotes in case 'clear'
- You need to add more description to error. You can add an `action.type` that falls into the default state.

src/transformStateWithClones.js
Anna-Tkachenko: Check @33Orange comment, you didn't it

Anna-Tkachenko (changes requested):

src/transformStateWithClones.js
SofiaLiubchenko: > Check @33Orange comment, you didn't it

@Anna-Tkachenko But then I can`t add properties(( on lesson in the morning we used 'let'...

SofiaLiubchenko: now I finally can) interesting point for me: if I use const there + just Object.assign (without copyState = Object.assign....) for adding, everything works fine

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
+  const previousStateVers = [];
```
Anna-Tkachenko, line 11: Don't make the variable name shortened. Use full names for better readability

Anna-Tkachenko (approved):

PR #2306, opened 2023-05-10
solution

thienlaoo:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const cloneState = { ...state };
+  const resultArray = [];
```
ruslanpashkov, line 11: it could be better to call it `stateVersions`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+          if (key in cloneState) {
+            delete cloneState[key];
+          }
+        }
+        break;
+      case 'clear':
+        for (const key in cloneState) {
+          delete cloneState[key];
+        }
+        break;
```
ruslanpashkov, line 29: let's think about code readability and add spaces between our cases

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+      default:
+        return ('Invalid action type:', action.type);
```
ruslanpashkov, line 31: let's use interpolation or concatenation to make it clearer

ruslanpashkov (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateVersions = { ...state };
```
naziamoff, line 10: this one is not stateVersions, this one is just a clone of the state

naziamoff (changes requested):
please be more patient when naming your variables

naziamoff (approved):

PR #2307, opened 2023-05-10
Solution

PanyaPrimal:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,43 @@
...
+        clonedState = {};
+        break;
+      }
+
+      default: {
+        throw Error(`Unknown action: ${action.type}`);
+      }
+    }
+
+    const clonedStateCopy = { ...clonedState };
```
33Orange, line 40: You can directly push the new state into the result array instead of creating a `clonedStateCopy` variable.

src/transformStateWithClones.js
```diff
@@ -6,8 +6,43 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
33Orange, line 11: Rename this variable to something more descriptive.

src/transformStateWithClones.js
```diff
@@ -6,8 +6,43 @@
...
+  let clonedState = { ...state };
```
33Orange, line 12: Use `const` instead `let` here.
Rename this variable to something more descriptive.

src/transformStateWithClones.js
```diff
@@ -6,8 +6,43 @@
...
+      case 'addProperties': {
+        Object.assign(clonedState, action.extraData);
+
+        break;
+      }
+
+      case 'removeProperties': {
+        for (const key of action.keysToRemove) {
+          delete clonedState[key];
+        }
```
33Orange, line 25: Move this code to separate function after `transformStateWithClones` function.

src/transformStateWithClones.js
```diff
@@ -6,8 +6,43 @@
...
+
+        break;
+      }
+
+      case 'clear': {
+        clonedState = {};
```
33Orange, line 31: You need to figure out how to clear the object differently, since now it is written as a `const`. 
Move the function after `transformStateWithClones` function.

33Orange (changes requested):
Good job.

A few moments to fix

naziamoff (approved):
👍

PR #2309, opened 2023-05-10
Solution

annholovko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const statesArray = [];
```
Anna-Tkachenko, line 10: Don't use data type name for variable name

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+  let stateCopy = { ...state };
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
```
Anna-Tkachenko, line 14: ⭐️

Anna-Tkachenko (approved):

PR #2311, opened 2023-05-10
Solution

Yuliia-kucheriavenko:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,39 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const copyState = Object.assign({}, state);
+  const arrayOfversions = [];
```
Anna-Tkachenko, line 12: Don't use data type name in variable name
```suggestion
  const arrayOfVersions = [];
```

Anna-Tkachenko (approved):

PR #2313, opened 2023-05-10
add solution

growingananas:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
+        stateClone = {
+          ...stateClone,
+          ...extraData,
+        };
```
33Orange, line 21: Use `Object.assign` to add properties

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const keyToRemove of keysToRemove) {
+          delete stateClone[keyToRemove];
+        }
```
33Orange, line 27: Let's move this code to separate function outside our function `transformStateWithClones`.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        break;
+
+      case 'clear':
+        stateClone = {};
```
33Orange, line 31: You need to figure out how to clear the object differently, since now it is written as a `const`.
Move the function outside `transformStateWithClones` function.

33Orange (changes requested):
Good job !

A few moments to fix.

silchyn (approved):

PR #2314, opened 2023-05-10
Solution

maryanmats:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,27 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const result = [];
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,27 @@
...
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    let nextState = { ...currentState };
+
+    if (action.type === 'addProperties') {
```
silchyn, line 17: `switch` would be better fit here

src/transformStateWithClones.js
```diff
@@ -8,6 +8,27 @@
...
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete nextState[key];
+      }
+    } else if (action.type === 'clear') {
+      nextState = {};
+    }
+
+    result.push(nextState);
+    currentState = nextState;
```
silchyn, line 28: is it needed?

silchyn (changes requested):

src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
...
+          delete nextState[key];
+        }
+        break;
+
+      case 'clear':
+        nextState = {};
+        break;
+
+      default:
+        throw new Error('Error');
```
33Orange, line 33: Describe the error more clearly. Try to insert `action.type` there, which fell into the `default` case

src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const transformedState = [];
```
33Orange, line 11: Use more descriptive name.
`versionsOfState`
`stateVersions` for example

src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
...
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    let nextState = { ...currentState };
```
33Orange, line 15: You don't need to declare new copy of state inside your iteration.
Instead you can provide copy to .push() as argument `.push({ ...currentState})`

src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
...
+
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(nextState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete nextState[key];
+        }
```
33Orange, line 25: Move this code to separated function outside main function `transformStateWithClones`.

src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
...
+        break;
+
+      case 'clear':
+        nextState = {};
```
33Orange, line 29: You need to figure out how to `clear` the object differently, since now it is written as a `const`.
Move the function outside `transformStateWithClones` function.

src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const transformedState = [];
+  let currentState = { ...state };
```
33Orange, line 12: Let's change the variable declaration to `const`.
Use more descriptive name. 
`stateCopy` for example

33Orange (changes requested):
Good job.

A few moments to fix

dinaminahi (approved):

PR #2315, opened 2023-05-10
Solution

Mykola-Maik:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let copyState = { ...state };
+  const newArray = [];
```
silchyn, line 11: what does `newArray` tell about the value it holds?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let copyState = { ...state };
```
```suggestion
  let stateCopy = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+  const newArray = [];
+
+  for (const index of actions) {
```
silchyn, line 13: what `index` means?

silchyn (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+  let copyState = { ...state };
+  const allPreviousVersionsOfState = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        copyState = {
+          ...copyState,
+          ...action.extraData,
+        };
```
33Orange, line 19: Let's use `Object.assign` instead

src/transformStateWithClones.js
33Orange, line 10: Consider using 'const' instead of 'let' for 'copyState' since it is never re-assigned directly.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const trash of action.keysToRemove) {
+          delete copyState[trash];
+        }
```
33Orange, line 25: Move this piece of code to separate function outside `transformStateWithClones` as function `removeObjectProperties`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        break;
+
+      case 'clear':
+        for (const key1 in copyState) {
+          delete copyState[key1];
+        }
```
33Orange, line 31: Why you name key as `key1`?
Move this piece of code to separate function outside `transformStateWithClones` as function `clearObject`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        break;
+
+      default:
+        break;
```
33Orange, line 35: `[CODE STYLE]: switch/case should always have default case for error handling.` Checklist p3.
Use `throw new Error` to send errors and try to describe error, you can use `action.type` which caused the error

33Orange (changes requested):
Good job!

A few moment to fix.

33Orange (approved):
Well done ! 🔥 👍

PR #2316, opened 2023-05-10
made js stateful clones task

vladyslav-lapshyn:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+  const copyOfState = { ...state };
+  const arrayOfCopies = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(copyOfState, action.extraData);
+        break;
+
+      case 'removeProperties':
```
33Orange, line 19: Instead of iterating through all keys in copyOfState, you should iterate through the keys in action.keysToRemove

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyOfState = { ...state };
+  const arrayOfCopies = [];
```
33Orange, line 11: Let's change name to more descriptive.

`versionsOfState` for example

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+      case 'addProperties':
+        Object.assign(copyOfState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key in copyOfState) {
+          if (action.keysToRemove.includes(key)) {
+            delete copyOfState[key];
+          }
+        }
```
33Orange, line 24: Instead of iterating through all keys in copyOfState, you should iterate through the keys in action.keysToRemove.
Refactor this code without `if` condition and with `for of`
Let's move this code to separated function outside main function `transformStateWithClones`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        break;
+
+      case 'clear':
+        for (const key in copyOfState) {
+          delete copyOfState[key];
+        }
```
33Orange, line 30: Let's move this code to separated function outside main function `transformStateWithClones`

33Orange (changes requested):
Good job.

A few moments to fix.

33Orange (changes requested):
1. You changed name of `stateCopy`. But I asked to change `arrayOfCopies` to something more descriptive :). Return name 'stateCopy' to state copy and rename `arrayOfCopies`. 

Everything else is well done ! 👍

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
+  for (const key in object) {
+    delete object[key];
+  }
+}
+
+function removeObjectProperties(propertiesObject, objectToModify) {
+  for (const key of propertiesObject) {
+    delete objectToModify[key];
+  }
 }
```
naziamoff, line 47: usually it is considered to be a bad practice to use data types in your variables/function names.
As you have it quite abstract here (it looks like you can pass here any object and it will reset it) - it is ok in this context. But consider it in your next Pull Requests

naziamoff (approved):
great job 🔥

PR #2322, opened 2023-05-11
solution

vitvadym:
Had some troubles with pushing 😃

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,7 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  let newState = { ...state };
+  const newState = { ...state };
   const stateVersions = [];
 
   for (const action of actions) {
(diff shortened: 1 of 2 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let newState = { ...state };
+  const stateVersions = [];
+
+  for (const action of actions) {
+    const { type, keysToRemove, extraData } = action;
```
dinaminahi, line 14: 👍

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let newState = { ...state };
```
dinaminahi, line 10: Let't make it const and then you should change the way you are clearing it

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        newState = {
+          ...newState,
+          ...extraData,
+        };
+        break;
+
+      case 'removeProperties':
+        for (const keyToRemove of keysToRemove) {
+          delete newState[keyToRemove];
+        }
```
dinaminahi, line 31: Consider of creating separate function for this to make `switch-case` smaller and more readable

dinaminahi (changes requested):
Well done 👍 
But few more things to improve it

naziamoff (approved):
🔥

PR #2325, opened 2023-05-12
Solution

Natalia-Tsiutsiura:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      for (const key of action.keysToRemove) {
+        delete copyState[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const part in copyState) {
+        delete copyState[part];
+      }
+    }
```
VitaliyBondarenko1982, line 30: instead to `if` consider to use `switch case`

VitaliyBondarenko1982 (changes requested):

sTorba24 (approved):
Good work!

PR #2326, opened 2023-05-12
Added Solution

reseneweb:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = actions;
+  let basis = { ...state };
```
witflash, line 11: It's better to name `clonedState` for example

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+
+  for (const meaning in actions) {
```
witflash, line 13: `for (const action in actions)`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+    const point = actions[meaning];
+
+    for (const keys in point) {
```
witflash, line 16: You don't need to create an additional cycle.
You have to check `action.type` instead of this. You can use switch/case or just `if` block.
For example:
```
if (action.type === 'addProperties') {
  // use your code with `extraData`
}

if (action.type === 'removeProperties') {
  ...
}
```

witflash (changes requested):
Check my comments, please. And let's improve a bit

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = actions;
```
Anastasiia-Svintsova, line 10: It's not recommended to assign 'actions' directly to 'result' as it would mutate the original 'actions' array. Instead, initialize 'result' as an empty array.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+      result[action] = { ...clonedState };
+
+      for (let i = 0; i < point.keysToRemove.length; i++) {
+        delete result[action][point.keysToRemove[i]];
+      }
+      clonedState = result[action];
+    }
+
+    if (point.type === 'clear') {
+      result[action] = {};
```
Anastasiia-Svintsova, line 33: Same. Just push clonedState to the result after the loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+    if (point.type === 'addProperties') {
+      result[action] = {
+        ...clonedState, ...point.extraData,
+      };
+      clonedState = result[action];
+    }
+
+    if (point.type === 'removeProperties') {
+      result[action] = { ...clonedState };
```
Anastasiia-Svintsova, line 24: Same

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = actions;
+  let clonedState = { ...state };
+
+  for (const action in actions) {
+    const point = actions[action];
+
+    if (point.type === 'addProperties') {
+      result[action] = {
```
Anastasiia-Svintsova, line 17: Instead of updating the 'result' array directly, create a new object and then push it to the 'result' array.

Anastasiia-Svintsova (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+
+  for (const action in actions) {
+    const point = actions[action];
+
+    if (point.type === 'addProperties') {
+      const resultAction = {
+        ...clonedState, ...point.extraData,
+      };
+
+      result.push(resultAction);
```
agniya-i, line 21: `result.push(resultAction)` repeated each time, so you can move it from `if` blocks and push just `clonedState`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+      clonedState = resultAction;
+    }
+
+    if (point.type === 'clear') {
+      const resultAction = {};
+
+      result.push(resultAction);
+      clonedState = resultAction;
+    }
```
agniya-i, line 40:
```suggestion
    }
    result.push({...clonedState})
    
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let clonedState = { ...state };
+
+  for (const action in actions) {
+    const point = actions[action];
+
+    if (point.type === 'addProperties') {
```
agniya-i, line 16: Better to use `switch/case` here

agniya-i (changes requested):

PR #2328, opened 2023-05-13
add task solution

sasdaiv:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  const stateCopy = {
+    ...state,
+  };
+  const previousVersions = [];
+
+  for (let i = 0; i < actions.length; i++) {
```
VitaliyBondarenko1982, line 15: consider to use array methods.
in this case forEach will be good
or `for of` loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+    switch (actions[i].type) {
```
VitaliyBondarenko1982, line 16:
```suggestion
   const { type, extraData, keysToRemove } = actions[i]
    switch (actions[i].type) {
```
consider to destructure action

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+      case 'addProperties':
+        for (const key in actions[i].extraData) {
+          stateCopy[key] = actions[i].extraData[key];
+        }
+        break;
+      case 'removeProperties':
+        for (let x = 0; x < actions[i].keysToRemove.length; x++) {
```
VitaliyBondarenko1982, line 23: better to use `for of` loop or forEach method, as we need to iterate through full array

VitaliyBondarenko1982 (changes requested):

src/transformStateWithClones.js
sasdaiv: > better to use `for of` loop or forEach method, as we need to iterate through full array

why isn't there anything about the forEach method in the lessons about objects, I don't seem to have missed anything

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = {
```
```suggestion
  let stateCopy = {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
```
```suggestion
        stateCopy = {};
```

etojeDenys (approved):
lgtm

PR #2329, opened 2023-05-15
Solution

Sergej161995:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const clone = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
TarasHoliuk, line 14: the same as in the previous task - use switch (do not forget about `default` case)

TarasHoliuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const clone1 = { ...state };
```
```suggestion
  const clone = { ...state };
```

sTorba24: just `clone` :)

sTorba24 (approved):
Good work!

PR #2338, opened 2023-05-18
Solution

Svyrydov-Alex:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let newState = Object.assign({}, state);
+
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 14: You can make destructuring to get all needed props of action

polosanya (changes requested):
Almost done!

PR #2339, opened 2023-05-18
add task solution

w315:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clones = [];
+  const newState = {};
+
+  for (const key in state) {
+    newState[key] = state[key];
```
vlad-demchuk, line 14: create a copy in this way ` { ...state }`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+  }
+
+  for (let i = 0; i < actions.length; i++) {
+    const currentAction = actions[i];
+
+    switch (currentAction.type) {
+      case 'addProperties':
+        for (const givenProperty in currentAction.extraData) {
```
vlad-demchuk, line 22: no need to iterate  through the object

```

vlad-demchuk (changes requested):

volodymyr-soltys97 (approved):
Great job 👍

PR #2341, opened 2023-05-19
Clone state

dubyuliia:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+        break;
+
+      case 'clear':
+        nextCopy = {};
+        break;
+      default:
+        return {};
+    }
+
+    result.push(nextCopy);
```
nazarmatsevych, line 41:
```suggestion
    result.push({ ...stateCopy  });
```

And no need to create nextCopy

nazarmatsevych (approved):

PR #2345, opened 2023-05-21
Solution

yurii-krymskiy:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
+  const result = [];
+
+  for (const obj of actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      }
+    }
+
+    if (obj.type === 'clear') {
+      for (const key in copyState) {
+        delete copyState[key];
+      }
+    }
+    result.push({ ...copyState });
+  }
```
nazarmatsevych, line 32: Let's rewrite it using switch/case statement since we rely on single value

nazarmatsevych (changes requested):

slavanosatskyi (approved):
👍

PR #2346, opened 2023-05-21
task was sloved

Nazarht:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,28 @@
...
+  for (const num in actions) {
+    const action = actions[num];
+
+    if (action.type === 'addProperties') {
+      Object.assign(obj, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete obj[key];
+      }
+    } else if (action.type === 'clear') {
```
vlad-demchuk, line 23: will be better to use switch case

vlad-demchuk (changes requested):
GJ
just small improvement

vlad-demchuk (approved):

PR #2350, opened 2023-05-23
Solution

hulchak:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultArr = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```
OleziO, line 14: Try to move all cases to the object field.

OleziO (approved):

PR #2353, opened 2023-05-24
added solution

sheva10barca:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const addAction = 'addProperties';
+  const removeAction = 'removeProperties';
+  const clearAction = 'clear';
```
OleziO, line 12: Create an object from these values

OleziO (changes requested):

DarkMistyRoom (approved):
Great!

PR #2359, opened 2023-05-27
Solution1

RodionSav:


nazarmatsevych (changes requested):
*

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const CopyState = { ...state };
```
vlad-demchuk, line 10: do not use capital letters as the first letter in variable names
```suggestion
  const сopyState = { ...state };
```

vlad-demchuk (changes requested):

Viktor-Kost (approved):

PR #2363, opened 2023-05-31
add solution

prostonadobobb:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const newArray = [];
+
+  for (const act of actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+      for (const removeItem of act.keysToRemove) {
+        delete newState[removeItem];
+      }
+    }
+
+    if (act.type === 'clear') {
+      for (const st in newState) {
+        delete newState[st];
+      }
+    }
```
nazarmatsevych, line 30: Rewrite it using switch/case statement because we rely on a single value

nazarmatsevych (changes requested):
Good work,
pls check comment to improve your solution

volodymyr-soltys97 (approved):
Awesome 👍

PR #2372, opened 2023-06-05
solution

hrynkevych:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const transformedStates = [];
+  const stateCopy = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
+        Object.keys(stateCopy)
+          .forEach(key => delete stateCopy[key]);
```
Viktor-Kost, line 17: You can simplify
```suggestion
         stateCopy = {};
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+        transformedStates.push({ ...stateCopy });
+        break;
+
+      case 'addProperties':
+        for (const data in action.extraData) {
+          stateCopy[data] = action.extraData[data];
+        }
```
Viktor-Kost, line 24: You can also simplify 
```suggestion
Object.assign(stateCopy, action.extraData);
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
-  // write code here
+  const transformedStates = [];
+  const stateCopy = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
+        Object.keys(stateCopy)
+          .forEach(key => delete stateCopy[key]);
+        transformedStates.push({ ...stateCopy });
```
Viktor-Kost, line 18: move `transformedStates.push({ ...stateCopy });` after switch case instead of using it for each case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+        for (const key in action.keysToRemove) {
+          delete stateCopy[action.keysToRemove[key]];
+        }
+        transformedStates.push({ ...stateCopy });
+        break;
+
+      default:
+        transformedStates.push({ ...stateCopy });
+
+        return { ...stateCopy };
```
Viktor-Kost, line 38: just break or throw an error for default case
```suggestion
  break
```

Viktor-Kost (changes requested):

src/transformStateWithClones.js
hrynkevych: thank you for comments!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const transformedStates = [];
+  let stateCopy = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
```
OleziO, line 15: Try to move all cases to the object

OleziO (approved):

PR #2373, opened 2023-06-06
add task js stateful clones

Lazarend:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
-  // write code here
+  const result = [];
+  let clonedState = Object.assign({}, state);
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties': {
+        const nextState = Object.assign({}, clonedState, action.extraData);
+
+        result.push(nextState);
```
sTorba24, line 18: move it after `switch .. case` to not repeat

src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+      case 'clear': {
+        const nextState = {};
+
+        result.push(nextState);
+        break;
+      }
+      default:
+        break;
+    }
+    clonedState = Object.assign({}, result[result.length - 1]);
```
sTorba24, line 44: i think it can be removed.

sTorba24 (changes requested):
Almost good! Lets improve

src/transformStateWithClones.js
Lazarend: Hi. Thank you. Ok, I try it now.

src/transformStateWithClones.js
OleziO: Not resolved. Just move it after switch case construction to avoid code duplication. If you have any questions - feel free to ask them in `fe_chat`

OleziO (changes requested):

Lazarend:
Updated the code by moving the cloned state update logic outside the switch..case block.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let clonedState = Object.assign({}, state);
+  let nextState;
```
OleziO, line 12: Try to rewrite without `nextState`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties': {
```
OleziO, line 16: Try to move all cases to the object.

OleziO (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+        }
+      }
+
+      clonedState = nextState;
+      result.push(clonedState);
+    },
+    clear: () => {
+      clonedState = {};
+      result.push(clonedState);
+    },
```
DarkMistyRoom, line 34: no need to make a push in each handler. You can do it once in the cycle at the end of each iteration

DarkMistyRoom (approved):
Good job

PR #2374, opened 2023-06-06
add task solution

Bonum-taurus:


volodymyr-soltys97 (changes requested):
Good job
But let's make your code more perfect

nazarmatsevych (approved):
👍

PR #2375, opened 2023-06-06
Added task solution

ruslan-ihnatenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultArr = [];
```
polosanya, line 10: It would be great to give more meaningful names

polosanya (changes requested):
Good job! Let's improve your code to make it cleaner

src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        if (resultArr.length === 0) {
+          resultArr.push(Object.assign({}, state, action.extraData));
+        } else {
+          resultArr.push(Object.assign(
+            {},
+            resultArr[resultArr.length - 1],
+            action.extraData));
+        }
```
sTorba24, line 22: just save state in variable and change it, then push copy in array

src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+          temporaryResult = { ...resultArr[resultArr.length - 1] };
+        }
+
+        if (action.keysToRemove !== 0) {
+          for (const key of action.keysToRemove) {
+            delete temporaryResult[key];
+          }
+        } else {
+          temporaryResult = { ...state };
+        }
```
sTorba24, line 40: do you need this check? If no keys to remove loop won't do something

src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+
+        resultArr.push(temporaryResult);
```
sTorba24, line 42: move push outside `switch case`, you are doing the same action for every case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+        break;
+
+      case 'clear':
+        resultArr.push({});
+        break;
```
sTorba24, line 47: add `default` case

sTorba24 (changes requested):
Almost good! lets improve( you can use logic from previous task) (just changing obj)

PR #2377, opened 2023-06-08
Solution MR

RuslanMelnykFE:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+      }
+    }
+
+    if (type === 'clear') {
+      for (const key in obj) {
+        delete obj[key];
+      }
+    }
+
+    newState = obj;
```
vlad-demchuk, line 37: you can just use `Object.assign` and add a new property to `newState`
instead of creating a new object and then reassign `newState`

vlad-demchuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultActions = [];
+  let newState = { ...state };
+
+  for (const action of actions) {
+    const { type } = action;
+    const obj = { ...newState };
```
anastasiiavorobiova, line 15: Unnecessary line of code you have already made a copy of the state, so you can use it here

anastasiiavorobiova (approved):

PR #2382, opened 2023-06-08
add task stateful-clones

P-Nazar:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,45 @@
...
-  // write code here
+  const result = [];
+
+  result.push({ ...state });
+
+  for (const action of actions) {
+    const currentState = { ...result[result.length - 1] };
+
+    switch (action.type) {
+      case 'addProperties':
```
OleziO, line 18: Try to move all cases to the object.

OleziO (approved):

PR #2383, opened 2023-06-09
add task solution

ilona-borkovska:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,53 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(STATE_COPY, action.extraData);
+
+        STATE_CHANGES_LIST.push({
+          ...{},
+          ...STATE_COPY,
+        });
+
```
MaxSchmide, line 22: put it after switch/case, no need to write it in every case.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,53 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (STATE_COPY.hasOwnProperty(key)) {
+            delete STATE_COPY[key];
```
MaxSchmide, line 28: Unnecessary if

MaxSchmide (changes requested):

MaxSchmide (approved):
Good

tetiana-ivanchenko (approved):

PR #2385, opened 2023-06-09
Solution

zuzu6:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
+
+    if (item.type === 'clear') {
+      const newObjClear = result.length === 0
+        ? { ...state } : { ...result[result.length - 1] };
+
+      Object.keys(newObjClear).forEach((key) => {
+        delete newObjClear[key];
+      });
+      result.push(newObjClear);
+    }
```
nazarmatsevych, line 43: Good work but let's rewrite it using switch/case since we rely on a single value

nazarmatsevych (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,53 @@
...
+        result.push(newObj);
+        break;
+
+      case 'removeProperties':
+        const newObjRemove = result.length === 0
+          ? { ...state }
+          : { ...result[result.length - 1] };
+
+        item.keysToRemove.forEach((key) => {
+          if (newObjRemove[key]) {
```
anastasiiavorobiova, line 33: Replace the check 'if (newObjRemove[key])' with 'if (key in newObjRemove)' to handle falsy property values.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,53 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+
+  actions.forEach((item, index, array) => {
```
anastasiiavorobiova, line 12: You don't need an index and an array in your code, so there is no need to define these variables here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,53 @@
...
+    let newObj;
+
+    switch (item.type) {
+      case 'addProperties':
+        newObj = result.length === 0
+          ? { ...state }
+          : { ...result[result.length - 1] };
```
anastasiiavorobiova, line 19: You are repeating this block of code in each case, consider moving this check upper in your scope

anastasiiavorobiova (changes requested):
Great solution!

nazarmatsevych (approved):

PR #2386, opened 2023-06-10
Solution

MariiaMonakhova:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateClone, action.extraData);
+        break;
+
+      case 'removeProperties':
+        const { keysToRemove } = action;
+
+        for (const key of keysToRemove) {
+          if (stateClone.hasOwnProperty(key)) {
```
olena-ole, line 23: This check is not obligatory here, cause If the property which you are trying to delete does not exist, delete will not have any effect and will return true. [(from MDN documentation)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClone = { ...state };
+  const combinedStates = [];
+
+  for (const action of actions) {
```
olena-ole, line 13: You can destructure action right here:
    _**for (const { type, extraData, keysToRemove } of actions)**_ 
    Doing so, you won't have to reach the properties later using dot notation (line 14, 16), and use destructuring in line 20.

olena-ole (approved):
Looks good to me, I added just a few suggestions for you to consider. Approve!

src/transformStateWithClones.js
tetiana-ivanchenko, line 23: agree
https://github.com/mate-academy/js_stateful-clones/pull/2386/files#r1225766180

tetiana-ivanchenko (changes requested):

PR #2387, opened 2023-06-10
create clone

LeoSpartaDeV:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  const resArr = [];
+  let copyState = { ...state };
+
+  for (const instr of actions) {
```
TarasHoliuk, line 13:
```suggestion
  for (const action of actions) {
```
better to name it in this way, it's clearer

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        for (const key of Object.values(keysToRemove)) {
+          if (key in copyState) {
+            delete copyState[key];
+          }
+        }
+        resArr.push({ ...copyState });
+        break;
+      case 'clear':
+        copyState = {};
+        resArr.push({ ...copyState });
```
TarasHoliuk, line 33: instead of writing it in each case write it once after `switch` block

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
```
TarasHoliuk, line 34: add `default` case (might be even empty - with `break` only) - that's a good practice

TarasHoliuk (changes requested):

src/transformStateWithClones.js
LeoSpartaDeV: hi, Even without copyState = { }
Just 
default:
break;
?

LeoSpartaDeV: @TarasHoliuk 
I understood simply 
```
default:
break;
```
, but what is the use of this ?, just wondering

VitaliyBodnar: @LeoSpartaDeV  )
Screenshot 2023-06-11 at 15 35 37

VitaliyBodnar (approved):
Well done!

src/transformStateWithClones.js
LeoSpartaDeV: Understood, thanks for example and material 👍🙌

PR #2389, opened 2023-06-10
Solution

olena-ole:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,49 @@
...
+        addProperties(stateClone, extraData);
+        break;
+
+      case 'removeProperties':
+        removeProperties(stateClone, keysToRemove);
+        break;
+
+      case 'clear':
+        clear(stateClone);
+        break;
```
NataliGru, line 25: You can use the `removeProperties` function to clear all object properties by passing an array of object keys as an argument✨

src/transformStateWithClones.js
```diff
@@ -7,7 +7,49 @@
...
+
+      default:
+        return;
```
NataliGru, line 28: checklist.md:
3.[CODE STYLE]: switch/case should always have default case for error handling.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,49 @@
...
+  }
+
+  return clones;
+}
+
+function addProperties(obj, properties) {
+  for (const property in properties) {
+    obj[property] = properties[property];
+  }
+}
```
NataliGru, line 41: It will be better if you use `Object.assign(target, source)` instead of using a` for...in` loop ✨

src/transformStateWithClones.js
```diff
@@ -7,7 +7,49 @@
...
+  for (const key of keys) {
+    delete obj[key];
+  }
+}
+
+function clear(obj) {
+  for (const key in obj) {
+    delete obj[key];
+  }
 }
```
NataliGru, line 53: If you use removeProperties to clear an object, you don't need this function. It duplicates the action of the previous one✨

NataliGru (changes requested):
Great job! ✨✨✨
Let’s add a few fixes to improve your code)

NataliGru (approved):
Nice✨✨✨

tetiana-ivanchenko (approved):

PR #2390, opened 2023-06-11
solution

cebanoleksandr:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,46 @@
...
+  }
+
+  function reducer(action) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          initialState[key] = action.extraData[key];
+        }
+
+        statesArray.push({ ...initialState });
```
dianadiv, line 24: checklist.md:
4. [DONT REPEAT YOURSELF] - If you perform same action in all switch cases - do it just once afterwards

src/transformStateWithClones.js
```diff
@@ -7,7 +7,46 @@
...
+  const statesArray = [];
+
+  for (const action of actions) {
+    reducer(action);
+  }
+
+  function reducer(action) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
```
dianadiv, line 20: It would be better to use Object.assign instead of using a for...in loop

dianadiv (changes requested):
Good job, just fix few things

dianadiv (approved):
Approved

tetiana-ivanchenko (approved):

PR #2396, opened 2023-06-11
added task solution

tetianakostiuk1:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+      case 'addProperties':
+        Object.assign(stateCopy, extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          if (stateCopy.hasOwnProperty(key)) {
+            delete stateCopy[key];
+          }
+        }
```
artkas03, line 26: I would rather put this in separate function. This will make your code more clear)

artkas03 (approved):
good solution)

src/transformStateWithClones.js
artkas03, line 26: i would rather put it in separate function to make your code more clean)

tetiana-ivanchenko (approved):

PR #2406, opened 2023-06-12
Solution

eleonora-demikh:


klmyana (approved):
God job and clear code as for me, approving

PR #2407, opened 2023-06-12
add task stateful clones

AnastasiiaKhemych:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          stateClone[key] = action.extraData[key];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const item of action.keysToRemove) {
```
outofs, line 22: Good work, but you should use proper variable names. For example for `const item` you can use more informative name such as `const key` or something else.

outofs, line 22: You should use proper variable names. For example for `const item` you can use more informative name such as `const key` or something else.

outofs (changes requested):
Good work!
Please, make little fix for variable name.

outofs (approved):
Well done! Approving!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+      case 'clear':
+        for (const key in stateClone) {
+          delete stateClone[key];
+        }
+        break;
+      default:
+        throw new Error('Incorrect action type');
+    }
+    resultArray.push(Object.assign({}, stateClone));
```
anstsot, line 35: there is no need to create the new objects here with `assign` or you can use destructuring
```suggestion
    resultArray.push({ ...stateClone });
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
-  // write code here
+  const stateClone = Object.assign({}, state);
+  const resultArray = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          stateClone[key] = action.extraData[key];
+        }
```
anstsot, line 18: `Object.assign` can be use here as well

anstsot (approved):
Good work, approving:sparkles:

PR #2408, opened 2023-06-12
add task solution

AlyonaSarapina:


IgorOksentyuk (approved):
Good job! Interesting decision with using desctructurisation)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+          delete stateCopy[key];
+        });
+        break;
+
+      case clearAction:
+        stateCopy = {};
+        break;
+
+      default:
+        break;
```
tetiana-ivanchenko, line 45: default case should handling error

tetiana-ivanchenko (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const cloneState = [];
+
+  let stateCopy = {
+    ...state,
+  };
```
anstsot, line 14: you can write in one line
```suggestion
  let stateCopy = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+
+  const addAction = 'addProperties';
+  const removeAction = 'removeProperties';
+  const clearAction = 'clear';
```
anstsot, line 18: it's better to use uppercase naming once you're setting the constants
```suggestion
  const ADD_ACTION = 'addProperties';
  //...
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+      case addAction:
+        stateCopy = {
+          ...stateCopy, ...extraData,
+        };
+        break;
+
+      case removeAction:
+        stateCopy = {
+          ...stateCopy,
+        };
```
anstsot, line 33: you already have the same destructive construction an the top (line 12) so there is no need to duplicate

anstsot (changes requested):
Almost done :+1:
Let's improve a little bit

src/transformStateWithClones.js
AlyonaSarapina: But construction at line 12 creates a copy of the state variable and on line 29 I copy the changed object to push in Clones state array. It's just not gonna work without this statement.

tetiana-ivanchenko (approved):

PR #2416, opened 2023-06-13
transformed State With Clones

KatiaGamal:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+          if (copy.hasOwnProperty(key)) {
+            delete copy[key];
+          }
+        }
+        break;
+
+      case 'clear':
+        for (const key in copy) {
+          delete copy[key];
+        }
```
loralevitska, line 34: It just copy, so we don't need to use loop
```suggestion
        copy = {};
```

loralevitska (approved):
Well done)

PR #2418, opened 2023-06-13
StateWithClones solution

ivanbilotserkivskiy:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClone = { ...state };
+  const stateChanges = [];
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
```
vladmakohon, line 14: In case you don't require the `action` object, you can directly extract the required variables within the `for` condition.
```suggestion
  for (const { type, extraData, keysToRemove } of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+
+    switch (type) {
+      case 'addProperties':
+        Object.assign(stateClone, extraData);
+        break;
+      case 'removeProperties':
```
vladmakohon, line 20: Would be better to add indentations between `case` blocks.

Valentyna08 (approved):
approve!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+        keysToRemove.forEach(key => delete stateClone[key]);
+        break;
+      case 'clear':
+        for (const key in stateClone) {
+          delete stateClone[key];
+        }
```
vladmakohon, line 26: It would make more sense to also use `forEach` here. To start, get an array of keys using the `Object.keys` method.

vladmakohon (approved):
Great job on finding a solution! I do have a few suggestions that may be helpful to you.

And678 (approved):
Good work, approve :heavy_check_mark: , but check out @vladmakohon comments

PR #2420, opened 2023-06-13
Develop

NazarBaraban:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
AnatoliiLykhohodenko, line 10: I would prefer more specific name than result)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+        break;
+
+      case 'removeProperties':
+        const updatedState = { ...currentState };
+
+        for (const key of keysToRemove) {
+          delete updatedState[key];
+        }
+
+        currentState = updatedState;
```
tetiana-ivanchenko, line 30: redundant copy of currentState

tetiana-ivanchenko (approved):

PR #2421, opened 2023-06-13
Solution

NadiaKovalchuk:


FlameInBrains (changes requested):
Great start! A few changes are necessary:
1. Since after the "break" command program stops "switch" operations at the current iteration, it is better to write "resultArray.push" after "switch". It will make your code shorter.
2. It is just a recommendation, but IMAO it is not necessary to create an additional object in every switch case since you can just write "resultArray.push(cloneOfState) which already has changes in it. You just add additional lines to your code by creating a new object

FlameInBrains (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
+        break;
+
+      case 'clear':
+        for (const objKey in cloneOfState) {
+          delete cloneOfState[objKey];
+        }
+        break;
+
+      default:
+        return 'Invalid action type';
```
tetiana-ivanchenko, line 40:
```suggestion
        throw new Error('Invalid action type');
```

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #2429, opened 2023-06-13
solution

FlameInBrains:


vlopatko (approved):
Good job, this code works.
P.S. I have one advice for you, maybe in the future for clarity and better understanding of the code it would be nice to decompose object and using only variables instead "object.key" construction, so as not to get confused, but this is at your discretion. Well done 👍

tetiana-ivanchenko (approved):

PR #2432, opened 2023-06-13
add solution

pavlokuraksin:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+  const newState = { ...state };
+  const arrayOfStates = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(newState, action.extraData);
+
+        break;
+      case 'removeProperties':
```
tetiana-ivanchenko, line 19:
```suggestion
        Object.assign(newState, action.extraData);
        break;
        
      case 'removeProperties':
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        }
+
+        break;
+      case 'clear':
+        Object.keys(newState)
+          .forEach(objectValue => delete newState[objectValue]);
+
+        break;
+      default:
+        break;
```
tetiana-ivanchenko, line 31: default case should handling error

tetiana-ivanchenko (changes requested):

src/transformStateWithClones.js
CasioPayYeah: Is it bad form to put a space before a break? I think it makes the code more readable.

And678: Usually we have spaces between cases, break is a part of the previous case, looks a bit weird to have it grouped with the next case.

And678: In Tetiana's suggestion, the space is moved to be below `break`.

And678 (approved):
Good work, approve :heavy_check_mark: 
Check out the comment thread above, about spaces in switch statements.

PR #2434, opened 2023-06-13
solution

ovynarchuk:


tetiana-ivanchenko (changes requested):
make all fixes from the prev task

tetiana-ivanchenko (approved):

PR #2439, opened 2023-06-13
Implement transformStateWithClones function

taras-bilas:


tetiana-ivanchenko (changes requested):
rewrite your solution according to checklist

And678 (changes requested):
Some things can be simplified, look at the comments below

src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        stateCopy.push(
+          Object.assign(
+            {},
+            stateCopy[stateCopy.length - 1],
+            action['extraData']
+          )
+        );
```
anstsot, line 21: better to make changes to separate object and then push to `stateCopy` array

anstsot (approved):

PR #2442, opened 2023-06-14
Solution

fawnique:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+      case 'clear':
+        newState = {};
+        break;
+
+      default:
+        throw new Error('Invalid action type');
+    }
+
+    arrayOfActions.push(newState);
+    currentState = newState;
```
And678, line 37: You can simplify your code a bit by avoiding `newState` alltogether, and just mutating `currentState` instead of it. After this, you can push the **copy** of current state to your array
```suggestion
    arrayOfActions.push({...currentState});
```

And678 (changes requested):
Good work, but your code can be simplified a bit here.

tetiana-ivanchenko (approved):

PR #2446, opened 2023-06-15
Solution AB

AytajBabayeva:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    let nextState = { ...currentState };
```
polosanya, line 14: There is actually no need to make a copy one more time

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+
+    switch (action.type) {
```
polosanya, line 16: You can destructure all the needed props of action (type, extraData, keysToRemove)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+      case 'addProperties':
+        Object.assign(nextState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        if (action.keysToRemove) {
+          for (const key of action.keysToRemove) {
+            if (nextState.hasOwnProperty(key)) {
+              delete nextState[key];
```
polosanya, line 25: It would be great to use forEach instead of the traditional loop. In that case, you could avoid extra checks

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+        }
+        break;
+
+      case 'clear':
+        nextState = {};
+        break;
+
+      default:
+        // Handle unknown action types, if needed
+        break;
```
polosanya, line 37: 👍

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
polosanya, line 10: It is better to avoid naming variables 'result'. It would be great to specify what exactly the result is

polosanya (changes requested):
Almost perfect! Let's improve the code a bit)

src/transformStateWithClones.js
AytajBabayeva: do you mean I shouldnt write this part:
    let nextState = { ...currentState }; 
?

If yes, how can I rewrite it?

src/transformStateWithClones.js
AytajBabayeva: like this:

for (const { type, extraData, keysToRemove } of actions) {}

src/transformStateWithClones.js
AytajBabayeva: const clone = [];

PR #2447, opened 2023-06-15
Solution

valet2102:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,51 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const stateClone = { ...state };
+  const object = [];
+
+  for (let i = 0; i < actions.length; i++) {
```
Andriy-Shymkiv, line 14: use `for of` loop here, you really don't need indexes

src/transformStateWithClones.js
```diff
@@ -8,6 +8,51 @@
...
+        object.push(addProperties(stateClone, actions[i].extraData));
+
+        break;
+      case 'removeProperties':
+        object.push(removeProperties(stateClone, actions[i].keysToRemove));
+        // actions[i].keysToRemove.forEach(key => delete stateClone[key]);
+        break;
+      case 'clear':
+        // Object.keys(state).forEach(key => delete state[key]);
+        object.push(clearProperties(stateClone));
```
Andriy-Shymkiv, line 26: it's confusing because you pushing to array result of `clearProperties` , so better rename or remove this function

Andriy-Shymkiv (changes requested):
I don't think that you need a separate function for all this cases, so consider just to remove them and do everything inside `switch/case` block

src/transformStateWithClones.js
```diff
@@ -6,8 +6,34 @@
...
+  let stateClone = { ...state };
+  const stateCopy = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        stateClone = {
+          ...stateClone, ...action.extraData,
+        };
+
```
KlymSudakov, line 21: No need for an empty line

KlymSudakov, line 21: No need for an empty lines before break
You can remove it next to break to make cases visually separated

src/transformStateWithClones.js
```diff
@@ -6,8 +6,34 @@
...
+      case 'removeProperties':
+        for (const index of action.keysToRemove) {
+          delete stateClone[index];
+        }
+        break;
+      case 'clear':
+        stateClone = {};
+
+        break;
+    }
```
KlymSudakov, line 32: [CODE STYLE]: switch/case should always have default case for error handling.

KlymSudakov (changes requested):

PR #2449, opened 2023-06-16
transformStateWithClones

zipic:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clearObject = [];
```
loralevitska, line 10: It's not object it's array. Improve variable naming, `result` for example
```suggestion
  const clearObject = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          delete modifiedObject[key];
+        }
+        break;
+      case 'clear':
+        for (const key in modifiedObject) {
+          delete modifiedObject[key];
+        }
```
loralevitska, line 29: For copy you can use just empty object
```suggestion
        modifiedObject = {};
```

loralevitska (changes requested):
Almost perfect) Let's improve

sTorba24 (approved):
GJ 🚀

PR #2452, opened 2023-06-16
Solution js_stateful-clones

Svitlanchick:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+          stateCopy[key] = action.extraData[key];
+        }
+
+        break;
+
+      case 'removeProperties':
+        const array = action.keysToRemove;
 
+        for (const item of array) {
+          if (stateCopy[item] !== undefined) {
```
sTorba24, line 26: not needed to check it, if no key - it won't be deleted

sTorba24 (approved):
Well done!

PR #2455, opened 2023-06-17
Solution

IrynaMP:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arrayWithClones = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
```
OleziO, line 17: Move to the object. Example of usage: `case Types.AddProperties:`

OleziO (changes requested):

src/transformStateWithClones.js
IrynaMP: Sorry, cannot get it. Is there a link with a description of this method? As I know how the classic 'switch case' works only. Thank you

PR #2457, opened 2023-06-19
Develop

Luk2asz:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const result = [];
```
DorotaLeniecDev, line 11: You could use more descriptive name for readability
```suggestion
  const stateHistory= [];
```

DorotaLeniecDev (approved):
Well done :)

PR #2476, opened 2023-06-27
transform state with clones solution

vetal-hovenko:


anastasiiavorobiova (changes requested):
Great solution! Added a tiny comment to fix

volodymyr-soltys97 (approved):
Well done 👍

PR #2477, opened 2023-06-27
Solution

Anmonnn:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let state2 = { ...state };
```
volodymyr-soltys97, line 11: Instead of creating a shallow copy of the state object at the beginning of the function, it's better to do this inside the loop just before modifying state. This way, you're sure to work with a fresh clone of the state on each iteration.

volodymyr-soltys97 (approved):
Nice work 👍

PR #2478, opened 2023-06-28
add solution

denis-kononenko:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
```
SerhiiKirik, line 10: delete all comments 
```suggestion
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
...
+        };
+        break;
+
+      case 'clear':
+        for (const key in inputState) {
+          delete inputState[key];
+        };
+        break;
+    }
+
```
SerhiiKirik, line 32: always add to default case in `switch`

SerhiiKirik (changes requested):
Great 🔥 
Let's add default case 🧑‍💻

volodymyr-soltys97 (approved):

PR #2480, opened 2023-06-28
solution

SashaKozoriz:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyOfState = { ...state };
+  const resultArray = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case "addProperties":
+        const { extraData } = action;
```
Viktor-Kost, line 16: you can use object destructing at uppercase instead of doing this twice
```suggestion
  for (const action of actions) {
    const { extraData, keysToRemove } = action;
    switch (action.type) {
      case "addProperties":
```

Viktor-Kost (approved):

PR #2482, opened 2023-06-29
added a completed task

olya-girnyak:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+  const result = [];
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
+        stateCopy = {
+          ...stateCopy, ...extraData,
+        };
```
```suggestion
        stateCopy = {
          ...stateCopy,
          ...extraData,
        };
```
By the way, it is better to use Object.assign to add extra data.
In that case, stateCopy can be a const

polosanya (changes requested):
Almost perfect!

PR #2483, opened 2023-06-30
add task solution

NikitaOrliak:


mykhalenych (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+  const newState = Object.assign({}, { ...state });
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    if (type === 'addProperties') {
+      for (const key in extraData) {
+        newState[key] = extraData[key];
+      }
+    } else
```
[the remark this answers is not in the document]
NikitaOrliak: I put 'else if' because don't need to check every 'if' in the loop. Or do I need to put them in the same row?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+      }
+    } else if (type === 'removeProperties') {
+      for (const key of keysToRemove) {
+        delete newState[key];
+      }
+    } else if (type === 'clear') {
+      for (const element in newState) {
+        delete newState[element];
+      }
+    }
```
sTorba24, line 28: use switch case instead of

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arrayOfStates = [];
+  const newState = Object.assign({}, { ...state });
```
```suggestion
  const newState ={ ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+      }
+    } else if (type === 'clear') {
+      for (const element in newState) {
+        delete newState[element];
+      }
+    }
+
+    const tempState = Object.assign({}, { ...newState });
+
+    arrayOfStates.push(tempState);
```
sTorba24, line 32:
```suggestion
    arrayOfStates.push({ ...newState });
```

sTorba24 (approved):
Much better :) Approved!

PR #2484, opened 2023-07-02
solution

monikaprochniewicz:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const result = [];
+
+  actions.forEach((action) => {
+    let currentState = { ...(result[result.length - 1] || state) };
```
Radoslaw-Czerniawski, line 14: You may declare this `currentState` before your `forEach` instead of creating a new one in each iteration. It will either way hold the last result of the modifications that are done in the forEach.

```js
let currentState = { ..state };
// or instead do let currentState = JSON.parse(JSON.stringify(state)) to avoid potential shallow copy

actions.forEach((action) => {
```
- small improvement of performance (still an improvement) as you are omitting redeclaration of this variable in each iteration
- this comment  with JSON... shows another way to copy objects, basically you are making a string out of the object and then again parsing it into an object, this will get rid of any references that could be nested in your object creating a true copy -> more on that [here](https://www.freecodecamp.org/news/copying-stuff-in-javascript-how-to-differentiate-between-deep-and-shallow-copies-b6d8c1ef09cd/)

Radoslaw-Czerniawski (approved):
Approved, added a comment with potential improvement and some insight into copying objects

PR #2486, opened 2023-07-02
Solution

Hladkyi03:
Solution

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,42 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClones = [
+    {
+      ...state,
+    },
+  ];
+
+  for (let i = 0; i < actions.length; i++) {
+    if (i !== 0) {
+      stateClones[i] = Object.assign({}, stateClones[i - 1]);
+    }
+
+    switch (actions[i].type) {
+      case 'addProperties':
+        for (const addKey in actions[i].extraData) {
+          stateClones[i][addKey] = actions[i].extraData[addKey];
+        }
+
+        break;
+      case 'removeProperties':
+        for (const removeKey of actions[i].keysToRemove) {
+          delete stateClones[i][removeKey];
+        }
+
+        break;
+
+      case 'clear':
+        stateClones[i] = {};
+
+        break;
+
+      default:
+        return; // for unexpected error handling.
+    }
+  }
+
+  return stateClones;
 }
 
 module.exports = transformStateWithClones;


src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClones = [
+    {
+      ...state,
+    },
+  ];
```
SerhiiKirik, line 14:
```suggestion
  const stateClones = [{ ...state }];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+
+  for (let i = 0; i < actions.length; i++) {
+    if (i !== 0) {
+      stateClones[i] = Object.assign({}, stateClones[i - 1]);
+    }
```
SerhiiKirik, line 19: Just move it after the `switch`
```suggestion
      stateClones[i] = Object.assign({}, stateClones[i]);
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+
```
SerhiiKirik, line 20: you can make a destructuring from `stateClones[i]`
Also, you don't need index here, so, do it with `for of` loop

SerhiiKirik (changes requested):
Great 🔥 

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+        break;
+
+      case 'clear':
+        stateClones[i] = {};
+
+        break;
+
+      default:
+        return; // for unexpected error handling.
```
loralevitska, line 41: Throw error here and remove comment

src/transformStateWithClones.js
loralevitska, line 14: Can be simplify

```suggestion
  const stateClones = [{ ...state }];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+      ...state,
+    },
+  ];
+
+  for (let i = 0; i < actions.length; i++) {
+    if (i !== 0) {
+      stateClones[i] = Object.assign({}, stateClones[i - 1]);
+    }
+
+    switch (actions[i].type) {
```
loralevitska, line 21: Destructure `action[i]`

loralevitska (changes requested):
Almost done) Let's improve

loralevitska (changes requested):
it turned out that we checked the homework at the same time, so arrange all the comments and solve them one by one and simply ignore the duplicates

Hladkyi03:
I can`t replace first if in loop to your code it will make function works incorrect
image
also i can`t do it with for of loop, i need index in that if, it's main idea of code

sTorba24 (changes requested):
Good work! Lets improve it

Hladkyi03:
Completed without indexes, destructing now done before switch

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+      case 'removeProperties':
+        for (const removeKey of keysToRemove) {
+          delete stateCopy[removeKey];
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
```
SanyaBratashchuk, line 33:
```suggestion
        stateCopy = {}
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+    switch (type) {
+      case 'addProperties':
+        stateCopy = {
+          ...stateCopy,
+          ...extraData,
+        };
+        break;
+
+      case 'removeProperties':
+        for (const removeKey of keysToRemove) {
```
SanyaBratashchuk, line 25:
```suggestion
        for (const keyToRemove of keysToRemove) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
+
+      default:
+        return 'error';
```
SanyaBratashchuk, line 37:
```suggestion
        return throw new Error('unknown action type');
```

SanyaBratashchuk (approved):

PR #2487, opened 2023-07-03
transform state with clones

Orchibald:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+
+  for (const action of actions) {
+    const { type } = action;
+    let nextState;
+
+    switch (type) {
+      case 'addProperties':
+        const { extraData } = action;
+
+        nextState = {
```
mykhalenych, line 21: Avoid using an unnecessary comma at the end of the last property. Although it is not a syntax error, it's better to keep the code clean and consistent.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+          ...currentState, ...extraData,
+        };
+        break;
+
+      case 'removeProperties':
+        const { keysToRemove } = action;
```
mykhalenych, line 27: Same issue as above. The variable 'keysToRemove' is declared inside a switch case without block braces.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+  const result = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    const { type } = action;
+    let nextState;
+
+    switch (type) {
+      case 'addProperties':
+        const { extraData } = action;
```
mykhalenych, line 19: The variable 'extraData' is declared inside a switch case without block braces, which might lead to unexpected behavior as the variable could be accessible in other cases too. It's better to always use block braces in switch cases.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+        nextState = { ...currentState };
+        keysToRemove.forEach((key) => delete nextState[key]);
+        break;
+
+      case 'clear':
+        nextState = {};
+        break;
+    }
+
+    result.push(nextState);
```
mykhalenych, line 38: Consider renaming 'nextState' to a more descriptive name, e.g., 'updatedState'. This will make the code easier to understand for other developers.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+  for (const action of actions) {
+    const { type } = action;
+    let nextState;
+
+    switch (type) {
+      case 'addProperties':
+        const { extraData } = action;
+
+        nextState = {
+          ...currentState, ...extraData,
```
mykhalenych, line 22: Next state is being created without considering the default case in the switch statement. A default case should be added to handle unexpected action types.

mykhalenych (changes requested):

IvanFesenko (approved):

PR #2494, opened 2023-07-05
Add stateWithClones task solution

Serhii-Dr:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+        stateClonesArray.push({ ...stateCopyObject });
+        break;
+      case 'clear':
+        Object.keys(stateCopyObject).forEach(
+          (key) => delete stateCopyObject[key]
+        );
+        stateClonesArray.push({ ...stateCopyObject });
+        break;
+      default:
+        return 'Error: invalid value received';
```
Viktor-Kost, line 30: The function should not return a string error message in case of invalid value received. It's better to throw an error or simply ignore the invalid action.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClonesArray = [];
+  const stateCopyObject = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopyObject, action.extraData);
+        stateClonesArray.push({ ...stateCopyObject });
```
Viktor-Kost, line 17: The state clone is pushed into the array after each operation. This could be optimized by pushing the state clone into the array only once, after all operations in the switch statement. This can be achieved by moving this line after the switch statement.

Viktor-Kost (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClonesArray = [];
+  const stateCopyObject = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```
const { type, extraData, keysToRemove } = action

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+        Object.assign(stateCopyObject, action.extraData);
+        break;
+      case 'removeProperties':
+        action.keysToRemove.forEach((key) => delete stateCopyObject[key]);
+        break;
+      case 'clear':
+        Object.keys(stateCopyObject).forEach(
+          (key) => delete stateCopyObject[key]
+        );
+        break;
```
loralevitska, line 25:
```suggestion
      case 'addProperties':
        Object.assign(stateCopyObject, action.extraData);
        break;
        
      case 'removeProperties':
        action.keysToRemove.forEach((key) => delete stateCopyObject[key]);
        break;
        
      case 'clear':
        stateCopyObject = {};
        break;
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+    }
```
loralevitska, line 26: Return default case but instead of string [throw error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw?retiredLocale=uk) according previous review comment

loralevitska (changes requested):
Almost perfect) Let's improve

Viktor-Kost (approved):

PR #2495, opened 2023-07-06
js_stateful-clones

KarolinaWiktoria:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        for (const StateKey in StateCopy) {
+          if (StateCopy.hasOwnProperty(StateKey)) {
+            delete StateCopy[StateKey];
+          };
+        }
+        EmptyArray.push(Object.assign({}, StateCopy));
+        break;
+
+      default:
+        return 'Whoops, something bad happened';
```
darokrk, line 37: In case of an invalid action type, you shouldn't return a string. It's better to throw an error or ignore the invalid action.

`throw new Error('Whoops, something bad happened')`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const StateCopy = { ...state };
+  const EmptyArray = [];
```
darokrk, line 11: Variable names should follow camelCase naming convention. Consider renaming 'EmptyArray' to 'emptyArray'.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete StateCopy[key];
+        }
+        EmptyArray.push(Object.assign({}, StateCopy));
+        break;
+
+      case 'clear':
+        for (const StateKey in StateCopy) {
+          if (StateCopy.hasOwnProperty(StateKey)) {
```
darokrk, line 29: The hasOwnProperty method is not necessary in this context. You can remove it to make your code cleaner.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const StateCopy = { ...state };
```
darokrk, line 10: Variable names should follow camelCase naming convention. Consider renaming 'StateCopy' to 'stateCopy'.

darokrk (changes requested):
Some improvements needed ⚒️

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+        }
+        emptyArray.push(Object.assign({}, stateCopy));
+        break;
+
+      default:
+        throw new Error('Whoops, something bad happened');
+    }
+  }
+
+  return emptyArray;
```
Radoslaw-Czerniawski, line 39: or better yet, instead of pushing these values to an array you could just return a map statement:
image

Radoslaw-Czerniawski (approved):
Solution works, take a look at my suggestions and consider implementing something similar in  the future

src/transformStateWithClones.js
KarolinaWiktoria: Okay! Thanks for your suggestions :)

PR #2496, opened 2023-07-06
Solution

Roman-Bilichenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+  let currentState = Object.assign({}, state);
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        currentState = Object.assign({}, currentState, action.extraData);
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (currentState.hasOwnProperty(key)) {
```
mykhalenych, line 20: There is no need to check if the property exists before deleting it. The delete operation is safe and will not do anything if the property does not exist.

mykhalenych (approved):

PR #2501, opened 2023-07-09
Solution

Illia-L:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const prop of action.keysToRemove) {
+          delete currentState[prop];
+        }
+        break;
+
+      case 'clear':
+        for (const prop in currentState) {
```
volodymyr-soltys97, line 26: When the 'clear' action is encountered, the currentState is cleared by deleting each property. A more efficient approach would be to assign an empty object to currentState.

volodymyr-soltys97 (changes requested):
Good job 👍 

nazarmatsevych (approved):

PR #2502, opened 2023-07-10
add task solution

Danylo-Chumakov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+          ...nextState, ...action.extraData,
+        };
+      }
+    }
+    stateVersions.push(nextState);
+    currentState = nextState;
+  }
+
+  return stateVersions;
 }
```
nazarmatsevych, line 38: Good work but let's rewrite it using switch/case since we rely on a single action.type value

nazarmatsevych (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateVersions = [];
+  let currentState = state;
```
volodymyr-soltys97, line 11: You are assigning the reference of state object to the currentState. Any modifications on currentState will also affect the original state object. Use spread syntax or Object.assign to create a shallow copy of the state object.

volodymyr-soltys97 (changes requested):
Good job 👍 

PR #2503, opened 2023-07-10
task

pti4ka121:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const actionArray = [];
+  let clonedState = { ...state };
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
```
Anastasiia-Svintsova, line 17: It's better to use constants for string values to avoid typos that can lead to unpredictable errors in the code

Anastasiia-Svintsova (approved):

PR #2506, opened 2023-07-12
complete task

Horizon-git:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,33 @@
  *
```
OleziO, line 6: Remove all comments from the code.

src/transformStateWithClones.js
```diff
@@ -6,8 +6,33 @@
...
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const stateAction = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```
OleziO, line 16: Move all cases to the object.

Example of usage: 

OleziO (changes requested):

anastasiiavorobiova (approved):
Awesome!

PR #2509, opened 2023-07-12
solution

julius0724:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+      Object.assign(currentState, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete currentState[key];
+      }
+    }
+
+    if (action.type === 'clear') {
```
Viktor-Kost, line 24: It'd be nice to use switch statement to manage cases

Viktor-Kost (changes requested):
Nice!
One moment can be improved

PR #2510, opened 2023-07-12
solution

BilokonYaroslav:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const semyResult = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```

Example of usage:

OleziO (changes requested):

src/transformStateWithClones.js
BilokonYaroslav: I don't really think that i understand correctly, but I hope

src/transformStateWithClones.js
```diff
@@ -11,18 +11,18 @@ function transformStateWithClones(state, actions) {
...
         for (const key of action.keysToRemove) {
           delete semyResult[key];
         };
         break;
 
-      case 'clear':
+      case (!action.extraData && !action.keysToRemove):
         for (const key in semyResult) {
           delete semyResult[key];
         };
```
demetrxx:
```
const operations = {
  add: 'addProperties',
  remove: 'removeProperties',
  clear: 'clearProperties',
}

switch (action.type) {
  case(operations.add) {...
  }
  case(operations.remove) ...
}
```

demetrxx (changes requested):
Let's try one more time, keep it up!

src/transformStateWithClones.js
BilokonYaroslav: Thank you

TarasHoliuk (approved):

PR #2512, opened 2023-07-13
add task solution

Vayts:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+      case 'clear':
+        Object.keys(copyState).forEach((key) => delete copyState[key]);
+        break;
+      case 'removeProperties':
+        item.keysToRemove.forEach((key) => {
+          delete copyState[key];
+        });
+        break;
+      default:
+        break;
```
tiserett, line 30: I guess it's gonna be better, if you add gaps between blocks

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+
+  actions.forEach((item) => {
+    const copyState = result.length
```
tiserett, line 13: copyState sounds like an action
I guess stateCopy is a better name

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+      ? { ...result[result.length - 1] }
+      : { ...state };
```
tiserett, line 15: to get last element it's easier to use at(-1)

btw, not sure if you need it here, you can just create this variable outside of loop, work with it, and every time you push it -> just push it's copy

Vayts: This works on LTS node, but not on 14.21.3..... I don't understand why...

edgergard (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+
+      case 'removeProperties':
+        action.keysToRemove.forEach((key) => {
+          delete stateCopy[key];
+        });
+        break;
+
+      default:
+        break;
```
Moroz-Dmytro, line 33: Would be great to throw an error in such case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+    }
+    stateHistory.push(stateCopy);
```
Moroz-Dmytro, line 35:
```suggestion
    }
    
    stateHistory.push(stateCopy);
```

Moroz-Dmytro (changes requested):
Well done 👍 
Small comments

roman-mirzoian (approved):
Well done 👍

PR #2513, opened 2023-07-13
Stateful Clones

Kacper-Leman:
Soultion

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,27 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let newState = { ...state };
+
+  actions.forEach((action) => {
+    let clonedState = { ...newState };
+
+    if (action.type === 'addProperties') {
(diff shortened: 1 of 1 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let newState = { ...state };
+
+  actions.forEach((action) => {
```
DorotaLeniecDev, line 13: According to the checklist:
_[CODE STYLE]: Use switch statement if you have limited amount of conditions._

Try refactoring using **switch**

src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
...
+    let clonedState = { ...newState };
```
DorotaLeniecDev, line 14: I think you dont need to create **clonedState** varibale as you can use **newState** that you have already created

DorotaLeniecDev (changes requested):
Refactor using switch

DorotaLeniecDev (approved):
Good job!

PR #2514, opened 2023-07-13
clones solution

Livan94:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [Object.assign({}, state)];
+  let cloneObject = result[0];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```

Example of usage:

OleziO (changes requested):

demetrxx (approved):
Interesting solution. However, mentor suggested smth like this:
```
const operations = {
  add: 'addProperties',
  remove: 'removeProperties',
  clear: 'clearProperties',
}

switch (action.type) {
  case(operations.add) {...
  }
  case(operations.remove) ...
}
```

PR #2516, opened 2023-07-15
Solution

RomanenkoStud:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete newState[key];
+        }
+        break;
+      case 'clear':
+        newState = {};
+        break;
+      default:
+        return Error('Bad action type');
```
volodymyr-soltys97, line 31: Instead of returning an Error object when the action type doesn't match any cases, consider throwing an error with 'throw new Error('Bad action type')'. Returning an Error object may lead to unexpected behavior as the function is expected to return an array of states.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [];
+  let prevState = { ...state };
+
+  for (const action of actions) {
+    let newState = { ...prevState };
+
+    switch (action.type) {
```
volodymyr-soltys97, line 16: You can use destructurization here for action
```suggestion
  for (const action of actions) {
    let newState = { ...prevState };
    const {} = action;

    switch (action.type) {
```

volodymyr-soltys97 (changes requested):
Good job 👍 

nazarmatsevych (approved):

PR #2519, opened 2023-07-16
Solution

oksana-vas:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+      Object.assign(stateCopy, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete stateCopy[key];
+      }
+    }
+
+    if (action.type === 'clear') {
```
VitaliyBondarenko1982, line 24: consider to use `switch/case` instead a lot of `if`

VitaliyBondarenko1982 (changes requested):

VitaliyBodnar (approved):
Good job!

PR #2523, opened 2023-07-19
Solution added

kyrylomanko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+
+      case 'clear':
+        Object.keys(stateCopy).forEach(key => delete stateCopy[key]);
+        break;
+
+      case 'default':
+        break;
+    }
+
+    stateClones.push({ ...stateCopy });
```
lerastarynets, line 31: Here it is not necessary to make a copy of stateCopy while pushing it into the stateClones array. Since we're using a new object for stateCopy on every iteration, you can just push the stateCopy directly.

lerastarynets (changes requested):

qoa11a (changes requested):
It seems like you forgot to push your changes.

qoa11a (changes requested):
^^

kyrylomanko:
> It seems like you forgot to push your changes.

Hi, Ihor. Anastasia told me in chat that I can leave copying 'cause tests start failing when I remove that

Anastasiia-Svintsova (approved):

PR #2524, opened 2023-07-19
Solution

oksom:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+
+        break;
+
+      default:
+        return state;
+    }
+
+    const newState = JSON.parse(JSON.stringify(copyState));
+
+    newArray.push(newState);
```
polosanya, line 44:
```suggestion
    newArray.push({ ...copyState });
```
You can just push a copy of copy instead

src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+
+        for (const key of removeKeys) {
+          delete copyState[key];
+        }
+
+        break;
+
+      case 'clear':
+
+        for (const key in copyState) {
```
polosanya, line 32:
```suggestion
      case 'clear':
        for (const key in copyState) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+          delete copyState[key];
+        }
+
+        break;
+
+      default:
+        return state;
+    }
```
polosanya, line 40: It would be great to better throw an error instead of returning state

polosanya (approved):
Good job!

PR #2526, opened 2023-07-20
Added solution

FurArt:
Hi, very nice task

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,63 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCloneArray = [];
+  let indexLastActions = 0;
+
+  actions.forEach((action) => {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'clear':
+        stateCloneArray.push(Object.assign({}));
+        break;
+
+      case 'addProperties':
+
+        if (!stateCloneArray.length) {
+          stateCloneArray.push(Object.assign({}, state, extraData));
+          break;
+        }
+
+        indexLastActions = stateCloneArray.length - 1;
+
+        stateCloneArray.push(
+          Object.assign({}, stateCloneArray[indexLastActions], extraData)
+        );
+
+        break;
+
+      default:
+        if (!stateCloneArray.length) {
+          stateCloneArray.push(Object.assign({}, state));
+
+          Object.values(keysToRemove).forEach((removeProperties) => {
+            if (removeProperties in stateCloneArray[0]) {
+              delete stateCloneArray[0][removeProperties];
+            }
+          });
+          break;
+        }
+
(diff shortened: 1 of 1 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,63 @@
...
+
+        indexLastActions = stateCloneArray.length - 1;
+
+        stateCloneArray.push(
+          Object.assign({}, stateCloneArray[indexLastActions], extraData)
+        );
+
+        break;
+
+      default:
```
volodymyr-soltys97, line 36: Instead of using 'default', better to use 'case 'removeProperties':' to improve readability and avoid other possible types of actions.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,63 @@
...
+        if (!stateCloneArray.length) {
+          stateCloneArray.push(Object.assign({}, state));
+
+          Object.values(keysToRemove).forEach((removeProperties) => {
```
volodymyr-soltys97, line 40: Use 'keysToRemove.forEach' instead of 'Object.values(keysToRemove).forEach' because keysToRemove is an array, not an object.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,63 @@
...
+        indexLastActions = stateCloneArray.length - 1;
+
+        stateCloneArray.push(
+          Object.assign({}, stateCloneArray[indexLastActions], extraData)
+        );
+
+        break;
+
+      default:
+        if (!stateCloneArray.length) {
```
volodymyr-soltys97, line 37: Again, no need to treat the first action different from others. You should always clone the last state and apply the action.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,63 @@
...
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'clear':
+        stateCloneArray.push(Object.assign({}));
+        break;
+
+      case 'addProperties':
+
+        if (!stateCloneArray.length) {
```
volodymyr-soltys97, line 23: No need to treat the first action different from others. You should always clone the last state and apply the action.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,63 @@
...
+
+        indexLastActions = stateCloneArray.length - 1;
+
+        stateCloneArray.push(
+          Object.assign({}, stateCloneArray[indexLastActions])
+        );
+
+        indexLastActions += 1;
+
+        Object.values(keysToRemove).forEach((removeProperties) => {
```
volodymyr-soltys97, line 56: Use 'keysToRemove.forEach' instead of 'Object.values(keysToRemove).forEach' because keysToRemove is an array, not an object.

volodymyr-soltys97 (changes requested):
Good job 👍 

volodymyr-soltys97 (approved):
Nice work 🔥

PR #2528, opened 2023-07-20
finished task

helenmarth:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        pushToResult();
+        break;
+      case 'clear':
+        for (const key in stateClone) {
+          delete stateClone[key];
+        }
+        pushToResult();
+        break;
+    }
+  }
```
IvanFesenko, line 38: You can push after switch/case once
Don't forget add default case
```suggestion

  for (const action of actions) {
    const { type, extraData, keysToRemove } = action;

    switch (type) {
      case 'addProperties':
        (Object.assign(stateClone, extraData));
        break;
      case 'removeProperties':
        for (const key of keysToRemove) {
          delete stateClone[key];
        }
        break;
      case 'clear':
        for (const key in stateClone) {
          delete stateClone[key];
        }
        break;
    }
    result.push({ ...stateClone });
  }
```

IvanFesenko (changes requested):
Good job. Check my comment

etojeDenys (approved):
lgtm

PR #2529, opened 2023-07-20
Solution

odolga:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(stateCopy)) {
+          delete stateCopy[key];
+        }
+        break;
+
+      default:
```
Viktor-Kost, line 33: add either break to a default case or it's better to throw an error

Viktor-Kost (changes requested):

nazarmatsevych (approved):

PR #2531, opened 2023-07-20
Solution

HaniaNassalska:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+    switch (typeOfAction) {
+      case 'addProperties':
+        stateCopy = {
+          ...stateCopy,
+          ...action.extraData,
+        };
+        break;
+
+      case 'removeProperties':
+        stateCopy = { ...stateCopy };
```
choeqq, line 25: you don't have to create another 'fresh' copy - the one before is also valid 👍

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
+        break;
+
+      case 'clear':
+        stateCopy = {};
```
choeqq, line 33: please delete each key instead of reassigning new, empty object

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+
+        break;
+      default:
+        throw Error('Something went wrong');
+    }
+    result.push(stateCopy);
```
choeqq, line 39: I think you should spread `stateCopy` inside there
```suggestion
    result.push({ ...stateCopy });
```

choeqq (changes requested):

choeqq (approved):
Good job ✅

PR #2534, opened 2023-07-22
made solution

vadymboichenkoAcupower:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const ResultObject = [];
+
+  for (const operation of actions) {
```
TarasHoliuk, line 12:
```suggestion
  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const ResultObject = [];
```
TarasHoliuk, line 10:
```suggestion
  const stateClones = [];
```
camelCase for variable names
use meaningful names (it's not an object - it's array with state clones)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+
+    if (ResultObject.length !== 0) {
+      Object.assign(CurrentObject, ResultObject[ResultObject.length - 1]);
+    }
+
+    if (ResultObject.length === 0) {
+      Object.assign(CurrentObject, state);
+    }
+
+    if (operation.type === 'addProperties') {
```
TarasHoliuk, line 23: use `switch` block instead of many ifs

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+  for (const operation of actions) {
+    const CurrentObject = {};
+
+    if (ResultObject.length !== 0) {
+      Object.assign(CurrentObject, ResultObject[ResultObject.length - 1]);
+    }
+
+    if (ResultObject.length === 0) {
+      Object.assign(CurrentObject, state);
+    }
```
TarasHoliuk, line 21: instead of doing it copy the initial state into `currentState` object before the loop
and before pushing it to the array make a copy

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const ResultObject = [];
+
+  for (const operation of actions) {
+    const CurrentObject = {};
```
TarasHoliuk, line 13:
```suggestion
    const currentState = Object.assign({}, state);
```
camelCase
move it out of loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+      Object.assign(CurrentObject, ResultObject[ResultObject.length - 1]);
+    }
+
+    if (ResultObject.length === 0) {
+      Object.assign(CurrentObject, state);
+    }
+
+    if (operation.type === 'addProperties') {
+      Object.assign(CurrentObject, operation.extraData);
+      ResultObject.push(CurrentObject);
```
TarasHoliuk, line 25: don't do it in every `if`/`case` - just do it at the end of each loop iteration
push a copy of the object

TarasHoliuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+  const stateClones = [];
+
+  for (const action of actions) {
+    const currentState = {};
+
+    if (stateClones.length !== 0) {
+      Object.assign(currentState, stateClones[stateClones.length - 1]);
+    }
+
+    if (stateClones.length === 0) {
```
anastasiiavorobiova, line 19: The condition 'if (stateClones.length === 0)' is not necessary. You can use Object.assign() directly to copy the state. This will simplify the code.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClones = [];
+
+  for (const action of actions) {
+    const currentState = {};
```
anastasiiavorobiova, line 13: The currentState object is being recreated for every action. A more efficient way would be to clone the state object once before the for loop and then modify that clone based on the actions.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+
+    if (stateClones.length !== 0) {
```
anastasiiavorobiova, line 15: The condition 'if (stateClones.length !== 0)' is not necessary. You can directly use Object.assign() on an empty object and it will just return an empty object back if there is nothing to copy.

anastasiiavorobiova (changes requested):

vadymboichenkoAcupower:
But its right to do this way. Because for every step i need to get copy from array or to have just empty, If i move it beside for loop i got errors. So i think it's right way. I do not crearly understand how to do it in way which you are asking

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+      Object.assign(currentState, stateClones[stateClones.length - 1]);
+    }
+
+    if (stateClones.length === 0) {
+      Object.assign(currentState, state);
+    }
+
+    switch (action.type) {
```
volodymyr-soltys97, line 23: Use a destructurization here for `action`
```suggestion
   const {} = action;
   
    switch (type) {
```

volodymyr-soltys97 (changes requested):
Still not fixed from the previous review.
You can make a copy of the object before the loop

src/transformStateWithClones.js
anastasiiavorobiova: Not resolved from the previous review

anastasiiavorobiova (changes requested):
Great work! Consider fixing previous comments

nazarmatsevych (approved):

PR #2535, opened 2023-07-23
Develop

myroslava-kukhta:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,42 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = { ...state };
```
maxim2310, line 12: There is no need to create an empty `newState` object here, as it is reassigned later in all cases. Creating a new object only to reassign it later is inefficient.

maxim2310 (approved):
Good Job!!!!

PR #2538, opened 2023-07-23
add task

VladyslavKolisnyk:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,33 @@
...
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const stateArr = [];
+
+  for (const action of actions) {
+    switch (action.type) {
```
volodymyr-soltys97, line 15: You can use a desctructurization here for action
```suggestion
  for (const action of actions) {
    const {} = action;
    
    switch (action.type) {
```

volodymyr-soltys97 (changes requested):
Good job 👍 

anastasiiavorobiova (approved):
Awesome!

PR #2547, opened 2023-07-25
Solution

klaudia-klis:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        clonedObject = {
+          ...clonedObject,
+          ...action.extraData,
+        };
+        break;
+      case 'removeProperties':
+        clonedObject = { ...clonedObject };
```
choeqq, line 22: we've already cloned base object 😄

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+
+        for (const key of action.keysToRemove) {
+          delete clonedObject[key];
+        }
+        break;
+      case 'clear':
+        clonedObject = {};
```
choeqq, line 29: please clear every object key instead of assigning new, empty one

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        break;
+      default:
+        return 'Error';
```
choeqq, line 32: I think throwing new Error would be little better

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+    }
+    stateWithClones.push(clonedObject);
```
choeqq, line 34: `clonedObject` should be spreaded
```suggestion
    stateWithClones.push({ ...clonedObject });
```

choeqq (changes requested):

choeqq (approved):
Good job ✅

PR #2548, opened 2023-07-25
Solution

szymonpachucki:


src/transformStateWithClones.js
```diff
@@ -3,11 +3,47 @@
 /**
  * @param {Object} state
  * @param {Object[]} actions
- *
- * @return {Object[]}
+ * @returns {Object[]}
  */
-function transformStateWithClones(state, actions) {
-  // write code here
+function transformState(state, actions) {
```
HaniaNassalska, line 8: Everything looks good, but why did you rename the function? 😉

HaniaNassalska (approved):

src/transformStateWithClones.js
asakevych: @HaniaNassalska is right, no need to rename function

src/transformStateWithClones.js
```diff
@@ -3,11 +3,47 @@
...
+  const result = [];
+  let cloneOfState = { ...state };
```
asakevych, line 10: please read about let and const for objects, i think it could be const 😉

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#const_in_objects_and_arrays

src/transformStateWithClones.js
```diff
@@ -3,11 +3,47 @@
...
+
+  for (const currentAction of actions) {
+    let nextState = { ...cloneOfState };
```
asakevych, line 13: no need to create this variable, you can just use `cloneOfState`

asakevych, line 13: no need to create new variable, you can just use `cloneOfState`

src/transformStateWithClones.js
asakevych, line 10: I think it should be const here 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#const_in_objects_and_arrays

src/transformStateWithClones.js
```diff
@@ -3,11 +3,47 @@
...
+      case 'clear':
+        nextState = {};
+        break;
+
+      default:
+        break;
+    }
+
+    result.push(nextState);
+    cloneOfState = nextState;
```
asakevych, line 43: just modify `cloneOfState` instead of `nextState` and then push it to result in the end, no need for 2 variables 😉

asakevych (changes requested):
Good job! one small change is needed 😉

choeqq (changes requested):
Waiting for any updates @szymonpachucki 👀

src/transformStateWithClones.js
```diff
@@ -1,13 +1,48 @@
...
+  const nextState = { ...state };
+
+  for (const currentAction of actions) {
+    switch (currentAction.type) {
+      case 'addProperties':
+        const extraData = currentAction.extraData;
+
+        for (const key in extraData) {
+          nextState[key] = extraData[key];
+        }
```
JoyCoffeeAddict, line 19: You can simplify it using `Object.assign`

src/transformStateWithClones.js
```diff
@@ -1,13 +1,48 @@
...
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const nextState = { ...state };
+
+  for (const currentAction of actions) {
+    switch (currentAction.type) {
+      case 'addProperties':
+        const extraData = currentAction.extraData;
```
JoyCoffeeAddict, line 15: You can use [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)

src/transformStateWithClones.js
```diff
@@ -1,13 +1,48 @@
...
+        break;
+
+      case 'clear':
+        for (const key in nextState) {
+          delete nextState[key];
+        }
+        break;
+
+      default:
+        break;
```
JoyCoffeeAddict, line 39: It would be nice to show some kind of error if `action.type` is different than we expect. Maybe use `console.error` to do this.

JoyCoffeeAddict (changes requested):

src/transformStateWithClones.js
```diff
@@ -1,13 +1,49 @@
...
+      case 'clear':
+        for (const key in nextState) {
+          delete nextState[key];
+        }
+        break;
+
+      default:
+
+        // console.error('This is an error message');
+        //  i can not use it as i get a "unexpected console statment" error
```
mvjl000, line 39: You could try this:
```suggestion
        throw Error('Something went wrong');
```

mvjl000 (changes requested):

choeqq (approved):
Looks good to me 🚀

choeqq (approved):

ihor-jpeg (approved):

PR #2550, opened 2023-07-25
add state transform function

Meg-Sowka:


domalewski (approved):
everything is fine

.editorconfig
```diff
@@ -8,3 +8,4 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 end_of_line = lf
+quote_type = single
```
asakevych, line 11: to format single qutes only for js files 😉 
```suggestion
end_of_line = lf
[*.js]
quote_type = single
```

asakevych (approved):
Great job!! 🚀 🔥 👏

.editorconfig
Meg-Sowka: Thank you so much!!! Works magic!

ihor-jpeg (approved):

PR #2551, opened 2023-07-25
Solution

obadyda:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  const result = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    const { type } = action;
```
VitaliyBondarenko1982, line 14:
```suggestion
    const { type, extraData, keysToRemove } = action;
```
other fields from action also possible to destructure

VitaliyBondarenko1982 (approved):
Looks good

PR #2552, opened 2023-07-25
Solution

Dolnys:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
-  // write code here
+  const result = [];
+  let currentState = Object.assign({}, state);
+
+  for (const action of actions) {
+    const { type } = action;
+
+    switch (type) {
+      case "addProperties":
+        currentState = Object.assign({}, currentState, action.extraData);
```
choeqq, line 18: we don't have to reassign currentState
```suggestion
        Object.assign(currentState, action.extraData);
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = Object.assign({}, state);
```
choeqq, line 11: if you'll be using that like down below you can change let to const

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+  for (const action of actions) {
+    const { type } = action;
+
+    switch (type) {
+      case "addProperties":
+        currentState = Object.assign({}, currentState, action.extraData);
+        result.push(Object.assign({}, currentState));
```
choeqq, line 19: [checklist ](https://github.com/mate-academy/js_stateful-clones/blob/master/checklist.md)point 4

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+      case "removeProperties":
+        for (const key of action.keysToRemove) {
+          if (currentState.hasOwnProperty(key)) {
+            delete currentState[key];
+          }
+        }
+        result.push(Object.assign({}, currentState));
```
choeqq, line 27: [checklist ](https://github.com/mate-academy/js_stateful-clones/blob/master/checklist.md)point 4

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+      case "clear":
+        currentState = {};
+        result.push(Object.assign({}, currentState));
```
choeqq, line 31: [checklist ](https://github.com/mate-academy/js_stateful-clones/blob/master/checklist.md)point 4, we can also just spread state clone 😄

choeqq (changes requested):

Dolnys:
I made corrections, I didn't just change 'const' to 'let' because, in the case of 'clear', I assign an empty object to it

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        for (const key of action.keysToRemove) {
+          if (currentState.hasOwnProperty(key)) {
+            delete currentState[key];
+          }
+        }
+        result.push({ ...currentState });
+        break;
+
+      case 'clear':
+        currentState = {};
```
choeqq, line 32: that's little hacky I think the point of this task is to modify the `currentState` object, instead of reassigning it to new one - please read: [passing by reference vs value](https://flexiple.com/javascript/javascript-pass-by-reference-or-value/)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+  const result = [];
+  let currentState = Object.assign({}, state);
+
+  for (const action of actions) {
+    const { type } = action;
+
+    switch (type) {
+      case 'addProperties':
+        Object.assign(currentState, action.extraData);
+        result.push({ ...currentState });
```
choeqq, line 19: we're doing that in every case so this can be moved outside of switch, to loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (currentState.hasOwnProperty(key)) {
+            delete currentState[key];
+          }
+        }
+        result.push({ ...currentState });
```
choeqq, line 28: we're doing that in every case so this can be moved outside of switch, to loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        break;
+
+      case 'clear':
+        currentState = {};
+        result.push({ ...currentState });
```
choeqq, line 33: we're doing that in every case so this can be moved outside of switch, to loop

choeqq (changes requested):
Sorry for my oversight, I missed 2 comments, please apply my suggestions and you're good to 🚀

choeqq (approved):
Well done ✅

PR #2553, opened 2023-07-25
add task solution

ValeriiaFomenko:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,46 @@
...
+        newState = {};
+        break;
+      default:
+        newState = currentState;
+        break;
+    }
+
+    result.push(newState);
+    currentState = newState;
+  });
```
etojeDenys, line 46: we can just clone state in push method

etojeDenys (changes requested):

ivbutsykin (approved):
Good job!

PR #2556, opened 2023-07-25
Solution

Vira-V:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,37 @@
...
+        for (const key of action.keysToRemove) {
+          delete clonedState[key];
+        }
+        break;
+
+      case 'clear':
+        clonedState = {};
+        break;
+
+      default:
```
AndrzejPec, line 32: default should throw an error I think, according to the checklist:

AndrzejPec (changes requested):
one minor case to improve found :)

src/transformStateWithClones.js
```diff
@@ -6,8 +6,37 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete clonedState[key];
+        }
+        break;
+
+      case 'clear':
+        clonedState = {};
```
choeqq, line 29: we should delete each key in object instead of reassigning it to empty object

choeqq (changes requested):

choeqq (approved):
Good job ✅

PR #2559, opened 2023-07-26
add stateful

WaveOfDandelions:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+
+        for (const key of action.keysToRemove) {
+          if (stateCopy.hasOwnProperty(key)) {
+            delete stateCopy[key];
+          }
+        }
+        break;
+
+      case 'clear':
+        stateCopy = {};
```
choeqq, line 32: please try to delete all object properties instead of reassigning new, empty one

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+        break;
+
+      default:
+        break;
```
choeqq, line 36: please throw an Error when default case is reached

choeqq (changes requested):

src/transformStateWithClones.js
WaveOfDandelions: > please try to delete all object properties instead of reassigning new, empty one

done

src/transformStateWithClones.js
WaveOfDandelions: > please throw an Error when default case is reached

done

choeqq (approved):
Good job ✅

PR #2564, opened 2023-07-26
solution

kuzgan:


src/transformStateWithClones.js
```diff
@@ -1,13 +1,41 @@
 'use strict';
 
+const ADD_PROPERTIES = 'addProperties';
+const REMOVE_PROPERTIES = 'removeProperties';
+const CLEAR = 'clear';
```
choeqq, line 5: nice 👍

src/transformStateWithClones.js
```diff
@@ -1,13 +1,41 @@
...
  * @param {Object[]} actions
  *
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClone = { ...state };
+  const stateLog = [];
+
+  for (const act of actions) {
```
choeqq, line 17:
```suggestion
  for (const action of actions) {
```
would be more reasonable, but that also works 😉

choeqq (approved):
Good job ✅

PR #2566, opened 2023-07-26
solution

MaciejKujawa:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,40 @@
...
+          delete stateCopy[key];
+        }
+        break;
+
+      default:
+        throw new Error('unsupported value');
+    }
+
+    actionsOnState.push(Object.assign({}, stateCopy));
+  }
```
AndrzejPec, line 40: my suggestion would be to use the spread operator when it is possible (i.e. wherever you're not assigning it to the function parameter), I find this more readable:
```suggestion
    actionsOnState.push({ ...stateCopy });
```

AndrzejPec (changes requested):
tiny detail for sake of readability improvement:

src/transformStateWithClones.js
```diff
@@ -6,8 +6,40 @@
  *
  * @return {Object[]}
  */
+
+const ACTION_ADD_PROPERTIES = 'addProperties';
+const ACTION_REMOVE_PROPERTIES = 'removeProperties';
+const ACTION_CLEAR = 'clear';
```
choeqq, line 12: nice 😉

choeqq (approved):
Good job ✅

PR #2568, opened 2023-07-26
Solution

emil-owczarek:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let currentState = { ...state };
+  const states = [];
+
+  for (const key of actions) {
```
choeqq, line 13: I think actions would be a better name for this variable 😄

choeqq (approved):
Good job ✅

PR #2574, opened 2023-07-29
Solution: Stateful clones

stunnerboyone:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const clonedState = [];
+  const currentStateData = { ...state };
+  const addDataToArray = () => {
+    clonedState.push({ ...currentStateData });
+  };
```
Andriy-Shymkiv, line 15: jesus, is that chatgpt helped you achieve this? 😄

src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
...
+
+  for (const action in actions) {
```
Andriy-Shymkiv: `actions` is array. so what loop we should use?

src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
...
+    switch (actions[action].type) {
+      case 'addProperties':
+        Object.assign(currentStateData, { ...actions[action].extraData });
+        addDataToArray();
+        break;
+      case 'removeProperties':
+        for (const keyToRemove in actions[action].keysToRemove) {
+          delete currentStateData[actions[action].keysToRemove[keyToRemove]];
+        }
```
Andriy-Shymkiv: this loop is to complicated, and you can make it simpler

Andriy-Shymkiv (changes requested):
left comments

src/transformStateWithClones.js
stunnerboyone: No, just watching too many bad youtube channels I think)
I can't use the arrow function or what's wrong here?

src/transformStateWithClones.js
stunnerboyone: already Fixed in next commit

src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const clonedState = [];
+  const currentStateData = { ...state };
+
+  function addDataToArray() {
+    clonedState.push({ ...currentStateData });
+  };
```
xiimiik, line 16: we don't need to do this, we only need to do it once in each iteration, immediately AFTER switch

src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(currentStateData, { ...action.extraData });
```
xiimiik, line 21: no need to use copy here

```suggestion
        Object.assign(currentStateData, action.extraData );
```

xiimiik (changes requested):
Nice work, a little bit left

src/transformStateWithClones.js
stunnerboyone: Thanks, got it!

xiimiik (approved):
GJ 🔥

PR #2575, opened 2023-07-29
stateful clones v1

oltarasov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+      for (const toDelete of obj.keysToRemove) {
+        delete myState[toDelete];
+      }
+      result.push({ ...myState });
+    }
+
+    if (obj.type === 'clear') {
+      myState = {}; // Очищаємо myState
+      result.push({ ...myState });
+    }
```
nazarmatsevych, line 32: Good work but let's rewrite it using switch/case as we rely on single `obj.type` value

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let myState = { ...state };
+  const result = [];
+
+  for (const obj of actions) {
```
```suggestion
  for (const action of actions) {
```

We have array of actions so each element of this array in one action

nazarmatsevych (changes requested):

nazarmatsevych (approved):

PR #2576, opened 2023-07-29
Solution

in7264:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,38 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of value.keysToRemove) {
+          if (currentState.hasOwnProperty(key)) {
+            delete currentState[key];
+          }
+        }
+
+        break;
```
xiimiik, line 32: it's better not to add extra empty lines to switch/case to avoid confusion about which belongs to which
```suggestion
      case 'addProperties':
        currentState = Object.assign({}, currentState, value.extraData);
        break;

      case 'removeProperties':
        for (const key of value.keysToRemove) {
          if (currentState.hasOwnProperty(key)) {
            delete currentState[key];
          }
        }
        break;
```

xiimiik (approved):
Nice work!

PR #2577, opened 2023-07-30
Solution

polinavafik:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      case 'removeProperties':
+        action.keysToRemove.forEach((key) => delete stateCopy[key]);
+        break;
+
+      case 'clear':
+        Object.keys(stateCopy).forEach((key) => delete stateCopy[key]);
+        break;
+
+      default:
+        break;
```
Moroz-Dmytro, line 28: Would be nice to return an error or error message in such case

Moroz-Dmytro (approved):
👍

PR #2578, opened 2023-07-30
soled task

bormaxvel:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = Object.assign({}, state);
+  const listOfStates = [];
+
+  for (const act of actions) {
```
SanyaBratashchuk, line 13:
```suggestion
  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+    switch (act.type) {
```
SanyaBratashchuk, line 14: use object destructuring

SanyaBratashchuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = Object.assign({}, state);
```
```suggestion
  let stateCopy = Object.assign({}, state);
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const property of action.keysToRemove) {
+          delete stateCopy[property];
+        }
+        break;
+
+      case 'clear':
+        Object.keys(stateCopy).forEach(key => delete stateCopy[key]);
```
```suggestion
        stateCopy = {};
```

etojeDenys (approved):

src/transformStateWithClones.js
bormaxvel: It is not entirely clear why this is necessary

src/transformStateWithClones.js
bormaxvel: Will the creation of new objects not affect the optimization if this function is called frequently?

PR #2584, opened 2023-07-31
Solution

neovendano:


AndrMar1939 (approved):
Good job! But let`s improve your code.
Every iteration of the cycle you have created a new "switch case". 
Think about how to optimize it.

alexander-ignatow:
> Good job! But let`s improve your code. Every iteration of the cycle you have created a new "switch case". Think about how to optimize it.

It is not critical - that may little hurt the readability, but not the productivity

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        }
+        break;
+      case 'clear':
+        for (const prop in currentState) {
+          if (currentState.hasOwnProperty(prop)) {
+            delete currentState[prop];
+          }
+        }
+        break;
+    }
```
alexander-ignatow, line 30: You need `default` for your `switch`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        Object.assign(currentState, action.extraData);
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete currentState[key];
+        }
+        break;
+      case 'clear':
+        for (const prop in currentState) {
+          if (currentState.hasOwnProperty(prop)) {
```
alexander-ignatow, line 25: This `if` clause is unnecessary, as it is always `true`

alexander-ignatow (changes requested):
Good, but some things are needed to be improved

src/transformStateWithClones.js
neovendano: I`ll add it, but just to be sure: is it necessary to add 'default' if it doesn`t do anything?

default:
  break;

src/transformStateWithClones.js
```diff
@@ -6,8 +6,44 @@
...
+      case 'removeProperties':
+        stateRemoveProperty(currentState, action.keysToRemove);
+        break;
+
+      case 'clear':
+        stateClear(currentState);
+        break;
+
+      default:
+        break;
```
dffUqp, line 40: would be nice to return an error or error message here

src/transformStateWithClones.js
```diff
@@ -6,8 +6,44 @@
...
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const statesArray = [];
+  const currentState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(currentState, action.extraData);
```
dffUqp, line 28: if you started creating separate functions for actions, why did you skip `addProperties`?
I know it is only one line but better to move it too

dffUqp (changes requested):
Almost done:smiley:

src/transformStateWithClones.js
neovendano: This function would do the same as Object.assign. IMHO, you should create "one-line" functions when we have specific case of method (for example, sqrt() is a specific case of pow() ). But that function would do exactly same thing: assign properties from one object to another.

src/transformStateWithClones.js
alexander-ignatow: It is a good practice - it is always better to be explicit about what your code is doing - even if you just skip the item you should do it

src/transformStateWithClones.js
```diff
@@ -6,8 +6,48 @@
...
+      case 'removeProperties':
+        stateRemoveProperty(currentState, action.keysToRemove);
+        break;
+
+      case 'clear':
+        stateClear(currentState);
+        break;
+
+      default:
+        throw new Error('Invalid input!');
```
alexander-ignatow, line 44: 👍

alexander-ignatow (approved):
LGTM 👍

PR #2585, opened 2023-07-31
add task solution

yar0man:


voievodik (changes requested):
Great job! Keep it going!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
+
+      default:
+        return 'Something went wrong';
```
alexander-ignatow, line 34: This is good!
You defend the code against potential changes

The only problem is the return value - it is invalid and breaks the function contract

I suggest you either return something meaningful or throw an `Error`

alexander-ignatow (changes requested):
Looks good 👍 
Just a minor fix is needed

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
+
+      default:
+        throw new Error('Something went wrong');
```
alexander-ignatow, line 34: Now much better

Hint: ideally error message should be short, but as informative as possible
For instance: 
```suggestion
        throw new Error(`Something went wrong: unexpected  action.type: ${action.type}`);
```

alexander-ignatow (approved):
Looks good now 👍 
Please see the comments ;)

PR #2588, opened 2023-07-31
add task solution

voievodik:


yar0man (approved):
Great job!

AndriiKuz (approved):
Good job!)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+  const operationsResult = [];
+  const stateCopy = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+      case 'removeProperties':
+        action.keysToRemove.map(key => delete stateCopy[key]);
```
alexander-ignatow, line 19: I see your point, but you are using `map` method incorrectly - it is intended to map the array onto some other data and generate a different array of the same size.
But `Array` has a different method that is perfect here ;)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+        break;
+      case 'clear':
+        Object.keys(stateCopy).map(key => delete stateCopy[key]);
```
alexander-ignatow, line 22: Same misuse of `map` as above

alexander-ignatow (changes requested):
Clean solution 👍 
But you need to fix some issues

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const operationsResult = [];
```
tiserett, line 10: stateHistory is a better name

tiserett (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        action.keysToRemove.forEach(key => delete stateCopy[key]);
+        break;
+
+      case 'clear':
+        Object.keys(stateCopy).forEach(key => delete stateCopy[key]);
+        break;
+
+      default:
+        return new Error(`
+          Something went wrong: unexpected  action.type: ${action.type}
```
tiserett, line 29: there are 2 whitespaces after word 'unexpected'

tiserett (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => delete stateCopy[key]);
+        break;
+
+      case 'clear':
+        Object.keys(stateCopy).forEach(key => delete stateCopy[key]);
+        break;
+
+      default:
+        return new Error(`
```
dffUqp, line 28: returning an Error object instead of throwing an exception
```suggestion
        throw new Error(`
```

dffUqp (approved):
Looks good:smiley:, try not to make these mistakes in the following tasks.

PR #2589, opened 2023-07-31
add task solution

Koliras:


davinnchii (commented):
Well done =)
I'd prefer to use forEach instead of common for-loop. Also in my opinion, adding a empty line before `break` will improve readability of your code.

davinnchii (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
-  // write code here
+  const stateCopy = { ...state };
+  const stateVersions = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+      case 'removeProperties':
```
Moroz-Dmytro, line 18:
```suggestion
        break;
        
      case 'removeProperties':
```

Moroz-Dmytro: And in other places, will be easier to read

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+        action.keysToRemove.forEach(key => {
+          delete stateCopy[key];
+        });
+        break;
+      case 'clear':
+        Object.keys(stateCopy).forEach(key => {
+          delete stateCopy[key];
+        });
+        break;
+      default: break;
```
Moroz-Dmytro, line 28: Would be nice to throw an error in such case

Moroz-Dmytro (changes requested):
Good job 👍 
Small comments

tiserett (changes requested):

PR #2593, opened 2023-07-31
solution

Patryk-Buczkowski:


DorotaLeniecDev (approved):
Good job! You could remove linebreaks at 16, 22, 30 and it would be perfect :)

PR #2596, opened 2023-07-31
add task solution

Wesses:


Orpheus29 (approved):
Good Job😎. Clever move with slice(-1)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,47 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const TYPE_ADD_PROPERTIES = 'addProperties';
+  const TYPE_REMOVE_PROPERTIES = 'removeProperties';
+  const TYPE_CLEAR = 'clear';
+  const statesList = [{ ...state }];
+
+  for (const action of actions) {
+    const currentObject = { ...statesList.slice(-1)[0] };
```
alexander-ignatow, line 16: The idea with `slice` is interesting indeed

src/transformStateWithClones.js
```diff
@@ -7,7 +7,47 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const TYPE_ADD_PROPERTIES = 'addProperties';
+  const TYPE_REMOVE_PROPERTIES = 'removeProperties';
+  const TYPE_CLEAR = 'clear';
```
alexander-ignatow, line 12: 👍  you are using constants, that is great

But these constants should be declared outside of the function, otherwise you will re-create them on each function call

alexander-ignatow (changes requested):
Interesting and valid solution 👍 

Still you need to fix some minor issues

src/transformStateWithClones.js
```diff
@@ -6,8 +6,50 @@
...
+        statesList.push(
+          {
+            ...currentObject,
+            ...objectWithNewFields,
+          });
+        break;
+      }
+
+      case TYPE_REMOVE_PROPERTIES: {
+        const arrayOfFields = action.keysToRemove;
```
tiserett, line 33: arrayOfFields -> arrayOfKeysToRemove / keysToRemove

src/transformStateWithClones.js
```diff
@@ -6,8 +6,50 @@
...
+
+        for (const field of arrayOfFields) {
```
tiserett, line 35: field -> key

tiserett (changes requested):
how about you just create an object, which you will push after certain action, not inside every of them

src/transformStateWithClones.js
```diff
@@ -6,8 +6,51 @@
...
+      }
+
+      case TYPE_REMOVE_PROPERTIES: {
+        const keysToRemove = action.keysToRemove;
+
+        for (const key of keysToRemove) {
+          delete currentObject[key];
+        }
+
+        currentObject = { ...currentObject };
```
dffUqp, line 38: unnecessary object copying

src/transformStateWithClones.js
```diff
@@ -6,8 +6,51 @@
...
+        break;
+      }
+
+      case TYPE_CLEAR: {
+        currentObject = {};
+        break;
+      }
+
+      default: return 'undefined type';
```
dffUqp, line 47: better to throw error here

src/transformStateWithClones.js
```diff
@@ -6,8 +6,51 @@
...
+const TYPE_ADD_PROPERTIES = 'addProperties';
+const TYPE_REMOVE_PROPERTIES = 'removeProperties';
+const TYPE_CLEAR = 'clear';
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const statesList = [{ ...state }];
+
+  for (const action of actions) {
+    let currentObject = { ...statesList.slice(-1)[0] };
```
dffUqp, line 18:
```suggestion
    let currentObject = { ...statesList[statesList.length - 1] };
```

dffUqp (changes requested):
Almost done:smiley:

src/transformStateWithClones.js
```diff
@@ -6,8 +6,49 @@
...
+
+        currentObject = {
+          ...currentObject,
+          ...objectWithNewFields,
+        };
+        break;
+      }
+
+      case TYPE_REMOVE_PROPERTIES: {
+        const keysToRemove = action.keysToRemove;
```
Moroz-Dmytro, line 32:
```suggestion
        const { keysToRemove } = action;
```

src/transformStateWithClones.js
```diff
@@ -6,8 +6,49 @@
...
+        };
+        break;
+      }
+
+      case TYPE_CLEAR: {
+        currentObject = {};
+        break;
+      }
+
+      default: return 'error';
```
Moroz-Dmytro, line 45: Better will be to throw an error, not return string 
Like : "throw new Error(`Something went wrong: unexpected action.type: ${action.type}`);"

Moroz-Dmytro (approved):
👍

PR #2598, opened 2023-08-01
add solution

KatOlista:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,54 @@
...
+        break;
+
+      case 'removeProperties':
+        if (resultArr.length === 0) {
+          resultArr.push({ ...state });
+        } else {
+          resultArr.push({ ...resultArr[resultArr.length - 1] });
+        }
+
+        for (const action of actions[i].keysToRemove) {
```
maksym-mishchanchuk, line 32: don*t use cycle in cycle.

src/transformStateWithClones.js
```diff
@@ -6,8 +6,54 @@
...
+          delete resultArr[resultArr.length - 1][action];
+        };
+
+        break;
+
+      case 'clear':
```
maksym-mishchanchuk, line 38: If I were you, I'd create a function for each case

maksym-mishchanchuk (changes requested):
and I would probably use the ternary operator

src/transformStateWithClones.js
```diff
@@ -6,8 +6,45 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultArr = [];
```
BudnikOleksii, line 11: Not the best name for an array.

src/transformStateWithClones.js
```diff
@@ -6,8 +6,45 @@
...
+  const createNewObj = (obj) => {
+    const newObj = resultArr.length === 0
+      ? Object.assign({ ...state })
+      : Object.assign({ ...resultArr[resultArr.length - 1] });
+
+    return newObj;
+  };
```
BudnikOleksii, line 18: It's not a good idea to create functions inside functions. And that depends on something outside the function. The function that creates new objects shouldn't know about "resultArr", it should just create a new object.

BudnikOleksii (changes requested):
Nice work👍
But could be even better, check the comments.

src/transformStateWithClones.js
KatOlista: it's only the way don't repeat this piece of code.  I use it twice

src/transformStateWithClones.js
```diff
@@ -8,34 +8,36 @@
  */
 
 function transformStateWithClones(state, actions) {
-  const resultArr = [];
-  const createNewObj = (obj) => {
-    const newObj = resultArr.length === 0
-      ? Object.assign({ ...state })
-      : Object.assign({ ...resultArr[resultArr.length - 1] });
+  const results = [];
```
dffUqp, line 11: This is from the checklist of one previous task. 

> 3. [NAMING] - use proper names for variables - if you want to name variable result, pick a better one :)

let's give more descriptive name for this variable.

Here are some good examples:
`stateSnapshots`
`stateHistory`

src/transformStateWithClones.js
```diff
@@ -6,8 +6,47 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const results = [];
+
+  const createNewObj = (obj, arr) => {
```
dffUqp, line 13: Are you really sure that you need this function?

I think we can use spred on line 24 and 30
```js
// 24
results.push(Object.assign(...state, action.extraData));

// 30
results.push( { ...results[results.length - 1] } );
```

What does this function do? Creates a new object? Why can't we create it without it? 
I mean creating a new object it's one line of code and this function is also one line of code but inside we have unnecessary logic. 



if you want to leave this solution, it will be better move `createNewObj`  outside `transformStateWithClones`

dffUqp (changes requested):
Almost done:smiley:

roman-mirzoian (approved):
Well done 👍

PR #2600, opened 2023-08-01
solution

Taras2705:


irahos (approved):
think of better naming for resultArray variable
other than that great job🤩

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
```
dffUqp, line 28: we clear all values from our object so can we just assign new empty object to avoid additional loop?

dffUqp (changes requested):
Almost done:smiley:

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let stateCopy = { ...state };
+  const stateChanges = [];
```
tiserett, line 11: I guess stateHistory is a better name, wdyt?

tiserett (changes requested):

alexander-ignatow (approved):
LGTM 👍

PR #2604, opened 2023-08-01
add task solution

rostyslav48:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        action.keysToRemove.map((removeKey) => delete stateCopy[removeKey]);
+
+        break;
+
+      case 'clear':
+        Object.keys(stateCopy).map((key) => delete stateCopy[key]);
+
+        break;
+
+      default: return null;
```
vserhiichuk, line 30: 3.[Сhecklist]: switch/case should always have default case for error handling.

vserhiichuk (changes requested):
It looks good
Just one moment to fix

src/transformStateWithClones.js
rostyslav48: Thanks for your comment, have already made all the changes required)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        action.keysToRemove.map((removeKey) => delete stateCopy[removeKey]);
+
+        break;
+
+      case 'clear':
+        Object.keys(stateCopy).map((key) => delete stateCopy[key]);
+
+        break;
+
+      default: return Error('Something went wrong');
```
dffUqp, line 30:
```suggestion
      default: 
          return Error('Something went wrong');
```

dffUqp (approved):
LGTM🔥

PR #2605, opened 2023-08-01
Solution

ydratsevska:


desevoker (changes requested):
Good job! 🚀

Some minor changes need to be done.
Also left you a recommendation.

desevoker:
You forgot to fix empty `default` in a switch stetement.

Also, you don't need to re-request reviews from your mates, that applies only to mentors' reviews.

dffUqp (approved):
LGTM🚀

PR #2612, opened 2023-08-02
add task solution

dpidlutska:


Denys14Samoilenko (approved):
Nice job!!! 🔥 
I wanted to add a comment that this code does not need additional verification you can just delete property, but I could be wrong 😄 
` for (const key in action.keysToRemove) {
          if (stateCopy.hasOwnProperty(action.keysToRemove[key])) {
            delete stateCopy[action.keysToRemove[key]];
          }
        }`

PR #2614, opened 2023-08-02
Added State Clone

Kapidastra:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+        Object.assign(stateCopy, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (stateCopy.hasOwnProperty(key)) {
+            delete stateCopy[key];
+          }
+        }
+        result.push({ ...stateCopy });
```
TarasHoliuk, line 30: you are doing the same 3 times (including `addProperties` case - anyway you have Object.assign there) - do it once after the `switch` block instead

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const stateCopy = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
+    const action = actions[i];
```
TarasHoliuk, line 14: you may use a common loop here for sure, but it'll look a bit better with `for of`

TarasHoliuk (changes requested):

loralevitska (approved):
Well done)

PR #2618, opened 2023-08-02
Solution for the task

thaiduk:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,36 @@
...
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
+
+      default:
+        return 'Error';
```
Yevheniia-Sid, line 33: would be better to throw an error❌

Yevheniia-Sid (changes requested):
Looks good! but one small change is needed😉

src/transformStateWithClones.js
```diff
@@ -6,8 +6,34 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  let stateCopy = { ...state };
+  const stateCopyActions = [];
```
Moroz-Dmytro, line 12: Try to make up better name, for example `stateHistory`. Because now It looks like array of actions

src/transformStateWithClones.js
```diff
@@ -6,8 +6,34 @@
...
+        break;
+
+      case 'clear':
+        stateCopy = {};
+        break;
+
+      default:
+        throw new Error('Unknown properties');
+    }
+    stateCopyActions.push({ ...stateCopy });
```
Moroz-Dmytro, line 33:
```suggestion
    }
    
    stateCopyActions.push({ ...stateCopy });
```

Moroz-Dmytro (changes requested):
Well done 👍 
Small comments

src/transformStateWithClones.js
thaiduk: thank you for comment. DonE

Moroz-Dmytro (approved):
LGTM 👍

PR #2619, opened 2023-08-02
add task solution

cty2802:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = ({ ...state });
+  const arrayFromState = [];
```
BudnikOleksii, line 11: The "arrayFromState" name doesn't match what actually this array is. statesHistory or statesLogs.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+
+  for (let i = 0; i < actions.length; i++) {
+    const extraDatavalue = actions[i].extraData;
+    const typeOfAction = actions[i].type;
+    const removeValue = actions[i].keysToRemove;
```
BudnikOleksii, line 16: It's better to use "for of" loop instead.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+      case 'removeProperties':
+        for (let j = 0; j < removeValue.length; j++) {
+          delete newState[removeValue[j]];
+        }
+        arrayFromState.push({ ...newState });
+        break;
+
+      default:
+        Object.keys(newState).forEach(key => delete newState[key]);
+        arrayFromState.push({ ...{} });
```
BudnikOleksii, line 35: You do arrayFromState.push({ ...newState }); in every case, you can do it after all cases.

BudnikOleksii (changes requested):
Good job👍
Left a few comments, check them.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+      case 'removeProperties':
+        for (let j = 0; j < removeValue.length; j++) {
+          delete newState[removeValue[j]];
+        }
+
+        break;
+
+      default:
+        Object.keys(newState).forEach(key => delete newState[key]);
+    }
```
annak1104, line 35: if I understand correct this code **"default: Object.keys(newState).forEach(key => delete newState[key])"** are clearing the properties of the newState object. The **default** case should not be used to clear the state object. You should use: default:throw new Error(``). But I'm not 100% sure about it, so I confirm the completion of the task, and the mentor will confirm or discard my opinion.

annak1104 (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+    switch (typeOfAction) {
+      case 'addProperties':
+        for (const newProperty in extraDatavalue) {
+          newState[newProperty] = extraDatavalue[newProperty];
+        }
+
+        break;
+
+      case 'removeProperties':
+        for (let j = 0; j < removeValue.length; j++) {
```
dffUqp, line 27: let's use [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) for this case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+          delete newState[removeValue[j]];
+        }
+
+        break;
+
+      case 'clear':
+        newState = {};
+
```
dffUqp, line 35:
```suggestion
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+          newState[newProperty] = extraDatavalue[newProperty];
+        }
+
+        break;
+
+      case 'removeProperties':
+        for (let j = 0; j < removeValue.length; j++) {
+          delete newState[removeValue[j]];
+        }
+
```
dffUqp, line 30:
```suggestion
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+  for (const action of actions) {
+    const extraDatavalue = action.extraData || {};
+    const typeOfAction = action.type;
+    const removeValue = action.keysToRemove || {};
+
+    switch (typeOfAction) {
+      case 'addProperties':
+        for (const newProperty in extraDatavalue) {
+          newState[newProperty] = extraDatavalue[newProperty];
```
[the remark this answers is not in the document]
dffUqp: good comment

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let newState = ({ ...state });
+  const renovatedState = [];
+
+  for (const action of actions) {
+    const extraDatavalue = action.extraData || {};
+    const typeOfAction = action.type;
+    const removeValue = action.keysToRemove || {};
```
dffUqp, line 16: using for...of you can remove those variables

dffUqp (changes requested):
Almost done:smiley:

src/transformStateWithClones.js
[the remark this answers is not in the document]
BudnikOleksii: The comment is still not fixed

BudnikOleksii (changes requested):
Please, fix the comments provided by previous reviewer.

BudnikOleksii (approved):
Good job👍

PR #2620, opened 2023-08-02
solution

00Mass00:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
+
+/**
+ * @param {Object} state
+ * @param {Object} action
+ */
+function transformState(state, action) {
+  switch (action.type) {
+    case 'addProperties':
+      Object.assign(state, action.extraData);
+      break;
```
BudnikOleksii, line 30: It's a bad practice to mutate parameters. Instead, create a copy and return it from the function.

BudnikOleksii (changes requested):
Good work👍
But it's a bad practice to mutate parameters, let's fix it.

serkrops (approved):
Good job!

dffUqp (approved):
GJ✔️

PR #2627, opened 2023-08-03
add task solution

NazarVotkanych:


thaiduk (changes requested):
Good job.
But, some little comments.
On my mind, its better add for default return some error

BudnikOleksii (approved):
Good job👍

PR #2629, opened 2023-08-03
added a solution

shikoryak:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,23 @@
...
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(newState, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete newState[key];
+      }
+    } else {
+      newState = {};
+    }
```
VitaliyBodnar, line 22: please use a switch case statement instead of if else

VitaliyBodnar (changes requested):
Almost done!

IvanFesenko (changes requested):
Check my comments

mykhalenych (approved):

PR #2630, opened 2023-08-03
add task solution

k1NAs:


bogemna:
Well done! But some mentors can reject because of variables name.

bogemna (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClone = { ...state };
+  const stateClones = [];
```
Moroz-Dmytro, line 11: Try to make up a better name for this. Now it not describe well what will be in this variable

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateClone, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (stateClone.hasOwnProperty(key)) {
+            delete stateClone[key];
+          }
```
Moroz-Dmytro, line 23: You don't need this check. If there is no such key in object - nothing happens

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(stateClone)) {
+          if (stateClone.hasOwnProperty(key)) {
+            delete stateClone[key];
+          }
```
Moroz-Dmytro, line 31: Same here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (stateClone.hasOwnProperty(key)) {
+            delete stateClone[key];
+          }
+        }
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(stateClone)) {
```
Moroz-Dmytro, line 28: You can use [for in ](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Statements/for...in)loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+          if (stateClone.hasOwnProperty(key)) {
+            delete stateClone[key];
+          }
+        }
+        break;
+
+      default:
+        return new Error('Oops, something went wrong!');
```
Moroz-Dmytro, line 36: Better will be to throw an error, not to return

Moroz-Dmytro (changes requested):
Let's improve your code 😉

roman-mirzoian (approved):
Well done 👍

PR #2631, opened 2023-08-03
Solution

Vitalii-Petruk:


AngerDESTROYS (approved):
Good

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultArray = [];
```
BudnikOleksii, line 10: [NAMING] - use proper names for variables - if you want to name variable result, pick a better one :)

BudnikOleksii (changes requested):
Good job👍
Just don't forget about meaningful names for variables

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+        break;
+      }
+
+      case 'clear': {
+        const keysToClear = Object.keys(stateCopy);
+
+        for (const keyToClear of keysToClear) {
+          delete stateCopy[keyToClear];
+        }
+        break;
```
dffUqp, line 33: we clear all values from our object so can we just assign new empty object to avoid additional loop?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+      }
+
+      default:
+        throw new Error();
```
dffUqp, line 37: would be better to add some text for the error

dffUqp (approved):
GJ✔️

PR #2634, opened 2023-08-04
Develop

bogemna:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+
+  actions.forEach((action) => {
+    const stateCopy = result.length
+      ? { ...result[result.length - 1] }
+      : { ...state };
```
tiserett, line 15: looks complicated, you can fix it by storing variable outsife of loop

tiserett (changes requested):

PR #2641, opened 2023-08-04
solution

mr-olezha:


alexander-ignatow (approved):
LGTM 👍 

Good approach with separate method

PR #2643, opened 2023-08-04
solution

Duster-1:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+      case 'addProperties':
+        firstCopyState = {
+          ...firstCopyState,
+          ...action.extraData,
+        };
+
+        break;
+
+      case 'removeProperties':
+        const secondCopyState = { ...firstCopyState };
```
TarasHoliuk, line 24: no need to create a separate variable here (and copy in the `addProperties` case) - you can mutate the `firstCopyState` object because you do the copy when pushing

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let firstCopyState = { ...state };
```
TarasHoliuk, line 11:
```suggestion
  let stateCopy = { ...state };
```
you will not have second)

TarasHoliuk (changes requested):

nazarmatsevych (approved):

PR #2648, opened 2023-08-06
сloning and transforming state

Q1W2E3R4T5Y6U7I8a:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clones = [Object.assign({}, state)];
+
+  actions.forEach((action) => {
+    const { extraData } = action;
```
volodymyr-soltys97, line 13: Consider it will be better
```suggestion
    const { extraData, type, keysToRemove } = action;
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+
+    if (extraData) {
```
volodymyr-soltys97, line 15: Using switch construction here will be better

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

PR #2649, opened 2023-08-06
Solution

Nikitavyskreb:


tiserett (changes requested):
let's add curly braces inside of case block, if there is more then 2-3 lines of code
let's use array methods, instead of for of / in loops

KlymSudakov (approved):

PR #2654, opened 2023-08-07
Solution

InnaSh23:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+      case 'clear':
+        currentState = {};
+
+        break;
+
+      default:
+        break;
+    }
+
+    cloneActions.push(Object.assign({}, currentState));
```
lerastarynets, line 38: You have repeatedly cloned the 'currentState' object throughout your code and in the beginning of every action. This could potentially use a lot of memory if the state and actions arrays are large. A more memory-efficient approach might be to only clone the state when you are about to modify it.

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit more

src/transformStateWithClones.js
InnaSh23: Did I understand the wish correctly?

PR #2657, opened 2023-08-09
Solution

divamal:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+
+    if (type === 'addProperties') {
+      const { extraData } = action;
+
+      Object.assign(currentState, extraData);
+    } else if (type === 'removeProperties') {
+      const { keysToRemove } = action;
+
+      for (const key of keysToRemove) {
+        if (currentState.hasOwnProperty(key)) {
```
VeronikaSynieokaia, line 24: The same as the previous comment. The 'delete' operator can be used without the 'hasOwnProperty' check.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+          delete currentState[key];
+        }
+      }
+    } else if (type === 'clear') {
+      for (const key in currentState) {
```
VeronikaSynieokaia, line 29: There's no need to check if a property belongs to an object before deleting it. The delete operator is safe to use regardless, because it doesn't throw an error if the property doesn't exist.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultArray = [];
+
+  for (const action of actions) {
+    const { type } = action;
+    const currentState = { ...resultArray[resultArray.length - 1] || state };
```
VeronikaSynieokaia, line 14: This line may cause performance issues when the length of 'resultArray' increases, as accessing the last element of an array with the 'length - 1' approach is expensive. Consider keeping a separate variable for the 'currentState' that gets updated with each action.

VeronikaSynieokaia (changes requested):

VadimMentorQA (approved):
Good job

PR #2668, opened 2023-08-13
Solution

MyroslavSynenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const copyState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(copyState, action.extraData);
+        result.push({ ...copyState });
```
VitaliyBodnar, line 17: 'result.push({ ...copyState });' is duplicated, let's put it after the switch case block

VitaliyBodnar (changes requested):
Almost done!

VitaliyBodnar (approved):
Good job!

PR #2669, opened 2023-08-14
Solution

oleg-matvienko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultArray = [];
```
roman-mirzoian, line 10: It is better to choose more meaningful names of the changed, for example: `transformedState`, `stateClone` etc.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+  let currentState = { ...state }; // Creating a copy of the initial state
```
roman-mirzoian, line 11: Why do you write in the comment that this is a copy of the state, but call it the `currentState`? 
Then it would be better to call this variable `copiedState`.
Also, this comment doesn't look very useful, because here the code is "self-explanatory".

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+
+      case 'clear':
+        currentState = {};
+        break;
+
+      default:
+        throw new Error(`Unknown action type: ${action.type}`);
+    }
+
+    // Adding a clone of the current state to the result array
```
roman-mirzoian, line 39: This comment seems redundant.

roman-mirzoian (changes requested):
Good work, it remains to refine the naming a little and remove unnecessary comments.

roman-mirzoian (approved):
Well done 👍

PR #2675, opened 2023-08-17
solution developed

StepanMustafin:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+  const res = [];
+
+  for (let i = 0; i < actions.length; i++) {
+    let temp = null;
+
+    if (i === 0) {
+      temp = { ...state };
+    } else {
+      temp = { ...res[i - 1] };
+    }
```
etojeDenys, line 19: looks redundant, we can just create the copy of state out of the loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+      case 'addProperties':
+        for (const addKey in actions[i].extraData) {
+          temp[addKey] = actions[i].extraData[addKey];
+        }
+        break;
+
+      case 'removeProperties':
+        if (Object.keys(temp).length === 1) {
+          temp = {};
+        } else {
```
etojeDenys, line 35: what happens if we pass the wrong property name to delete it will delete the correct one
do we really want this logic? (i don't think so)

etojeDenys (changes requested):
.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+
+      case 'addProperties':
+        for (const addKey in actions[i].extraData) {
+          temp[addKey] = actions[i].extraData[addKey];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const rmKey of actions[i].keysToRemove) {
+          if (temp.hasOwnProperty(rmKey)) {
```
Anastasiia-Svintsova, line 27: Checking if an object has a property before deleting it is unnecessary, as the delete operation will fail silently if the property does not exist. You can safely delete the property without the check.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const res = [];
+  let temp = { ...state };
```
Anastasiia-Svintsova, line 11: image

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+
+  for (let i = 0; i < actions.length; i++) {
+    switch (actions[i].type) {
```
Anastasiia-Svintsova, line 14: Use destructuring for the objects for better readability

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+      case 'clear':
```
Anastasiia-Svintsova, line 15: Would be great to move all string values to variables to avoid errors connected with typos

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const res = [];
```
Anastasiia-Svintsova, line 10: What does `res` mean? Use more descriptive names

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
-  // write code here
+  const ADD_PROPERTIES = 'addProperties';
+  const REMOVE_PROPERTIES = 'removeProperties';
+  const CLEAR = 'clear';
+
+  const statesArr = [];
+  let stateCopy = { ...state };
+
+  for (const action of actions) {
+    const { type: actionType } = action;
```
Anastasiia-Svintsova, line 18: So, what about the other properties? 
```suggestion
    const { type: actionType, extraData, keysToRemove } = action;
```

Anastasiia-Svintsova (approved):

PR #2679, opened 2023-08-19
Solution

denedweb:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+        break;
+
+      case 'clear':
+        copyState = {};
+        result.push({ ...copyState });
+        break;
+
+      default:
+        break;
+    }
```
etojeDenys, line 36: we can push the object after switch case only time

etojeDenys (changes requested):

SanyaBratashchuk (approved):

PR #2680, opened 2023-08-19
add task solution

kitpes:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,46 @@
...
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+
+        break;
+      }
+
+      default: {
+        break;
+      }
```
nazarmatsevych, line 42: It's better to handle errors in default block, for example you can throw an error with message like "Unknown action type"

nazarmatsevych (approved):
Good work, left minor comment

src/transformStateWithClones.js
kitpes: Added some changes. 
Is this update okay?

PR #2683, opened 2023-08-21
add task solution

markiuan:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    let newState = { ...currentState };
```
Viktor-Kost, line 14: You are creating a new copy of the currentState on every iteration. This is unnecessary as you are already making a new copy inside each action type. You can remove this line.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        delete newState[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      newState = {};
+    }
+
+    result.push(newState);
+    currentState = newState;
```
Viktor-Kost, line 33: The reassigning of currentState to newState is not needed as you are already updating currentState inside each action type.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    let newState = { ...currentState };
+
+    if (action.type === 'addProperties') {
```
Viktor-Kost, line 16: it'd be better to apply `switch` statement instead of `if` one

Viktor-Kost (changes requested):

PR #2691, opened 2023-08-23
Solution

julwer1k:
image

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const lengthActions = [];
+  let stateCopy = { ...state };
+
+  for (const action of actions) {
+    const TYPE_MESSAGE = action.type;
+    const secondValue = action.extraData;
+    const keyRemove = action.keysToRemove;
+
+    switch (TYPE_MESSAGE) {
+      case 'addProperties':
+        stateCopy = {
+          ...stateCopy,
+          ...secondValue,
+        };
+
+        lengthActions.push({ ...stateCopy });
+        break;
+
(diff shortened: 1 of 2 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+    const keyRemove = action.keysToRemove;
+
+    switch (TYPE_MESSAGE) {
+      case 'addProperties':
+        stateCopy = {
+          ...stateCopy,
+          ...secondValue,
+        };
+
+        lengthActions.push({ ...stateCopy });
```
anastasiiavorobiova, line 25: Avoid repeating code
Consider pushing once after the switch statement

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of keyRemove) {
+          if (stateCopy.hasOwnProperty(key)) {
```
anastasiiavorobiova, line 30: This check is not necessary, the deletion of a property that does not exist doesn't throw an Error

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const lengthActions = [];
+  let stateCopy = { ...state };
+
+  for (const action of actions) {
+    const TYPE_MESSAGE = action.type;
+    const secondValue = action.extraData;
+    const keyRemove = action.keysToRemove;
+
```
anastasiiavorobiova, line 17: ```const ADD = 'addProperties'```
* Consider using constants instead of magic strings

anastasiiavorobiova (changes requested):

anastasiiavorobiova (approved):
Awesome!

PR #2692, opened 2023-08-23
js_stateful-clones

aamelnyk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+          delete stateCopy[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
+    }
```
Viktor-Kost, line 30: it is a good practice when you add a default behavior for switch statement. Yet, if you don't have such one - just throw an Error with text like "unexpected action"

Viktor-Kost (approved):
GJ!
Just pay attention at comment I left for you

PR #2693, opened 2023-08-23
first

ColumbusDeveloper:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,61 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const COMP_ARR = [];
```
mykhalenych, line 11: why are all your variables with uppercase?
uppercase variables must be only for static constants, for all other use camelCase

src/transformStateWithClones.js
```diff
@@ -6,8 +6,61 @@
...
+    const OBJ_FOR_WORK = {};
+
+    OBJ_FOR_WORK[OBJ_FOR_WORK_KEY] = OBJ_FOR_WORK_VALUE;
+    COMP_ARR.push(OBJ_FOR_WORK);
+  }
+
+  for (let i = 0; i < COMP_ARR.length; i++) {
+    const el = COMP_ARR[i];
+
+    if (el.hasOwnProperty('clear')) {
```
mykhalenych, line 31: change to switch case

mykhalenych (changes requested):

maxim2310 (changes requested):
Good work, but fix previous comments

src/transformStateWithClones.js
```diff
@@ -1,13 +1,61 @@
...
+    objForWork[objForWorkKey] = objForWorkValue;
+    compArr.push(objForWork);
+  }
+
+  for (let i = 0; i < compArr.length; i++) {
+    const el = compArr[i];
+    const controlKey = Object.keys(el)[0];
+
+    switch (controlKey) {
+      case 'clear':
```
anastasiiavorobiova, line 26:
```suggestion
      case CLEAR:
```
It's better to use constants instead of magic strings

src/transformStateWithClones.js
```diff
@@ -1,13 +1,61 @@
...
+        for (const key in compObj) {
+          delete compObj[key];
+        }
+        result.push(compObj);
+        break;
+      case 'addProperties':
```
anastasiiavorobiova, line 32:
```suggestion
      case ADD':
```
It's better to use constants instead of magic strings

src/transformStateWithClones.js
```diff
@@ -1,13 +1,61 @@
...
+        const objForWorkAdd = { ...result[result.length - 1] };
+
+        Object.assign(objForWorkAdd, el.addProperties);
+        result.push(objForWorkAdd);
+        break;
+      case 'removeProperties':
```
anastasiiavorobiova, line 38: The same here

src/transformStateWithClones.js
```diff
@@ -1,13 +1,61 @@
...
+        const objForWorkRem = { ...result[result.length - 1] };
+        const keysToDelete = el.removeProperties;
+
+        for (const itemToDel of keysToDelete) {
+          for (const key in objForWorkRem) {
```
anastasiiavorobiova, line 43: There is no need to use 2 cycles

anastasiiavorobiova (changes requested):

nazarmatsevych (approved):
Good work!

PR #2694, opened 2023-08-24
Solution

GIroSTarK:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateArray = [];
+  const stateClone = { ...state };
+
```
anastasiiavorobiova, line 12: ```const ADD = 'addProperties'```
Consider creating variables and using them instead of magic strings

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```
anastasiiavorobiova, line 15:
```suggestion
      case ADD:
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+        for (const key in action.extraData) {
```
anastasiiavorobiova, line 16: Consider using Object.assign instead of looping through all data for better performance

anastasiiavorobiova (changes requested):
Great work! Almost done!

PR #2695, opened 2023-08-25
Solution

flocky674:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      for (const key of property.keysToRemove) {
+        delete stateCopy[key];
+      }
+    }
+
+    if (property.type === 'clear') {
+      for (const key in stateCopy) {
+        delete stateCopy[key];
+      }
+    }
```
sTorba24, line 28: use `switch ... case` as it checking value of the same field

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const stateCopy = { ...state };
+
+  for (const property of actions) {
```
sTorba24, line 13:
```suggestion
  for (const action of actions) {
```

sTorba24: it is array of actions

sTorba24 (changes requested):
Almost good! Lets make it perfect
Also would be good move action types to variables

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const stateCopy = { ...state };
```
```suggestion
  let stateCopy = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+      case REMOVE:
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
+        break;
+
+      case CLEAR:
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
```
etojeDenys, line 31:
```suggestion
        stateCopy = {};
```

etojeDenys (approved):
good job

PR #2699, opened 2023-08-27
v1

DVasylenko18:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,59 @@
...
+      return {};
+    }
+  }
+
+  return cloneState;
+}
+
 function transformStateWithClones(state, actions) {
   // write code here
+  const actionArray = [];
```
TarasHoliuk, line 50: you store states here, not actions

src/transformStateWithClones.js
```diff
@@ -6,8 +6,59 @@
...
+
+  for (const action of actions) {
+    // apply every action to the state
+    if (actionArray.length > 0) {
```
TarasHoliuk, line 54: interesting solution :+1: 
but you can make it simpler by just mutating the `cloneState` as you actually do, and then just pushing its copy to the result array

src/transformStateWithClones.js
```diff
@@ -6,8 +6,59 @@
  *
  * @return {Object[]}
  */
+function transform(state, action) {
+  const cloneState = {};
+
+  for (const key in state) {
+    cloneState[key] = state[key];
+  }
```
TarasHoliuk, line 14:
```suggestion
  const cloneState = { ...state };
```
use spread operator to make copy

src/transformStateWithClones.js
```diff
@@ -6,8 +6,59 @@
...
+
+  switch (action.type) {
+    case 'addProperties':
+      // add all extraData props to the state
```
TarasHoliuk, line 18: it's clear from the code - remove such comments

TarasHoliuk (changes requested):

src/transformStateWithClones.js
```diff
@@ -8,6 +8,39 @@
...
+          delete cloneState[key];
+        }
+
+        break;
+      }
+
+      case 'clear': {
+        for (const key of Object.keys(cloneState)) {
+          delete cloneState[key];
+        }
```
```suggestion
        cloneState = {}
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,39 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const statesArray = [];
+  const cloneState = { ...state };
```
maxim2310, line 12: https://github.com/mate-academy/js_stateful-clones/pull/2699#discussion_r1306618210
fix this const naming

maxim2310 (changes requested):

src/transformStateWithClones.js
DVasylenko18: cloneState in my case is constant, had to change it to let

src/transformStateWithClones.js
DVasylenko18: done

maxim2310 (approved):
Good work!!!

PR #2707, opened 2023-08-30
First branch

MaxMaksym:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [state];
+
+  for (let i = 0; i < actions.length; i++) {
```
polosanya, line 12:
```suggestion
  for (const action of actions) {
```
Better use `for...of` instead of default loops for better readability

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+    const previousStateCopy = { ...states[i] };
+
+    switch (actions[i].type) {
```
polosanya, line 15: You can destructure all the needed props from current action (type, extraData, keysToRemove)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [state];
```
polosanya, line 10: It should be an empty array to avoid using shift at the end

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+
+  for (let i = 0; i < actions.length; i++) {
+    const previousStateCopy = { ...states[i] };
```
polosanya, line 13: You can create a copy outside of the loop and reassign if needed

polosanya (changes requested):
Good job! Let's improve it a bit)

PR #2710, opened 2023-08-31
stateful-clones

Masha-Cactus:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+
+    if (actions[i].type === 'removeProperties') {
+      for (const key in actions[i].keysToRemove) {
+        delete resultState[i][actions[i].keysToRemove[key]];
+      }
+    }
+
+    if (actions[i].type === 'clear') {
+      resultState[i] = {};
+    }
```
VitaliyBondarenko1982, line 31: it is good place to use switch/case instead multiply ifs

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some improvements.

VitaliyBondarenko1982 (approved):
Well done

PR #2712, opened 2023-08-31
add task solution

Liubomyr19:
**Read [the guideline](https://github.com/Liubomyr19/js_task-guideline/blob/master/README.md) before start**


VitaliyBondarenko1982 (approved):
Looks good

PR #2715, opened 2023-09-02
transformStateWithClones

olena-nimakova:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,32 @@
...
+          delete resultObject[key];
+        }
+      }
+      resultArray.push({ ...resultObject });
+    }
+
+    if (actions[i].type === 'clear') {
+      resultArray.push({});
+      resultObject = {};
+    }
```
nazarmatsevych, line 32: Good work but let's rewrite it using switch/case because we rely on a single value `actions[i].type`

src/transformStateWithClones.js
```diff
@@ -6,8 +6,32 @@
...
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  let resultObject = { ...state };
+  const resultArray = [];
+
+  for (let i = 0; i < actions.length; i++) {
```
nazarmatsevych, line 14:
```suggestion
  for (const action of actions) {
```

You are not using indexes in your code so we can use for of loop

nazarmatsevych (changes requested):

src/transformStateWithClones.js
```diff
@@ -6,8 +6,38 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  let resultObject = { ...state };
+  const resultArray = [];
+
```
anastasiiavorobiova, line 13: const ADD = 'addProperties';
Consider using constants instead of magic strings

src/transformStateWithClones.js
```diff
@@ -6,8 +6,38 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties': {
```
anastasiiavorobiova, line 16:
```suggestion
      case ADD: {
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Awesome!

PR #2717, opened 2023-09-03
Solution

Viktoriia-Vita:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let newState = { ...state };
+  const arrOfCopies = [];
+
+  for (const action of actions) {
+    let cloneState = { ...newState };
+
+    if (action.type === 'addProperties') {
```
anastasiiavorobiova, line 16: Consider using the switch here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let newState = { ...state };
+  const arrOfCopies = [];
+
```
anastasiiavorobiova, line 12: ```const ADD = 'addProperties';```

Consider using constants instead of magic strings, because it's an antipattern

anastasiiavorobiova (changes requested):
Great work! Almost done!

maxim2310 (approved):
Good work!!!

PR #2718, opened 2023-09-04
add task solution

Vovchukkul:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+
+      case CLEAR_PROP: {
+        cloneState = {};
+
+        break;
+      }
+    }
+    newArray.push(cloneState);
+    newCopy = cloneState;
+  }
```
maxim2310, line 42: you dont need to create new copy  let cloneState = { ...newCopy };
```suggestion
  for (const action of actions) {
    switch (action.type) {
      case ADD_PROP:
        Object.assign(newCopy, action.extraData);

        break;

      case REMOVE_PROP: {
        for (const key of action.keysToRemove) {
          delete newCopy[key];
        }

        break;
      }

      case CLEAR_PROP: {
        newCopy = {};

        break;
      }
    }
    newArray.push({...newCopy });
  }
```

maxim2310 (changes requested):

anastasiiavorobiova (approved):
Awesome!

PR #2721, opened 2023-09-05
Solution

mrsvolodya:


src/transformStateWithClones.js
```diff
@@ -1,13 +1,27 @@
...
+      arrayOfObjects.push({ ...newObject });
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete newObject[key];
+      }
+      arrayOfObjects.push({ ...newObject });
+    } else if (action.type === 'clear') {
+      newObject = {};
+      arrayOfObjects.push({ ...newObject });
+    }
```
anastasiiavorobiova, line 21: Consider pushing once ```arrayOfObjects.push({ ...newObject });``` after conditional statements to avoid code repeating

src/transformStateWithClones.js
```diff
@@ -1,13 +1,27 @@
...
- * @return {Object[]}
- */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arrayOfObjects = [];
+  let newObject = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      for (const key in action.extraData) {
```
anastasiiavorobiova, line 9: Consider using Object.assign instead of looping through every key in the object to improve the performance

anastasiiavorobiova (changes requested):
Great work! Added tiny comments to fix

src/transformStateWithClones.js
```diff
@@ -1,13 +1,25 @@
...
+        newObject[key] = action.extraData[key];
+      }
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete newObject[key];
+      }
+    } else if (action.type === 'clear') {
+      newObject = {};
+    }
```
etojeDenys, line 18: let's use switch case instead of multiple if else

etojeDenys (changes requested):

anastasiiavorobiova (approved):
&#128170; &#129504;

PR #2724, opened 2023-09-07
add task solution

ghost:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+        break;
+      }
+
+      case REMOVE: {
+        for (const key of action.keysToRemove) {
+          if (resultObject.hasOwnProperty(key)) {
+            delete resultObject[key];
+          }
+        }
+        resultArray.push({ ...resultObject });
```
maxim2310, line 30: move it in end of func, because you use it in eny cases, just make manipulation with result obj and in the end make a copy and push

maxim2310 (changes requested):

src/transformStateWithClones.js
ololopos: done

ololopos:
code fixed , please review

etojeDenys (approved):
awesome

PR #2726, opened 2023-09-08
Stateful clones

tetyana-shatrova:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
-  // write code here
+  const newStateArray = [];
+  const newState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          newState[key] = action.extraData[key];
+        }
```
VitaliyBodnar, line 18: it's better to use  `Object.assign(stateCopy, action.extraData);` as it will make a code shorter

VitaliyBodnar (changes requested):
Almost done!

mykhalenych (approved):

PR #2728, opened 2023-09-09
the js_stateful-clones task was solved

Dmytro-Bashlai:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClone = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateClone, action.extraData);
+        result.push({ ...stateClone });
```
lerastarynets, line 17: There is repetition of the same logic for cloning the state object. You can clone the object once at the end of each loop iteration.

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(stateClone)) {
+          delete stateClone[key];
+        }
+        break;
+    }
+    result.push(Object.assign({}, stateClone));
```
Anastasiia-Svintsova, line 31: You can use spread syntax instead. But your variant is also correct. So, it's up to you 
```suggestion
    result.push({ ...stateClone });
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClone = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateClone, action.extraData);
```
Anastasiia-Svintsova, line 16: You could apply destructuring to the action object for better code readability

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+        break;
+
+      case 'removeProperties':
```
Anastasiia-Svintsova, line 19: It would be great to move string variables to a constant to avoid errors connected with typos in the future

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+        for (const key of action.keysToRemove) {
+          delete stateClone[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(stateClone)) {
+          delete stateClone[key];
+        }
```
Anastasiia-Svintsova, line 28: It's generally considered a better practice to set an object to an empty object `{}` rather than clearing its properties individually with a loop

Anastasiia-Svintsova (changes requested):

maxim2310 (approved):
Good work!!!

PR #2731, opened 2023-09-10
done task

Dmytrosp:
[CHECKLIST](https://github.com/Dmytrosp/js_stateful-clones/blob/master/checklist.md)

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,40 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const array = [];
+  const obj = { ...state };
+
+  for (const element of actions) {
+    const { type, extraData, keysToRemove } = element;
+
+    switch (type) {
+      case 'addProperties': {
+        for (const key in extraData) {
+          obj[key] = extraData[key];
+        }
+        break;
+      }
+
+      case 'removeProperties': {
+        for (const key of keysToRemove) {
+          if (obj[key]) {
+            delete obj[key];
+          }
+        }
+        break;
+      }
+
+      case 'clear': {
+        for (const key in obj) {
+          delete obj[key];
+        }
+        break;
+      }
+    }
+    array.push({ ...obj });
+  }
+
+  return array;
 }
 
 module.exports = transformStateWithClones;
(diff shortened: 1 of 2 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const array = [];
+  const obj = { ...state };
+
```
anastasiiavorobiova, line 12: const ADD = 'addProperties';

Consider using constants instead of magic strings here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+  for (const element of actions) {
+    const { type, extraData, keysToRemove } = element;
+
+    switch (type) {
+      case 'addProperties': {
```
anastasiiavorobiova, line 17:
```suggestion
      case ADD: {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+        for (const key in extraData) {
```
anastasiiavorobiova, line 18: Consider using Object.assign instead of looping through every property for better performance

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+          obj[key] = extraData[key];
+        }
+        break;
+      }
+
+      case 'removeProperties': {
```
anastasiiavorobiova, line 24:
```suggestion
      case REMOVE: {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+        for (const key of keysToRemove) {
+          if (obj[key]) {
```
anastasiiavorobiova, line 26:
```suggestion

```
There is no need to provide an additional check here, the delete operator wouldn't throw an error if the property doesn't exist

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+            delete obj[key];
+          }
+        }
+        break;
+      }
+
+      case 'clear': {
```
anastasiiavorobiova, line 33:
```suggestion
      case CLEAR: {
```

anastasiiavorobiova (changes requested):
Great work! Added small comments to fix

nazarmatsevych (approved):

PR #2735, opened 2023-09-12
1

viktor-kniahnytskyi:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+        }
+        break;
+
+      case CLEAR:
+        cloneState = {};
+        break;
+    }
+    arrOfCopies.push(cloneState);
+    newState = cloneState;
+  }
```
maxim2310, line 37: would be a bit better, but you solution also good
```suggestion
  for (const action of actions) {
    switch (action.type) {
      case ADD:
        Object.assign(newState, action.extraData);
        break;

      case REMOVE:
        for (const key of action.keysToRemove) {
          delete newState[key];
        }
        break;

      case CLEAR:
        newState= {};
        break;
    }
    
    arrOfCopies.push({...newState });
  }
```

maxim2310 (approved):
Good work!!!

PR #2736, opened 2023-09-12
Add task solution

Shostyy:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+
+  const toModified = { ...state };
+
+  for (const action of actions) {
+    change(toModified, action);
+    allStatement.push(Object.assign({}, toModified));
+  }
+
+  return allStatement;
+}
```
Viktor-Kost, line 20: this is redundant
```suggestion
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const allStatement = [];
+
+  const toModified = { ...state };
```
Viktor-Kost, line 12:
```suggestion
  const coppiedState = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+        delete state[key];
+      }
+      break;
+
+    case 'clear':
+      for (const key in state) {
+        delete state[key];
+      }
+      break;
+  }
```
Viktor-Kost, line 39: 1. right here you can just invoke a loop
2. at the end of the loop you can add changes
3. after loop you can just return an array
```suggestion
  for (const action of actions) {
    switch (action.type) {
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;
  
      case 'removeProperties':
        for (const key of action.keysToRemove) {
          delete state[key];
        }
        break;
  
      case 'clear':
        for (const key in state) {
          delete state[key];
        }
        break;
    }
    
    allStatement.push({ ...coppiedState });
  }
  
  return allStatement;
```

Viktor-Kost (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const allStatement = [];
+
+  const coppiedState = { ...state };
```
etojeDenys, line 12:
```suggestion
  let coppiedState = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        for (const key of action.keysToRemove) {
+          delete coppiedState[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in coppiedState) {
+          delete coppiedState[key];
+        }
+        break;
```
```suggestion
      case 'clear':
        coppiedState = {}
        break;
```

etojeDenys (approved):
lgtm

PR #2747, opened 2023-09-15
Solution

OlegPopovych:


sintax1s (approved):
Excellent job!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+      case 'removeProperties':
+        const removePropertiesObj = { ...temporaryPreviousObject };
+
+        for (const key of item.keysToRemove) {
+          delete removePropertiesObj[key];
+        }
+
+        temporaryPreviousObject = removePropertiesObj;
+
+        results.push(removePropertiesObj);
```
StasSokolov, line 36: you can just delete from `temporaryPreviousObject`

StasSokolov (changes requested):
Good job!

1. Check please this point in requirements
Снимок экрана 2023-09-19 в 10 04 34

src/transformStateWithClones.js
OlegPopovych: Thanks, I fixed it.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const results = [];
+  let temporaryPreviousObject = { ...state };
+
+  for (const item of actions) {
+    const type = item.type;
```
```suggestion
  for (const { type... } of actions) {
```
You can destructure all the needed props directly

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+
+    switch (type) {
+      case 'addProperties':
```
polosanya, line 17: Not critical but would be great to use variables instead of the hardcoded values

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+        const addPropertiesObj = {
+          ...temporaryPreviousObject,
+          ...item.extraData,
+        };
+
+        temporaryPreviousObject = addPropertiesObj;
```
polosanya, line 23: Consider using Object.assign instead

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+
+        results.push({ ...addPropertiesObj });
+        break;
+      case 'removeProperties':
+        for (const key of item.keysToRemove) {
+          delete temporaryPreviousObject[key];
+        }
+
+        results.push({ ...temporaryPreviousObject });
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+        break;
+      case 'clear':
+        temporaryPreviousObject = {};
+
+        results.push({});
+        break;
+      default:
+        continue;
```
polosanya, line 40: It's better to throw some error instead

polosanya (changes requested):
Keep up the good work! Let's improve the code a bit)

src/transformStateWithClones.js
OlegPopovych: Yeah, I forgot to use all JSs features

src/transformStateWithClones.js
```diff
@@ -1,13 +1,43 @@
...
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const results = [];
+  let temporaryObject = { ...state };
+
+  for (const { type, extraData, keysToRemove } of actions) {
+    switch (type) {
+      case ACTION_ADD_PROPERTIES:
+        temporaryObject = Object.assign(temporaryObject, extraData);
```
StasSokolov, line 20:
```suggestion
        Object.assign(temporaryObject, extraData);
```

there is no need to set result of `Object.assign` to the `temporaryObject`

StasSokolov (changes requested):
Great job, only one comment left to fix!

src/transformStateWithClones.js
OlegPopovych: Yes, there was a redundant piece of code

StasSokolov (approved):
Awesome job!

PR #2748, opened 2023-09-15
Solution

mrogozhinskaya:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+  const stateCopy = Object.assign({}, state);
+  const arrayResult = [];
+
+  for (const item of actions) {
+    switch (item.type) {
+      case 'addProperties': {
+        for (const key in item.extraData) {
+          stateCopy[key] = item.extraData[key];
+        }
+        arrayResult.push({ ...stateCopy });
```
sTorba24, line 19: move this push after switch case actions to not repeat. You're doing same action for all cases, so it could be moved outside

sTorba24 (approved):

PR #2749, opened 2023-09-16
Solution

SerhiyKompaniets:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+  const copyState = Object.assign({}, state);
+
+  for (const action of actions) {
+    transform(copyState, action);
+  }
+
+  function transform(newState, action) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(copyState, action.extraData);
```
sTorba24, line 20: it's not good idea to use `dirty` function here
`dirty` because it's using variable outside a function for no reason

src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const copyState = Object.assign({}, state);
+
+  for (const action of actions) {
+    transform(copyState, action);
+  }
+
+  function transform(newState, action) {
```
sTorba24, line 17: also you are not using `newState` variable
What sense of this function?

sTorba24 (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+  const copyState = Object.assign({}, state);
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(copyState, action.extraData);
+
+        const addState = { ...copyState };
+
+        result.push(addState);
```
etojeDenys, line 20: let's  remove it from cases, and push the copyState at the end of the loop

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #2750, opened 2023-09-16
Solution

yslyshchenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const stateClone = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties': {
+        Object.assign(stateClone, action.extraData);
```
lerastarynets, line 16: Consider renaming 'extraData' to a more descriptive name that reflects its purpose.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+      case 'removeProperties': {
+        for (const key of action.keysToRemove) {
+          delete stateClone[key];
+        }
+        break;
+      }
+
+      case 'clear': {
+        for (const key in stateClone) {
```
lerastarynets, line 28: Instead of using a loop to delete all properties, you can simply set 'stateClone' to an empty object. This will improve performance.

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

src/transformStateWithClones.js
yslyshchenko: How can I rename it if it's a key of the object? Do you mean creating a new const?

lerastarynets (approved):
You've got this, keep pushing forward!

src/transformStateWithClones.js
lerastarynets: @yslyshchenko ah I see, never mind then)

PR #2752, opened 2023-09-17
Refactor transformStateWithClones function to avoid object mutations …

Azazel-hell:
…and improve code style

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,40 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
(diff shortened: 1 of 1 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+      case 'addProperties':
+        currentState = {
+          ...currentState,
+          ...extraData,
+        };
+        break;
+
+      case 'removeProperties':
+        keysToRemove.forEach(key => {
+          if (Object.hasOwnProperty.call(currentState, key)) {
```
Andriy-Shymkiv, line 26: try it like this, hope it will work:
```suggestion
          if (Object.hasOwnProperty(key)) {
```

Andriy-Shymkiv (approved):
nice work, just try to simplify object method. approved in advance

PR #2754, opened 2023-09-17
add task solution

VasilPan:


src/transformStateWithClones.js
```diff
@@ -1,13 +1,34 @@
...
+    }
+
+    if (actions[i].type === 'clear') {
+      for (const element in newState) {
+        delete newState[element];
+      }
+      masiv.push({ ...newState });
+    }
+  }
+
```
etojeDenys, line 30: let's rewrite it using switch case

src/transformStateWithClones.js
```diff
@@ -1,13 +1,34 @@
...
-/**
- * @param {Object} state
- * @param {Object[]} actions
- *
- * @return {Object[]}
- */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const masiv = [];
```
etojeDenys, line 5:
```suggestion
  const arr = [];
```

src/transformStateWithClones.js
```diff
@@ -1,13 +1,34 @@
...
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
+      for (const add in actions[i].extraData) {
+        newState[add] = actions[i].extraData[add];
+      }
+      masiv.push({ ...newState });
```
etojeDenys, line 12: we can move this out of the conditions and push only one time at the end of the loop

etojeDenys (changes requested):

VitaliyBodnar (approved):

PR #2755, opened 2023-09-17
task solution

RomanO3GIT:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copiedStateObject = { ...state };
+  const newState = [];
+
+  for (const action of actions) {
+    switch (action.type) {
```
volodymyr-soltys97, line 14: Use a destructuring for action
```suggestion
  for (const action of actions) {
    const {} = action;
   
    switch (type) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      case 'addProperties':
+
+        Object.assign(copiedStateObject, action.extraData);
+        break;
+
+      case 'removeProperties':
```
volodymyr-soltys97, line 20: Consider will be better if you create a variable for all magic words and them everywhere, it's a good practice

volodymyr-soltys97 (changes requested):
Great work 👍 
Let's improve your code

PR #2757, opened 2023-09-18
solve transform state with clone task

OlhaKoziuk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
+        delete state[key];
+      }
+      break;
+
+    default:
+      const errorObject = { error: 'Unknown action type' };
+
+      result.push(errorObject);
+      break;
+  }
```
A1eeex, line 44: will be better better add variable.

```suggestion
  const ADD_PROPERTIES = 'addProperties';
  const REMOWE_PROPERTIES = 'removeProperties';
  const CLEAR = 'clear';
  const ERROR_MESSAGE = 'Unknown action type';
  
  const copyState = { ...state };
  const copyActions = [ ...actions ];
  const result = [];

  for (const action of copyActions) {
    findActions(copyState, action, result);
  }

  return result;
}

function findActions(state, action, result) {
  switch (action.type) {
    case ADD_PROPERTIES :
      Object.assign(state, action.extraData);
      break;

    case REMOWE_PROPERTIES :
      for (const key of action.keysToRemove) {
        delete state[key];
      }
      break;

    case CLEAR :
      for (const key of Object.keys(state)) {
        delete state[key];
      }
      break;

    default:
      const errorObject = { error: ERROR_MESSAGE };

      result.push(errorObject);
      break;
  }
```

A1eeex (approved):
Good job!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
+  const copyActions = [ ...actions ];
+  const result = [];
+
+  for (const action of copyActions) {
+    findActions(copyState, action, result);
```
StasSokolov, line 15: 🚀

StasSokolov (approved):
Awesome job🚀

PR #2761, opened 2023-09-18
stateful clones

tetyana-pol:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+        }
+        resultObject.push({ ...stateCopy });
+        break;
+      }
+
+      case CLEAR: {
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        resultObject.push({});
```
Drach-Illya, line 37: I would move **resultObject.push({ ...stateCopy });** outside the switch. You don't need to declare it separately in each type of action. If the actions are done correctly, then just having **resultObject.push({ ...stateCopy });** outside of switch would have the same effect with it only being written once instead of thrice.

Drach-Illya (approved):
Very good code. I found only one thing that can be improved, but a great job nonetheless.

PR #2765, opened 2023-09-18
added task solution

YevhenKharko:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,39 @@
...
+        break;
+      }
+
+      case 'clear': {
+        newState = {};
+        break;
+      }
+
+      default: {
+        throw new Error('Something is wrong...');
```
adioskv, line 35: I think we should write there more specific error, for example 'Wrong action type' etc.

adioskv (approved):
Good job!

src/transformStateWithClones.js
YevhenKharko: Ok! Thank you!

src/transformStateWithClones.js
```diff
@@ -6,8 +6,39 @@
...
  */
-function transformStateWithClones(state, actions) {
-  // write code here
-}
+const transformStateWithClones = (state, actions) => {
+  const results = [Object.assign({}, state)];
+
+  for (const action of actions) {
+    const type = action.type;
+    let newState = Object.assign({}, results[results.length - 1]);
```
StasSokolov, line 14: can be more simplify
```suggestion
    let newState = { ...results };
```

StasSokolov (approved):
LGTM!

PR #2766, opened 2023-09-18
Develop

adioskv:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+        break;
+
+      case 'clear':
+        stateCopy = {};
+        break;
+
+      default:
+        throw new Error('wrong action type');
+    }
+    currentState = stateCopy || currentState;
```
tetyana-pol, line 38: you use unexpected way. It is better to use _Object.assign_ or _spread_ operator

tetyana-pol (approved):

src/transformStateWithClones.js
adioskv: thank you !

StasSokolov, line 38: Снимок экрана 2023-09-19 в 11 09 35
Check this point please

StasSokolov (changes requested):
Good job, let's improve your code a bit more!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const logs = [];
+  let currentState = state;
+
```
superpooperxxx, line 12: I suggest to create copy of state once at the begging and then mutate the copy
`const currentState = { ...state }`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+  for (const action of actions) {
+    let stateCopy = { ...currentState };
+
+    switch (action.type) {
+      case 'addProperties':
+        if ('extraData' in action) {
+          Object.assign(stateCopy, action.extraData);
+        }
```
superpooperxxx, line 20: Not sure you need this if statement, we always have `extraData` for this type of action

superpooperxxx (changes requested):
Hey, good job🔥 My first comment is the most important one, it will make your solution more optimised.

src/transformStateWithClones.js
adioskv: Do i need just create stateCopy before the loop and mutate it at the beginning ?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const logs = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 14: You can destructure the action to get all the needed props of it

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      case 'addProperties':
+        Object.assign(currentState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const remove of action.keysToRemove) {
```
```suggestion
        for (const key of action.keysToRemove) {
```
This name is more accurate

polosanya (approved):
Nice! I'm approving it, but please pay attention to the comments)

PR #2767, opened 2023-09-18
Solution

vlevik:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultsOfEachAction = [];
+  const stateCopy = { ...state };
+
+  for (const action of actions) {
+    switch (true) {
+      case action.type === 'addProperties':
```
mpohorenyi, line 15:
```suggestion
    switch (action.type) {
      case 'addProperties':
```
You can write the key of an object in the condition in `switch`, and check the value of the key in `case`.

mpohorenyi (approved):
Great Job! 🔥
Please note the comments and make changes.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultsOfEachAction = [];
+  const stateCopy = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 14: You can destructure all the needed properties of action (extraData, keysToRemove, type)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+      case 'addProperties':
+        for (const key in action.extraData) {
+          stateCopy[key] = action.extraData[key];
```
polosanya, line 17: Better use Object.assign instead of the loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
+
+      default :
+        throw new Error('This is not a command!');
```
polosanya, line 34: 👍

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
+        break;
+
+      case 'clear':
```
polosanya, line 27: Not critical, but it's better to use variables instead of the hardcoded values

polosanya (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultsOfEachAction = [];
+  const stateCopy = { ...state };
+  const commandAddProperties = 'addProperties';
+  const commandRemoveProperties = 'removeProperties';
+  const commandClear = 'clear';
```
```suggestion
  const COMMAND_CLEAR = 'clear';
```
Same with other static variables

polosanya (approved):
Well done!

PR #2769, opened 2023-09-18
Completed the task

Maksim114488:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const newArray = [];
+
+  for (let i = 0; i < actions.length; i++) {
```
yulabakun, line 13: You can also use 'for of' loop here. There's no need for indexes.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+          delete newState[actions[i].keysToRemove[y]];
+        };
+        break;
+
+      case 'clear':
+        for (const key in newState) {
+          delete newState[key];
+        }
+        break;
+    }
```
yulabakun, line 32: Also could you please add to switch case 'default'
[CHECKLIST] 3. [CODE STYLE]: switch/case should always have default case for error handling.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+  for (let i = 0; i < actions.length; i++) {
+    switch (actions[i].type) {
+      case 'addProperties':
+        for (const key in actions[i].extraData) {
+          newState[key] = actions[i].extraData[key];
+        };
+        break;
+
+      case 'removeProperties':
+        for (let y = 0; y < actions[i].keysToRemove.length; y++) {
```
yulabakun, line 22: Same here, I think 'for of' loop with make your code look cleaner;

yulabakun (changes requested):
You've done a good job!
Just add default case to your switch as it is in the checklist.
All other suggestions are just code style preferences.

yulabakun (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const newArray = [];
+
+  for (const key of actions) {
+    switch (key.type) {
+      case 'addProperties':
```
StasSokolov, line 15: non critical but will be good move type to constants

StasSokolov (approved):
Good job, i'm approving it!

PR #2770, opened 2023-09-18
first solution

Clover247:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newStates = [Object.assign({}, state)];
+
+  for (const action of actions) {
+    const type = action.type;
```
polosanya, line 13: You can destructure all the needed props of action

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newStates = [Object.assign({}, state)];
```
polosanya, line 10: Better make it an empty array so you don't need to use slice later

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+
+  for (const action of actions) {
+    const type = action.type;
+    let newState = Object.assign({}, newStates[newStates.length - 1]);
```
polosanya, line 14: Just create a copy of state and move it out of the loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+      }
+
+      case 'removeProperties': {
+        for (const key of action.keysToRemove) {
+          delete newState[key];
+        }
+        break;
+      }
+
+      case 'clear': {
```
polosanya, line 29: Not critical, but would be great to use variables instead of the hardcoded values

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        newState = {};
+        break;
+      }
```

polosanya (changes requested):
Nice! Here are some moments that would be great to improve

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newStates = [state];
```
StasSokolov, line 10: maybe you mean `copyState`? `{ ...state }`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+
+      default: {
+        throw new Error(`Невідомий тип дії: ${type}`);
+      }
+    }
+
+    newStates.push(newState);
+  }
+
+  return newStates.slice(1);
```
StasSokolov, line 45: let's do without slice

StasSokolov (changes requested):
Great job, only few comments left to fix!

StasSokolov (commented):
LGTM!

StasSokolov (approved):
LGTM!

PR #2774, opened 2023-09-19
add task solution stateful-clones

denzzlle171:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
-  // write code here
+  const result = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    let temporalState = { ...currentState };
+
+    if (action.type === 'addProperties' && action.extraData) {
+      for (const key in action.extraData) {
+        temporalState[key] = action.extraData[key];
```
shchipina, line 18: I don't understand why you need to use  logical and (&&) here
` if (action.type === 'addProperties' && action.extraData) {`

shchipina: and also maybe I wrong but in check list, need to use switch

shchipina (changes requested):
Good gob!  just a few comments

src/transformStateWithClones.js
denzzlle171: Yeah, you're right. I didn't look at the checklist. Thanks.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+
+      case 'removeProperties':
+        if (action.keysToRemove) {
+          for (const key of action.keysToRemove) {
+            delete temporalState[key];
+          }
+        }
+        break;
+
+      case 'clear':
```
StasSokolov, line 33: non critical but will be good move `type` into constants

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
-  // write code here
+  const result = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    let temporalState = { ...currentState };
+
+    switch (action.type) {
+      case 'addProperties':
+        if (action.extraData) {
```
StasSokolov, line 18: redundant `if` in this case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    let temporalState = { ...currentState };
```
StasSokolov, line 14: maybe better use `currentState` variable for this? i think this is redundant variable

StasSokolov (changes requested):
Almost done! Let's fix only a few comments!

src/transformStateWithClones.js
denzzlle171: thank, I fixed it all

StasSokolov (approved):
Awesome job🔥

PR #2775, opened 2023-09-19
add task solution

Stim1:


BetterSol (approved):
Looks great! As a suggestion, You can push only needed values without the initial state to `clonedStates` and you won't need  `slice` when returning it.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+      case 'addProperties':
+        Object.assign(newState, action.extraData);
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete newState[key];
+        }
+        break;
+      case 'clear':
+        for (const key in newState) {
```
polosanya, line 25: The 'if' condition inside the 'for...in' loop is not necessary because the loop already checks if the key exists in the object.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clonedStates = [{ ...state }];
+
+  for (const action of actions) {
+    const newState = { ...clonedStates[clonedStates.length - 1] };
+
+    switch (action.type) {
```
polosanya, line 15: You can destructure all the needed props from action

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clonedStates = [{ ...state }];
```
polosanya, line 10: Just make it an empty array so you don't need to use slice at the end

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+
+  for (const action of actions) {
+    const newState = { ...clonedStates[clonedStates.length - 1] };
```
polosanya, line 13: Better create a copy of the state and move it out of the loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(newState, action.extraData);
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete newState[key];
+        }
+        break;
+      case 'clear':
```
polosanya, line 24: Not critical, but would be great to use variables instead of the hardcoded values

polosanya (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const TYPE_VALUE_CLEAR = 'clear';
+  const TYPE_VALUE_REMOVE_PROPERTIES = 'removeProperties';
```
vasylmiskiv, line 11: you tried to do like in redux but these variables are not necessary in this case

vasylmiskiv (approved):

PR #2776, opened 2023-09-19
solution transformStateWithClones.js task

ivaZaiets:


Stim1 (approved):
Nice solution!

src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const arr = [];
+  let copyOfState = { ...state };
+
+  for (const values of actions) {
+    if (values.type === 'addProperties') {
```
vasylmiskiv, line 15: Would be great to use `switch case` instead of `if`

src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const arr = [];
+  let copyOfState = { ...state };
```
vasylmiskiv, line 12: Also you could use Object assign with all of key values as a second param to create a copy of the state. It's not mistake, just another solve

src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
...
+
+  for (const values of actions) {
+    if (values.type === 'addProperties') {
+      Object.assign(copyOfState, values.extraData);
+      arr.push({ ...copyOfState });
+    }
+
+    if (values.type === 'removeProperties') {
+      for (const keysToRemoveValues of values.keysToRemove) {
```
vasylmiskiv, line 21: Would be better to use something like so
```
for (const key of values.keysToRemove) {
  delete copyOfState[key];
}
```
So in this case we don't need to compare we just delete all of the required keys

src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
...
+        for (const copyOfStateValues in copyOfState) {
+          if (keysToRemoveValues === copyOfStateValues) {
+            delete copyOfState[copyOfStateValues];
+          }
+        }
+      }
+      arr.push({ ...copyOfState });
```
vasylmiskiv, line 28: Actually you don't need to push the object in array every if condition. You can do it only once after all the conditions

src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const arr = [];
+  let copyOfState = { ...state };
+
+  for (const values of actions) {
```
vasylmiskiv, line 14: Would be better to rename `values` to `value`

vasylmiskiv (changes requested):

StasSokolov (changes requested):
Good job, left to fix only one comment from me!

1. check this point in the checklist
Снимок экрана 2023-09-25 в 10 04 20

StasSokolov (approved):
Great job, i', approving it!

PR #2777, opened 2023-09-19
Solution

dsfreedom:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+      case 'addProperties':
+        if (action.extraData) {
+          Object.assign(stateCopy, action.extraData);
+        }
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
+        break;
```
Ayogoharo, line 24: you can do like so, it could be easier. But anyway, good job.
```
 case 'removeProperties':
        action.keysToRemove.forEach(key => delete stateCopy[key]);
        break;
```

Ayogoharo (commented):
Great job!

Ayogoharo (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const stateCopy = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 14: You can destructure all the needed props from the action

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
polosanya, line 10: It's recommended to specify what exactly the result is. The name should be descriptive

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+          delete stateCopy[key];
+        }
+        break;
+      case 'clear':
+        for (const keys in stateCopy) {
+          delete stateCopy[keys];
+        }
+        break;
+      default:
+        throw new Error('wrong input');
```
polosanya, line 31: Try to provide more descriptive errors

polosanya (changes requested):
Great solution! Here are some moments that would be nice to improve

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+  const currentState = { ...state };
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
+        if (extraData) {
+          Object.assign(currentState, extraData);
+        }
```
StasSokolov, line 20: i think this condition is redundant in this case

StasSokolov (approved):
Great job, i'm approving it!

PR #2778, opened 2023-09-19
task solved

BetterSol:
Hey, check my stateful-clones task, please :). I'm sure though there are ways to make it better, would be thankful to get a useful hint)

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let stateClone = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      stateClone = Object.assign(stateClone, action.extraData);
+      result.push({ ...stateClone });
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete stateClone[key];
+      }
+      result.push({ ...stateClone });
+    }
+
+    if (action.type === 'clear') {
+      stateClone = {};
+      result.push({ ...stateClone });
+    }
+  }
+
+  return result;
 }
 
 module.exports = transformStateWithClones;


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let stateClone = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
DenysShurubura, line 14: Good job, but better use switch/case as u can read in checklist!
Good luck

DenysShurubura (changes requested):

src/transformStateWithClones.js
BetterSol: thank you! looked more understandable for me though :)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
polosanya, line 10: Try to provide more descriptive names, let's specify what exactly the result is

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+  let stateClone = { ...state };
```
polosanya, line 11: Consider using const instead of let without extra need

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 14: You can destructure all the needed props from action

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+      case 'addProperties':
+        stateClone = Object.assign(stateClone, action.extraData);
```
polosanya, line 16:
```suggestion
        Object.assign(stateClone, action.extraData);
```
Reassignment is not necessary

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateClone[key];
+        }
+        break;
+
+      case 'clear':
```
polosanya, line 25: It's not necessary but would be great to use variables instead of the hardcoded values

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+        stateClone = {};
+        break;
```
polosanya, line 27: image

polosanya (changes requested):
Nice! Let's improve the code a bit)

src/transformStateWithClones.js
BetterSol: +

src/transformStateWithClones.js
BetterSol: ++

src/transformStateWithClones.js
BetterSol: I'm creating `stateClone` intentionally as a copy of the initial `state` so I can modify it in the code below.

src/transformStateWithClones.js
BetterSol: ++

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const transformedState = [];
+  const stateClone = { ...state };
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+    const ADD_PROPERTIES = 'addProperties';
+    const REMOVE_PROPERTIES = 'removeProperties';
+    const CLEAR = 'clear';
```
polosanya, line 17: You can move them outside of the loop, or even function

polosanya (approved):
Well done! 🚀

PR #2780, opened 2023-09-19
task solution

dolphinnq:


YePolishchuk (approved):
Good job!!!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateModified = [];
+  const stateObject = { ...state };
+
+  for (const obj of actions) {
+    for (const key in obj) {
```
polosanya, line 14: Nested loop is redundant here, also it's a bad practice for performance.
Consider using only one here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+      if (obj[key] === 'addProperties') {
```
polosanya, line 15: image

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+        Object.assign(stateObject, obj.extraData);
+
+        const addedObject = { ...stateObject };
+
+        stateModified.push(addedObject);
```
```suggestion
        stateModified.push({ ...stateObject });
```
There is no need to create an extra variable.
Also, pay attention to the checklist
image

polosanya (changes requested):
Good, let's improve your code a bit)
Pay attention to the checklist

StasSokolov (changes requested):
Good job, let's improve your code a bit more!

1. Add please default case with error for switch
Снимок экрана 2023-10-02 в 10 18 46

StasSokolov (approved):
Great job, i'm approving it🔥

PR #2782, opened 2023-09-19
transform with clones

Ayogoharo:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const log = [];
+
+  for (const action of actions) {
+    switch (action.type) {
```
Kuper18, line 14: you can destructure keys from object

Kuper18 (approved):
Good job!

src/transformStateWithClones.js
polosanya: ➕ , you can destructure all the needed properties

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+        Object.keys(stateCopy).forEach(key => delete stateCopy[key]);
+        break;
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => delete stateCopy[key]);
+        break;
+      default:
+        throw new Error('something is wrong');
```
polosanya, line 25: Would be great to provide more descriptive error text

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
-  // write code here
+  const stateCopy = { ...state };
+  const log = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
+        Object.keys(stateCopy).forEach(key => delete stateCopy[key]);
+        break;
+      case 'addProperties':
```
polosanya, line 18: Not critical, but better to use variables instead of hardcoded values

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const log = [];
+  const clear = 'clear';
```
polosanya, line 12:
```suggestion
  const CLEAR = 'clear';
```
Better write static variables in upper case

polosanya (approved):
👍

PR #2791, opened 2023-09-20
add task solution

annabasenyuk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultObject = [];
+  const stateClone = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+      case 'addProperties':
+        Object.assign(stateClone, action.extraData);
+        break;
+
+      case 'removeProperties':
```
polosanya, line 19: Not critical, but it would be great to use variables instead of the hardcoded values

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+
+      case 'clear':
+        for (const keyClone in stateClone) {
+          delete stateClone[keyClone];
+        }
+        break;
+
+      default:
+        resultObject.push(stateClone);
```
polosanya, line 32: It think it's better to throw some error instead

polosanya (changes requested):

superpooperxxx (approved):
Good job🔥 The only things I am concerned about is `resultObject` name.

PR #2792, opened 2023-09-20
clones task

mifyy:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copiedState = Object.assign({}, state);
+  const stateHistory = [];
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
```
polosanya, line 17: Not critical, but would be great to use variables instead of the hardcoded values

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        break;
+
+      case 'clear':
+        for (const key in copiedState) {
+          delete copiedState[key];
+        }
+        break;
+
+      default:
+        break;
```
polosanya, line 34: It is better to throw some error on default case

polosanya (approved):
👍

PR #2794, opened 2023-09-21
Add task solution

Sireus:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+      case 'clear': {
+        for (const key of Object.keys(stateClone)) {
+          delete stateClone[key];
+        }
+
+        break;
+      }
+
+      default: {
+        result.push(stateClone);
```
VitaliyBodnar, line 38: you don't need to push stateClone in a default case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+
+      case 'removeProperties': {
+        for (const key of action.keysToRemove) {
+          delete stateClone[key];
+        }
+
+        break;
+      }
+
+      case 'clear': {
```
VitaliyBodnar, line 29: it's better to move 'addProperties', 'removeProperties', 'clear' to the constants as it will make a code cleaner

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):

PR #2797, opened 2023-09-22
add task solution

ihor-pavlov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
+        addNewState(stateCopy);
+        break;
+      case 'clear':
+        for (const i in stateCopy) {
+          delete stateCopy[i];
+        }
+        addNewState(stateCopy);
```
VitaliyBodnar, line 29: addNewState(stateCopy) is duplicated, so make sense to move it after switch case block

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        break;
+      default:
+        resultArray.push('You have an error');
```
VitaliyBodnar, line 32: don't need to push an error 
```suggestion
        resultArray.push('You have an error');
```

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):

PR #2800, opened 2023-09-23
add function transformStateWithClones

Dizelkos:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const result = [ ];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties' :
+        Object.assign(stateCopy, action.extraData);
+        result.push({ ...stateCopy });
```
Anastasiia-Svintsova, line 17: You repeat this push action for each case. Just move it after the switch/case construction and call it only once

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        break;
+
+      case 'clear' :
+        for (const key of Object.keys(stateCopy)) {
+          delete stateCopy[key];
+        }
+        result.push({ ...stateCopy });
+        break;
+    }
```
Anastasiia-Svintsova, line 33: Follow the checklist
image

Anastasiia-Svintsova (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+
+      case 'clear' :
+        for (const key of Object.keys(stateCopy)) {
+          delete stateCopy[key];
+        }
+        break;
+
+      default:
+        result.push({});
+        break;
```
VitaliyBodnar, line 33: don't need to push empty object to result 
```suggestion
      default:
        break;
```

VitaliyBodnar (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+      case 'removeProperties' :
+        for (const value of keysToRemove) {
+          delete stateCopy[value];
+        }
+        break;
+
+      case 'clear' :
+        for (const key of Object.keys(stateCopy)) {
+          delete stateCopy[key];
+        }
```
Andriy-Shymkiv, line 28: you can do like this, imho it's better to understand:
```suggestion
        for (const key in stateCopy) {
          delete stateCopy[key];
        }
```

Andriy-Shymkiv (approved):
good job, just fix the loop for `clear` action. approved in advance

PR #2802, opened 2023-09-25
solution

Illia0308:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+          copyState[keys] = el.extraData[keys];
+        }
+        resultArray.push(Object.assign({}, copyState));
+        break;
+
+      case 'removeProperties':
+        for (const keys of el.keysToRemove) {
+          delete copyState[keys];
+        }
+        resultArray.push(Object.assign({}, copyState));
```
VitaliyBodnar, line 26: 'resultArray.push(Object.assign({}, copyState));' is duplicated, so make sense to move it after the switch case block

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        break;
 
+      case 'clear':
+        for (const keys in copyState) {
+          delete copyState[keys];
+        }
+        resultArray.push(Object.assign({}, copyState));
+        break;
+    }
```
VitaliyBodnar, line 35: don't forged for adding default case
`default:`
`  break;`

VitaliyBodnar (changes requested):

PR #2808, opened 2023-09-27
js_task_transform_state

DianaFakhim:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const stateCopy = Object.assign({}, state);
+  const arrayOfOriginals = [];
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
TarasHoliuk, line 15: the perfect situation to use `switch` block - do it

src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
...
+        delete stateCopy[actionKey];
+      }
+      arrayOfOriginals.push(Object.assign({}, stateCopy));
+    }
+
+    if (action.type === 'clear') {
+      for (const key in stateCopy) {
+        delete stateCopy[key];
+      }
+      arrayOfOriginals.push(Object.assign({}, stateCopy));
```
TarasHoliuk, line 35: you are doing the same in every `if`- instead do it after all blocks (after the switch block), at the end of the loop body

TarasHoliuk (changes requested):

src/transformStateWithClones.js
DianaFakhim: Thank you for your suggestion! please find new commit using 'switch'

src/transformStateWithClones.js
DianaFakhim: please review new commit, thank you!

src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const stateCopy = Object.assign({}, state);
```
```suggestion
  const stateCopy = { ...state };
```

sTorba24: can be so

sTorba24 (approved):

PR #2821, opened 2023-10-02
Solution

yuliiababych:


plodnikate (changes requested):
Better to use switch case here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const finalState = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    switch (true) {
```
anastasiia-tilikina, line 14: using a switch with a boolean condition is not a good practice. You can set action.type as a condition.

anastasiia-tilikina (changes requested):
Let's add some more improvements.

anastasiia-tilikina (changes requested):
.

nazarmatsevych (approved):
👍

PR #2822, opened 2023-10-02
task solution

Anton201220:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      const clonedState = { ...statesArr[statesArr.length - 1] };
+
+      for (const key of action.keysToRemove) {
+        delete clonedState[key];
+      }
+      statesArr.push(clonedState);
+    } else if (action.type === 'clear') {
+      statesArr.push({});
+    }
+  }
```
etojeDenys, line 31: let's use switch case instead of multiple if else

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+  statesArr.shift();
```
etojeDenys, line 32: looks redundant

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      }
+      statesArr.push(clonedState);
+    } else if (action.type === 'removeProperties') {
+      const clonedState = { ...statesArr[statesArr.length - 1] };
+
+      for (const key of action.keysToRemove) {
+        delete clonedState[key];
+      }
+      statesArr.push(clonedState);
+    } else if (action.type === 'clear') {
```
etojeDenys, line 28: we can push only one time at the end of the loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const initialState = { ...state };
+  const statesArr = [initialState];
```
etojeDenys, line 11: should be empty by default

etojeDenys (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const statesArr = [];
+  let clonedState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```
volodymyr-soltys97, line 14: Use a destruction for action
```suggestion
  for (const action of actions) {
    const { type, extraData, keysToRemove} = action;
    
    switch (type) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+      case 'addProperties':
```
volodymyr-soltys97, line 15: Consider will be better if you create a variables for these words and use them everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

etojeDenys (approved):
lgtm

PR #2828, opened 2023-10-03
implemented task

ErnestMk1:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClone = { ...state };
+  const stateVersions = [];
+
+  for (const object of actions) {
+    if (object.type === 'addProperties') {
```
volodymyr-soltys97, line 14: Use a destruction for `object` and try to use a switch instead of if/else if

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClone = { ...state };
```
```suggestion
  let stateClone = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        for (const toRemove of keysToRemove) {
+          delete stateClone[toRemove];
+        }
+
+        stateVersions.push({ ...stateClone });
+        break;
+      case 'clear':
+        for (const key in stateClone) {
+          delete stateClone[key];
+        }
```
etojeDenys, line 32:
```suggestion
      case 'clear':
        stateClone = {};
```

etojeDenys, line 32: we can push only one time at the end of the loop

etojeDenys (changes requested):

VitaliyBodnar (approved):

PR #2830, opened 2023-10-03
add task solution Stateful clones

kate0427:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,38 @@
...
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clones = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    const copyOfState = { ...currentState };
```
volodymyr-soltys97, line 15: No need to make a copy state here> remove it

src/transformStateWithClones.js
```diff
@@ -6,8 +6,38 @@
...
+
+    switch (action.type) {
```
volodymyr-soltys97, line 17: You can use a destruction for `action`
```suggestion
    const { type, .......} = action;
 
    switch (type) {
```

src/transformStateWithClones.js
```diff
@@ -6,8 +6,38 @@
...
+      case 'addProperties':
+        Object.assign(copyOfState, action.extraData);
+        break;
+
+      case 'removeProperties':
```
volodymyr-soltys97, line 22: Consider will be better if you create a variables for these words and use them everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/transformStateWithClones.js
```diff
@@ -6,8 +6,38 @@
...
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clones = [];
+  const currentState = { ...state };
+
+  for (const action of actions) {
+    const { type } = action;
+    const { extraData } = action;
+    const { keysToRemove } = action;
```
etojeDenys, line 17:
```suggestion
    const { type, extraData, keysToRemove} = action;
```

src/transformStateWithClones.js
```diff
@@ -6,8 +6,38 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clones = [];
+  const currentState = { ...state };
```
etojeDenys, line 12:
```suggestion
  let currentState = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -6,8 +6,38 @@
...
+        for (const key of keysToRemove) {
+          delete currentState[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(currentState)) {
+          delete currentState[key];
+        }
+        break;
```
etojeDenys, line 34:
```suggestion
        currentState = {};
        break;
```

etojeDenys (changes requested):

etojeDenys (approved):
looks good

PR #2834, opened 2023-10-05
solution

Daniilart01:


src/transformStateWithClones.js
```diff
@@ -3,11 +3,32 @@
...
- *
- * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateToModify = { ...state };
+  const history = [];
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
Viktor-Kost, line 12: it's better to use `switch` here

Viktor-Kost (changes requested):
GJ!
Just an improvement needed

src/transformStateWithClones.js
Daniilart01: sure, truly

Viktor-Kost (approved):

PR #2840, opened 2023-10-06
transformStateWithClones solution

ArtemSydun:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let statedCopy = { ...state };
+  const addProp = 'addProperties';
+  const removeProp = 'removeProperties';
+  const clear = 'clear';
```
MrHunterBoi, line 13: Would be better to refactor constant variables to uppercase letters like this: `addprop => ADD_PROP`

MrHunterBoi (changes requested):
Nice work

ksyvvkaa (approved):
nice job😜

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let statedCopy = { ...state };
+  const ADDPROP = 'addProperties';
+  const REMOVEPROP = 'removeProperties';
+  const CLEAR = 'clear';
+  const ACTIONSRESULT = [];
```
ArtemTeslenko, line 14: it is better to use UPPER_SNAKE_CASE when you use constants, also do not use contractions, example: 
ADDPROP -> ADD_PROPERTY

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let statedCopy = { ...state };
```
ArtemTeslenko, line 10:
```suggestion
    const stateCopy = { ...state };
```
it is better to use const here, object is not primitive, you can change it's properties if it will be const
stated -> state

ArtemTeslenko (approved):
Good job 👍, pay more attention to naming, repeat about let/const)

PR #2842, opened 2023-10-07
Stateful clones

TetianaMD:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+      currentState = { ...currentState, ...extraData };
+    } else
+    if (type === 'removeProperties') {
+      currentState = Object.keys(currentState).reduce((newState, key) => {
+        if (!keysToRemove.includes(key)) {
+          newState[key] = currentState[key];
+        }
+
+        return newState;
+      }, {});
```
VitaliyBodnar, line 26: let's make it simpler
```suggestion
     for (const value of actions.keysToRemove) {
        delete state[value];
      }
      break;
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    if (type === 'addProperties') {
```
VitaliyBodnar, line 16: it's better to use switch case statement instead of multiple if blocks

VitaliyBodnar (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    const { extraData, keysToRemove } = action;
+
+    switch (action.type) {
```
volodymyr-soltys97, line 16:
```suggestion
    const { extraData, keysToRemove, type } = action;

    switch (type) {
```

volodymyr-soltys97 (approved):
Great job 🔥

PR #2843, opened 2023-10-08
HW task "js_stateful-clones"

vashivan:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
-  // write code here
-}
+  const actionSteps = [];
+  let currentState = { ...state };
+  const add = 'addProperties';
+  const del = 'removeProperties';
+  const clear = 'clear';
+
+  for (const action of actions) {
+    switch (action.type) {
```
volodymyr-soltys97, line 17: You can use a destructuring for `action`
```suggestion
  for (const action of actions) {
    const { type, .......... } = action;
    switch (type) {
```

volodymyr-soltys97 (approved):
Well done 🔥

PR #2845, opened 2023-10-09
Solution

basiak89:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const result = [];
+  const cloneState = { ...state };
```
vlasiuksasha, line 12: The 'cloneState' variable is not really a clone since it gets modified in the loop. Consider renaming it to 'currentState' or something similar for better code readability.

src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
...
+        for (const key in cloneState) {
+          delete cloneState[key];
+        }
+        break;
+
+      default:
+        break;
+    }
+
+    result.push({ ...cloneState });
```
vlasiuksasha, line 34: Instead of pushing a clone of 'cloneState' to 'result' after every action, you can push the current state at the beginning of each iteration. This will ensure that the initial state is also included in your 'result' array.

src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const result = [];
```
vlasiuksasha, line 11: It is unnecessary to create an empty 'result' array at the beginning. You can declare it just before the loop where you actually start using it.

vlasiuksasha (changes requested):
check the comments

src/transformStateWithClones.js
basiak89: fixed

src/transformStateWithClones.js
basiak89: fixed

src/transformStateWithClones.js
basiak89: fixed

basiak89:
hello @vlasiuksasha  please could you have a look at second commit fixed?

PR #2846, opened 2023-10-10
add task solution

MariannaBo:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+      currentState = {
+        ...currentState, ...action.extraData,
+      };
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete currentState[key];
+      }
+    } else if (action.type === 'clear') {
+      currentState = {};
+    }
```
etojeDenys, line 25: let's use switch case instead of the multiple if else

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #2849, opened 2023-10-12
solution task

skaletskyi:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+      result.push({ ...stateClone });
+    }
+
+    if (action.type === 'clear') {
+      for (const keys in stateClone) {
+        delete stateClone[keys];
+      }
+
+      result.push({ ...stateClone });
+    }
```
etojeDenys, line 38: let's use switch case instead of the multiple If else

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+  const result = [];
+  const stateClone = { ...state };
+
+  for (const action of actions) {
+    if (action.extraData) {
+      for (const key in action.extraData) {
+        stateClone[key] = action.extraData[key];
+      }
+
+      result.push({ ...stateClone });
```
etojeDenys, line 19: we can push only one time at the end of the loop

etojeDenys (changes requested):

Andriy-Shymkiv (approved):

PR #2850, opened 2023-10-12
feat: add task solution

DominoSl:
image

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -6,8 +6,41 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClones = [];
+  const stateClone = Object.assign({}, state);
+
+  for (const action of actions) {
+    switch (action.type) {
+      case ('addProperties') :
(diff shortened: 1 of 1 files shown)


src/transformStateWithClones.js
```diff
@@ -6,8 +6,40 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClones = [];
+  const stateClone = Object.assign({}, state);
```
maxim2310, line 12: better to use {...state}

src/transformStateWithClones.js
```diff
@@ -6,8 +6,40 @@
...
+        for (const key of action.keysToRemove) {
+          delete stateClone[key];
+        }
+
+        break;
+
+      case ('clear') :
+        for (const key in stateClone) {
+          delete stateClone[key];
+        };
```
```suggestion
        stateClone = {}
```

maxim2310 (changes requested):
Good work, but lets make some improvement

src/transformStateWithClones.js
DominoSl: Gort it, thanks.

PR #2853, opened 2023-10-12
transformStateWithClones #1

VolodymyrLiubyn:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+    const varState = { ...stateCopy };
+
+    stateArray.push(varState);
+  }
+
+  return stateArray;
+}
+
+function transform(state, action) {
+  switch (action.type) {
```
volodymyr-soltys97, line 25: You a destruction for action

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+    case 'addProperties':
```
volodymyr-soltys97, line 26: Consider will be better if you create a variables for these magic words and use them everywhere

volodymyr-soltys97 (changes requested):
Good job 🔥 
Let's improve your code

src/transformStateWithClones.js
VolodymyrLiubyn: > Consider will be better if you create a variables for these magic words and use them everywhere

I got it. Fixed it

VolodymyrLiubyn (commented):
Code fixed

VolodymyrLiubyn (commented):
Code fixed

src/transformStateWithClones.js
VitaliyBodnar: not fixed 
The Previous mentor means that you need to create constant for addProperties, removeProperties, clear and reuse it, for instance:
```
const ACTIONS ={
 ADD,
 REMOVE,
 CLEAR
};
...
case ACTIONS.ADD:
```

VitaliyBodnar (changes requested):

src/transformStateWithClones.js
VolodymyrLiubyn: Ok, I realized what the mistake was. Attempt №3

anastasiiavorobiova (approved):
Great work!

PR #2855, opened 2023-10-13
Stateful-clones

Zava96:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+      for (const key in extraData) {
+        newState[key] = extraData[key];
+      }
+    } else if (type === 'removeProperties') {
+      for (const key of keysToRemove) {
+        delete newState[key];
+      }
+    } else if (type === 'clear') {
+      newState = {};
+    }
```
VitaliyBodnar, line 28: - is better to use switch case statement if instead of multiple if blocks
- move addProperties, removeProperties, clear to the constants as it will make a code cleaner

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+
+    result.push(newState);
+
+    currentState = newState;
```
VitaliyBodnar, line 32: don't need to reassign currentState
```suggestion
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    let newState = { ...currentState };
```
VitaliyBodnar, line 16: newState variables it redundant, because you can directly push copy of state to result
`result.push({ ...currentState })`

VitaliyBodnar (changes requested):

etojeDenys (approved):
good job

PR #2857, opened 2023-10-13
write transform state with clones function

natalia-serogina:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,56 @@
...
+          }
+
+          clonesArray.push(initialObjectCopy);
+        }
+        break;
+
+      case CLEAR_TYPE:
+        clonesArray.push({});
+        break;
+    }
```
sTorba24, line 56:
```suggestion
    switch (action.type) {
      case ADD_PROPERTIES_TYPE:
        stateCopy = {
          ...stateCopy,
          ...action.extraData,
        };

        break;

      case REMOVE_PROPERTIES_TYPE:
        for (const key of action.keysToRemove) {
            delete stateCopy[key];
         }
        break;

      case CLEAR_TYPE:
        stateCopy = {}
        break;
    }
    
    clonesArray.push({ ...stateCopy }) //push object copy
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,56 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const ADD_PROPERTIES_TYPE = 'addProperties';
+  const REMOVE_PROPERTIES_TYPE = 'removeProperties';
+  const CLEAR_TYPE = 'clear';
+  const PROPERTIES_TO_ADD_KEY = 'extraData';
+  const PROPERTIES_TO_REMOVE_KEY = 'keysToRemove';
```
sTorba24, line 14:
```suggestion
  const ADD_PROPERTIES = 'addProperties';
  const REMOVE_PROPERTIES= 'removeProperties';
  const CLEAR = 'clear';
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,56 @@
...
+
+  const clonesArray = [];
```
sTorba24, line 16:
```suggestion
  const clonesArray = [];
  let stateCopy = { ...state };
```

sTorba24: just create copy then work with it

sTorba24 (changes requested):
Good work! But it can be done much simpler

TarasHoliuk (approved):

PR #2865, opened 2023-10-17
Solution

Neison32:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,33 @@
...
  */
-function transformStateWithClones(state, actions) {
-  // write code here
+function cloneObject(obj) {
+  // Create a shallow clone of the object
+  return { ...obj };
 }
 
+function transformStateWithClones(state, actions) {
+  const stateHistory = [cloneObject(state)];
```
etojeDenys, line 15:
```suggestion
  const stateHistory = [];
```
and we can remove the slice at all

src/transformStateWithClones.js
```diff
@@ -6,8 +6,33 @@
...
+    } else if (action.type === 'removeProperties') {
+      // Remove specified properties from the state
+      for (const key of action.keysToRemove) {
+        delete newState[key];
+      }
+    } else if (action.type === 'clear') {
+      // Create an empty state
+      newState = {};
+    }
+
```
etojeDenys, line 32: let's use switch case instead of the multiple if else

PR #2870, opened 2023-10-20
solution

Markianek:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const transformState = [state];
```
maxim2310, line 10: in this way it would be better, you will change the clone and push it to state in the end of switch
```suggestion
  const state = []
  const clone= {...state};
```

maxim2310 (changes requested):

src/transformStateWithClones.js
Markianek: why redefine the state?

volodymyr-soltys97: This is to make a copy of the state, because working with the parameter directly is not a good idea

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+
+  for (const action of actions) {
+    let stateCopy = { ...transformState[transformState.length - 1] };
+
+    switch (action.type) {
```
volodymyr-soltys97, line 15: Use a destruction for `action`
```suggestion
  for (const action of actions) {
  
    const { type, ........ } = action;

    switch (type) {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/transformStateWithClones.js
Markianek: if i make these changes i get an error : 10:9  error  Parsing error: Identifier 'state' has already been declared

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+  let stateCopy = { ...state };
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
+        for (const key in extraData) {
+          stateCopy[key] = extraData[key];
+        }
```
anastasiiavorobiova, line 20: Consider using Object.assign() here instead of the loop for better performance

anastasiiavorobiova (changes requested):
Great work! Added a tiny comment to fix

volodymyr-soltys97 (approved):
Well done 🔥

PR #2871, opened 2023-10-21
Solution

Miponkin:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let current = { ...state };
```
vlasiuksasha, line 11: The cloning of the 'state' object is unnecessary in this line as it is already cloned inside the 'switch' statement.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+      case 'addProperties': {
+        current = {
+          ...current, ...action.extraData,
+        };
+        break;
+      }
+
+      case 'removeProperties': {
+        for (const key of action.keysToRemove) {
+          delete current[key];
```
vlasiuksasha, line 24: Deleting properties from an object is not recommended as it can lead to performance issues. It would be better to create a new object without these properties.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        }
+        break;
+      }
+
+      case 'clear': {
+        current = {};
+        break;
+      }
+
+      default: {
```
vlasiuksasha, line 34: Empty default case is unnecessary. It can be removed if it doesn't serve any purpose.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        break;
+      }
+    }
+    result.push({ ...current });
```
vlasiuksasha, line 38: The cloning of the 'current' object is unnecessary in this line as it is already cloned inside the 'switch' statement.

vlasiuksasha (changes requested):
check the comments

src/transformStateWithClones.js
Miponkin: [CODE STYLE]: switch/case should always have default case for error handling.

PR #2874, opened 2023-10-22
add task solutoin

7yrak34:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,46 @@
...
+          const deleteProp = { ...result[i - 1] };
+
+          for (const key of actions[i].keysToRemove) {
+            delete deleteProp[key];
+          }
+
+          result.push(deleteProp);
+          break;
+        }
+
```
etojeDenys, line 42: looks too complex, you can create the state object before the loop, and manipulate them in `for`
and at the end of the loop, only one time push that state object

etojeDenys (changes requested):
feel free to ask for some help in the chat

VitaliyBodnar (approved):

PR #2879, opened 2023-10-25
 add task solution

ABahatiy:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,43 @@
...
+        break;
+
+      default:
+        break;
+    }
+
+    // Push the copy of the state into the result array
+    result.push(nextState);
+    currentState = nextState;
+  }
```
sTorba24, line 43:
```suggestion
  const result = [];

  let currentState = { ...state };
  
  for (const action of actions) {
    switch (action.type) {
      case 'addProperties':
        Object.assign(currentState, action.extraData);
        break;

      case 'removeProperties':
        for (const key of action.keysToRemove) {
          delete currentState[key];
        }
        break;

      case 'clear':
        currentState = {}
        break;

      default:
        break;
    }

    result.push({ ...currentState });
  }
```

sTorba24 (changes requested):
Almost good! Lets improve it, can be use only 1 state variable to handle it

ABahatiy:
Done! Please check new changes :)

Andriy-Shymkiv (approved):

PR #2882, opened 2023-10-26
finished stateful clones

MrHunterBoi:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+          delete currentState[key];
+        }
+
+        break;
+
+      case 'clear':
+        for (const key in currentState) {
+          delete currentState[key];
+        }
+
```
ArtemSydun, line 33: no default case

ArtemSydun (commented):
default case?

src/transformStateWithClones.js
ksyvvkaa, line 33: 3. [CODE STYLE]: switch/case should always have default case for error handling.

ksyvvkaa (changes requested):
good job, add default

MrHunterBoi:
default case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const statesArray = [];
+  const currentState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          currentState[key] = action.extraData[key];
```
polosanya, line 17: You can use Object.assign instead

polosanya (approved):
Great!

PR #2883, opened 2023-10-26
Solution

Miishca:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClones = [Object.assign({}, state)];
+
+  for (const action of actions) {
+    let newState = Object.assign({}, stateClones[stateClones.length - 1]);
+
+    switch (action.type) {
+      case 'addProperties':
+        if (typeof action.extraData === 'object') {
```
anastasiia-tilikina, line 17: you don't need make this check since if type is 'addProperties' the sectonde argument will be always an object

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+          Object.assign(newState, action.extraData);
+        }
+        break;
+
+      case 'removeProperties':
+        if (Array.isArray(action.keysToRemove)) {
```
anastasiia-tilikina, line 23: The same here with array

anastasiia-tilikina (changes requested):
let's make some improvements

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClones = [Object.assign({}, state)];
```
```suggestion
  const stateClones = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        break;
+
+      default:
+        break;
+    }
+
+    stateClones.push(newState);
+  }
+
+  return stateClones.slice(1);
```
etojeDenys, line 37:
```suggestion
  return stateClones;
```

etojeDenys (changes requested):
we can simplify it

etojeDenys (changes requested):
.

src/transformStateWithClones.js
Miishca: when I did it I got errors

etojeDenys (approved):
good job

PR #2886, opened 2023-10-27
Develop

oleksandr-rudiuk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+  for (const action of actions) {
+    transform(copy, action);
+  }
+
+  function transform(result, action) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(result, action.extraData);
+
+        array.push({ ...result });
```
sTorba24, line 22: move this push after switch, u are doing it for every case, so it makes sense to not repeat and write it only once

sTorba24 (changes requested):

src/transformStateWithClones.js
oleksandr-rudiuk: In difference cases, difference actions. Object result has difference content. So how I can add difference data in array if I will make it only after switch

sTorba24: you are working with the same variable
the main point is 

`i want to change object`

okay, let me instruction

case 1 instruction:
  do this to same variable
case 2 inctruction:
  do smth different but to the same variable
  
so in result you have changed variable that you want to save in history array

sTorba24 (approved):

PR #2887, opened 2023-10-27
Solution

mxpcsr:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+          delete currentState[key];
+        }
+      }
+    } else if (action.type === 'clear') {
+      // Create an empty state object
+      currentState = {};
+    }
+
+    // Create a clone of the current state and add it to the result array
+    result.push({ ...currentState });
```
nazarmatsevych, line 33: Good work but in this case i'd recommend to use switch/case statement as we rely on single action.type value

nazarmatsevych (changes requested):

src/transformStateWithClones.js
mxpcsr: Better now?

nazarmatsevych (approved):
🔥

PR #2888, opened 2023-10-27
solution

yevtok:
[x] - don't mutate object or arrays - it will cause unexpected results later on. You should make copy using Object.assign or spread operator

[x] - use switch statement if you have limited amount of conditions.

[x] - use proper names for object copy

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const stateArr = [stateCopy];
+
+  for (const action of actions) {
+    let currentState = { ...stateArr[stateArr.length - 1] };
+
+    switch (action.type) {
+      case 'addProperties' :
+        currentState = Object.assign({}, currentState, action.extraData);
+        break;
+
+      case 'removeProperties' :
+        for (const key of action.keysToRemove) {
+          delete currentState[key];
+        }
+        break;
+      case 'clear' :
+        currentState = {};
+    }
+
+    stateArr.push(currentState);
+  }
+
+  stateArr.shift();
+
+  return stateArr;
 }
 
 module.exports = transformStateWithClones;


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+
+      case 'removeProperties' :
+        for (const key of action.keysToRemove) {
+          delete currentState[key];
+        }
+        break;
+      case 'clear' :
+        currentState = {};
+    }
+
```
ArtemSydun, line 29: no default case for switch

ArtemSydun (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const stateArr = [stateCopy];
```
ArtemTeslenko, line 11: Do not use contraction, also stateArray is too common name, use something more meaningful

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+
+  for (const action of actions) {
+    let currentState = { ...stateArr[stateArr.length - 1] };
```
ArtemTeslenko, line 14: no need to use let, it is no primitive data

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+
+      default: {
+        break;
+      }
+    }
+
+    stateArr.push(currentState);
+  }
+
+  stateArr.shift();
```
ArtemTeslenko, line 39: it can work without shift if you set initial value of the `stateArr` as `[]`

ArtemTeslenko (changes requested):
Good job) Simplify your code, do not overfill your code by logic, feel free to ask in chat if you need 😉

src/transformStateWithClones.js
yevtok: don't working

src/transformStateWithClones.js
yevtok: https://imageup.ru/img50/4596128/error-js.png.html

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
-  // write code here
+  const stateCopy = { ...state };
+  const outputStates = [stateCopy];
+
+  for (const action of actions) {
+    let currentState = { ...outputStates[outputStates.length - 1] };
+
+    switch (action.type) {
+      case 'addProperties' :
+        currentState = Object.assign({}, currentState, action.extraData);
```
BudnikOleksii, line 18:
```suggestion
       Object.assign(currentState, action.extraData);
```
The previous reviewer means that you don't need to assign it to the current state at all

BudnikOleksii (approved):
Good work 👍

PR #2892, opened 2023-10-30
add task solution

yefimov77:
- [x] Done!

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -36,6 +36,11 @@ function transform(copyState, action) {
       for (const key in copyState) {
         delete copyState[key];
       }
+      break;
+
+    default : {
+      throw new Error('Enter correct value');
+    }
   }
 }
 


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+      for (const key of action.keysToRemove) {
+        delete copyState[key];
+      }
+
+      break;
+
+    case 'clear':
+      for (const key in copyState) {
+        delete copyState[key];
+      }
```
yevtok, line 38: Good job, but need to add defaut case.

yevtok (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+    transform(copyState, action);
+    result.push({ ...copyState });
+  }
+
+  return result;
+}
+
+function transform(copyState, action) {
+  switch (action.type) {
+    case 'addProperties':
```
roman-mirzoian, line 23: It would be cool to put the option names in separate constants.

roman-mirzoian (approved):
Well done, but pay attention to the comment 👍

src/transformStateWithClones.js
yefimov77: U mean, make constants as sting values for different cases?

PR #2893, opened 2023-10-30
add task solution

mykhailoyurkovskyi:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,42 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const stateCopy = { ...state };
+  const PROPERTY_ADD = 'addProperties';
+  const PROPERTY_REMOVE = 'removeProperties';
+  const PROPERTY_CLEAR = 'clear';
+  const RESULT_OF_ACTIONS = [];
```
Anton-Kuchmasov, line 15: seems like it is not mandatory to use UPPER_CASE while naming variables such as RESULT_OF_ACTIONS 

rather use another name, like 'actionsLog' or 'arrayOfActions' instead

src/transformStateWithClones.js
```diff
@@ -8,6 +8,42 @@
...
+      case PROPERTY_REMOVE: {
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
+
+        break;
+      }
+
+      case PROPERTY_CLEAR: {
+        for (const key of Object.keys(stateCopy)) {
```
Anton-Kuchmasov, line 32: you can only do 
`stateCopy = {};`

just doing simple and better =)

Anton-Kuchmasov (changes requested):
nice! but fix just a couple of issues before approving

Anton-Kuchmasov (approved):
Great!

BudnikOleksii (approved):
Good job 👍

PR #2895, opened 2023-10-30
Solution

Patrykmclaren14:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,43 @@
  *
  * @return {Object[]}
  */
+
+const add = 'addProperties';
+const remove = 'removeProperties';
+const clear = 'clear';
```
zaura333, line 12: I'd consider placing those variables inside the function maybe? Just for clarity in this case

zaura333 (commented):
Rest looks good to me!

zaura333 (changes requested):
I think I like commented not reviewed last time, because I still have a pending review to do, and it sent me here again.

So... 😭 sorry!

Same as before: looks good, but I would consider moving variables into the function.

Patrykmclaren14:
> I think I like commented not reviewed last time, because I still have a pending review to do, and it sent me here again.
> 
> So... 😭 sorry!
> 
> Same as before: looks good, but I would consider moving variables into the function.

Patrykmclaren14:
> I think I like commented not reviewed last time, because I still have a pending review to do, and it sent me here again.
> 
> So... 😭 sorry!
> 
> Same as before: looks good, but I would consider moving variables into the function.

No problem bro 😎

mvjl000 (approved):
Nailed it! 🚀

PR #2897, opened 2023-10-30
complete task

boytrend0108:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,45 @@
...
+      return { ...newClone };
+
+    case 'removeProperties':
+      for (const el in action.keysToRemove) {
+        if (action.keysToRemove[el] in newClone) {
+          delete newClone[action.keysToRemove[el]];
+        }
+      };
+
+      return Object.keys(newClone).length === 0 ? {} : { ...newClone };
```
sTorba24, line 37: if newClone empty object, what a difference between  {} and { ...newClone }?

src/transformStateWithClones.js
```diff
@@ -6,8 +6,45 @@
...
+
+    case 'clear':
+      for (const key in newClone) {
+        delete newClone[key];
+      }
+
+      return {};
+
+    default: throw new Error('Enter correct value');
+  }
```
sTorba24, line 47:
```suggestion
  }
  
  return { ...newClone }
```

src/transformStateWithClones.js
```diff
@@ -6,8 +6,45 @@
...
+      };
+
+      return Object.keys(newClone).length === 0 ? {} : { ...newClone };
+
+    case 'clear':
+      for (const key in newClone) {
+        delete newClone[key];
+      }
+
+      return {};
```
sTorba24, line 44: move return outside switch as it it the same action for every case

sTorba24 (changes requested):
Almost good! Lets improve it

etojeDenys (approved):
looks good

PR #2900, opened 2023-10-30
add task solution

OstinM:


mberezhna (changes requested):
Good job! Almost done.
Please fix your code.

src/transformStateWithClones.js
```diff
@@ -1,13 +1,41 @@
...
 /**
- * @param {Object} state
+ * @param {Object} stateCopy
  * @param {Object[]} actions
  *
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const STATE_COPYS = [];
```
BudnikOleksii, line 10: It's not a constant, it's just a variable as you modify it, so the name should be in camelCase.

src/transformStateWithClones.js
```diff
@@ -1,13 +1,41 @@
...
+          if (stateCopy[keyRemove] !== undefined) {
+            delete stateCopy[keyRemove];
+          }
+        }
+        break;
+
+      case 'clear':
+        stateCopy = {};
+        break;
+    }
```
BudnikOleksii, line 32: Checklist:
- [CODE STYLE]: switch/case should always have default case for error handling.

BudnikOleksii (changes requested):
Good work 👍 
Just a few moments that should be fixed

src/transformStateWithClones.js
```diff
@@ -1,13 +1,44 @@
...
+    let stateCopy = { ...currentState };
+
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const keyRemove of action.keysToRemove) {
+          if (stateCopy[keyRemove] !== undefined) {
```
polosanya, line 23:
```suggestion
          if (stateCopy[keyRemove]) {
```

polosanya (approved):
Well done!

PR #2901, opened 2023-10-30
Develop

Fluebubble:
Task finished

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -8,6 +8,45 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const result = [];
+  let prevState = { ...state };
+
+  for (const action of actions) {
+    let stepResult = { ...prevState };
+
+    if (action.type === 'addProperties') {
+      stepResult = {
+        ...stepResult,
+        ...action.extraData,
+      };
+
+      result.push(stepResult);
+
+      prevState = { ...stepResult };
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        if (stepResult[key]) {
+          delete stepResult[key];
+        }
+      }
+
+      result.push(stepResult);
+
+      prevState = { ...stepResult };
+    }
+
+    if (action.type === 'clear') {
+      stepResult = {};
+
+      result.push(stepResult);
+
+      prevState = { ...stepResult };
+    }
+  }
+
+  return result;
 }
 
 module.exports = transformStateWithClones;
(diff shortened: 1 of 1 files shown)


anna-daryna (approved):
Good job)

src/transformStateWithClones.js
```diff
@@ -8,6 +8,53 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,53 @@
...
+        prevState = { ...stepResult };
+
+        break;
+
+      case 'clear':
+        stepResult = {};
+
+        result.push(stepResult);
+
+        prevState = { ...stepResult };
```
roman-mirzoian, line 48: It is not necessary to create a variable under an empty object.
```suggestion
        result.push({});
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,53 @@
...
+
+        break;
+
+      default:
+        throw Error;
```
roman-mirzoian, line 53: An error should tell the function user exactly what he did wrong.

roman-mirzoian (changes requested):

Fluebubble:
> Good job, but there is quite a bit of work to be done.

Fixed

src/transformStateWithClones.js
```diff
@@ -7,7 +7,50 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
BudnikOleksii, line 10: `result` is a bad name, it doesn't describe a variable, `statesHistory` or something similar would be better

src/transformStateWithClones.js
```diff
@@ -7,7 +7,50 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (stepResult[key]) {
+            delete stepResult[key];
+          }
+        }
+
+        result.push(stepResult);
```
BudnikOleksii, line 36: You push `stepResult` to an array in all cases.
Checklist:
- [DONT REPEAT YOURSELF] - If you perform same action in all switch cases - do it just once afterwards.

BudnikOleksii (changes requested):
Good work 👍 
Just one moment from the checklist, fix it.

src/transformStateWithClones.js
Fluebubble: But in `'clear'` case I need to push empty object, so I can't put it outside switch

BudnikOleksii (approved):
Good work 👍

PR #2902, opened 2023-10-30
Solution

BoblykNazar:


ilyapgitt (approved):
Nicely Done!!!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  const stateCopy = { ...state };
+  const arrOfClones = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, { ...action.extraData });
```
BudnikOleksii, line 16:
```suggestion
        Object.assign(stateCopy, action.extraData);
```
You can simplify in this case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        break;
+
+      case 'removeProperties':
+        const values = action.keysToRemove;
+
+        values.map(el => delete stateCopy[el]);
```
BudnikOleksii, line 22: `map` is used when you need to gain a new modified array, in this case you only iterate throw an array, so better to use `forEach` or `forof`

BudnikOleksii (approved):
Good work 👍 
Approved, but I left a few suggestions to improve your solution.

PR #2903, opened 2023-10-30
fead: add finished tack solution

kate-prakofyeva:
- [x] -[CODE STYLE] - don't mutate object or arrays - it will cause unexpected results later on. You should make copy using Object.assign or spread operator
- [x] [CODE STYLE]: Use switch statement if you have limited amount of conditions.
- [x] [CODE STYLE]: switch/case should always have default case for error handling.
- [x] [DONT REPEAT YOURSELF] - If you perform same action in all switch cases - do it just once afterwards.
- [x] [NAMING] - use proper names for object copy


DanyloSemenov (changes requested):
The code repeats the action of adding currentStep to the result array for each case in the switch statement multiple times. To avoid repetition, you can move this action outside of the switch statement.

Please correct it. Other than that, it's a great job 🔥

kate-prakofyeva:
> The code repeats the action of adding currentStep to the result array for each case in the switch statement multiple times. To avoid repetition, you can move this action outside of the switch statement.
> 
> Please correct it. Other than that, it's a great job 🔥

done, thanks

DanyloSemenov (approved):
Good job 🔥

src/transformStateWithClones.js
```diff
@@ -7,7 +7,42 @@
...
+    ...state,
+  };
+
+  const result = [STATE_COPY];
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+    const currentStep = {
+      ...result[result.length - 1],
+    };
```
nataliaklonowska, line 20: try to simplify the solution by not creating this variable :)

nataliaklonowska (changes requested):

src/transformStateWithClones.js
kate-prakofyeva: done

nataliaklonowska (approved):

PR #2905, opened 2023-10-30
solution

joanna-konopacka:


src/transformStateWithClones.js
```diff
@@ -1,13 +1,33 @@
...
- */
+"use strict";
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultArray = [{ ...state }];
+
+  for (let i = 0; i < actions.length; i++) {
+    const valueBeforePassing =
+      i === 0 ? { ...resultArray[i] } : { ...resultArray[i - 1] };
```
vlasiuksasha, line 8: You are creating a new copy of the state at the start of each iteration. This is unnecessary because you already have a copy from the previous iteration.

src/transformStateWithClones.js
```diff
@@ -1,13 +1,33 @@
...
+
+        resultArray[i] = Object.assign(valueBeforePassing, addProperties);
+        break;
+      case "removeProperties":
+        const arrayElementToDeleted = actions[i].keysToRemove;
+
+        arrayElementToDeleted.forEach((removeElement) => {
+          delete valueBeforePassing[removeElement];
+        });
+        resultArray[i] = valueBeforePassing;
```
vlasiuksasha, line 22: Instead of deleting key-value pairs one by one in a loop, you can use reduce function which is more efficient and cleaner.

src/transformStateWithClones.js
```diff
@@ -1,13 +1,33 @@
...
+
+  for (let i = 0; i < actions.length; i++) {
+    const valueBeforePassing =
+      i === 0 ? { ...resultArray[i] } : { ...resultArray[i - 1] };
+
+    switch (actions[i].type) {
+      case "addProperties":
+        const addProperties = actions[i].extraData;
+
+        resultArray[i] = Object.assign(valueBeforePassing, addProperties);
```
vlasiuksasha, line 14: Instead of using Object.assign to add properties, you can simply use the spread operator to add properties to an object. This is a more modern and cleaner approach.

src/transformStateWithClones.js
```diff
@@ -1,13 +1,33 @@
...
- *
- * @return {Object[]}
- */
+"use strict";
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultArray = [{ ...state }];
+
+  for (let i = 0; i < actions.length; i++) {
```
vlasiuksasha, line 6: Consider using a foreach loop instead of a for loop for better readability and simplicity.

src/transformStateWithClones.js
```diff
@@ -1,13 +1,33 @@
...
-
-/**
- * @param {Object} state
- * @param {Object[]} actions
- *
- * @return {Object[]}
- */
+"use strict";
+
 function transformStateWithClones(state, actions) {
```
vlasiuksasha, line 3: The function name 'transformStateWithClones' is unclear. Consider using a more descriptive name for the function.

src/transformStateWithClones.js
```diff
@@ -1,13 +1,33 @@
...
-  // write code here
+  const resultArray = [{ ...state }];
+
+  for (let i = 0; i < actions.length; i++) {
+    const valueBeforePassing =
+      i === 0 ? { ...resultArray[i] } : { ...resultArray[i - 1] };
+
+    switch (actions[i].type) {
+      case "addProperties":
```
vlasiuksasha, line 11: The switch-case statements lack default case. It is always good to have a default case in your switch statement for catching unexpected cases.

vlasiuksasha (changes requested):
check the comments

Ivan-Shvorak (approved):
LGTM

PR #2908, opened 2023-10-31
solution

bertsu02:


Barosz30 (approved):
Thought about putting return clone at the end of switch, but then i noticed how was your case 'clear' made. Nice one :D

nataliaklonowska (approved):

PR #2909, opened 2023-10-31
js_stateful-clones

Barosz30:
Totaly valid comment here


src/transformStateWithClones.js
```diff
@@ -6,8 +6,44 @@
...
+
+  for (const object of actions) {
+    switch (object.type) {
+      case 'addProperties':
+      // tried spread operator, but somethig is off with it,
+      // so let's go with for loop.
+
+        for (const key in object.extraData) {
+          stateCopy[key] = object.extraData[key];
+        }
```
Patrykmclaren14, line 24: Avoid nasted loops instead, you can use `Object.assign`.

src/transformStateWithClones.js
```diff
@@ -6,8 +6,44 @@
...
+      case 'removeProperties':
+        for (const key of object.keysToRemove) {
+          delete stateCopy[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
```
Patrykmclaren14, line 36: If we want to avoid nested loops, we can change `stateCopy = {}`;

src/transformStateWithClones.js
```diff
@@ -6,8 +6,44 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
   // write code here
+  const array = [];
+  const stateCopy = { ...state };
+  let pushedCopy = {};
```
Patrykmclaren14, line 14: `let pushedCopy `is unnecessary. You could perform all operations in `let stateCopy`

patrykyo20 (changes requested):
Address these issues, and it should be good.

MarcinLigmann (changes requested):
Apart from the added commands, the code looks good

src/transformStateWithClones.js
mvjl000: @MarcinLigmann We have to make some workarounds in order to use forEach on objects.
Also keep in mind that forEach is also a type of loop ;)

src/transformStateWithClones.js
```diff
@@ -6,8 +6,37 @@
...
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
   // write code here
+  const array = [];
+  let stateCopy = { ...state };
+  let pushedCopy = {};
+
+  for (const object of actions) {
```
mvjl000, line 16: I suggest to change the name of this constant
```suggestion
  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -6,8 +6,37 @@
...
+    switch (object.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, object.extraData);
+        break;
+
+      case 'removeProperties':
+        object.keysToRemove.forEach(Element => {
+          delete stateCopy[Element];
+        });
```
mvjl000, line 25: You always want to start those type of things in lowercase (camelCase). Btw maybe more descriptive name like `key`?
```suggestion
        object.keysToRemove.forEach(element => {
          delete stateCopy[element];
        });
```

src/transformStateWithClones.js
```diff
@@ -6,8 +6,37 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
   // write code here
+  const array = [];
+  let stateCopy = { ...state };
+  let pushedCopy = {};
```
mvjl000, line 14: You could do it with just one constant for a state copy
```suggestion
  const stateCopy = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -6,8 +6,37 @@
...
+        break;
+
+      case 'removeProperties':
+        object.keysToRemove.forEach(Element => {
+          delete stateCopy[Element];
+        });
+        break;
+
+      case 'clear':
+        stateCopy = {};
```
mvjl000, line 29: Now, since the stateCopy is a const, you can get back to a loop for deleting properties

Rembember to check `hasOwnProperty` before deleting it!

mvjl000 (changes requested):

src/transformStateWithClones.js
Barosz30: I put hasOwnProperty on deleting chosen keys. On clear it's not necessary, becouse we're removing ALL the keys.

mbruhler (approved):
Approved

PR #2910, opened 2023-10-31
add task solution

anna-daryna:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
BudnikOleksii, line 10: `result` is a bad name, the variable should better describe what is it. For example, you name it `statesHistory` or something similar

BudnikOleksii (approved):
Good job 👍

PR #2912, opened 2023-10-31
add task solution

ViktorAfk:


emilionila (changes requested):
Well done 🔥 
According to checklist: 2. Use switch statement if you have limited amount of conditions

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const COPY_OF_STATE = { ...state };
+  const ARRAY_OF_STATES = [];
```
ArtemTeslenko, line 11: we should use CONSTANT_VALUES if there is permanent value between different function's calls, here you can use just stateCopy and resultStates

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+
+      case 'removeProperties':
+        const removeValue = action.keysToRemove;
+
+        for (const key of Object.keys(COPY_OF_STATE)) {
+          if (removeValue.includes(key)) {
+            delete COPY_OF_STATE[key];
+          }
+        }
+        break;
```
ArtemTeslenko, line 27: you can iterate over the keysToRemove here and remove item from your array by key, it will simplify logic

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+
+      case 'clear':
+        for (const value of Object.keys(COPY_OF_STATE)) {
+          delete COPY_OF_STATE[value];
+        }
+        break;
```
ArtemTeslenko, line 33: there is possibility to avoid using loop if you use let instead of const on 11 line, you can do like `COPY_OF_STATE = {}` here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+    }
```
ArtemTeslenko, line 34: [Checklist](https://github.com/mate-academy/js_stateful-clones/blob/master/checklist.md) p. 3 - switch/case should always have default case for error handling.

ArtemTeslenko (changes requested):
Good job 👍 You can simplify your code a little bit, see comments 😉 Do not forget about checklist)

PR #2913, opened 2023-10-31
Solution

mickhakk:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,30 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  let newState = { ...state };
```
ViktorAfk, line 11: NewState is never reassigned. Use 'const' instead.

ViktorAfk (approved):
Well done! Fix minor mistake.

src/transformStateWithClones.js
ArtemTeslenko: It will be reassigned on 27 line

src/transformStateWithClones.js
```diff
@@ -8,6 +8,30 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete newState[key];
+        }
+        break;
+
+      case 'clear':
+        newState = {};
+        break;
+    }
```
ArtemTeslenko, line 29: [Checklist](https://github.com/mate-academy/js_stateful-clones/blob/master/checklist.md) p. 3 - switch/case should always have default case for error handling

ArtemTeslenko (changes requested):
Good job 👍 Just one moment, see comments. Do not forget about checklist 😉

ArtemTeslenko (approved):
Great 🔥

PR #2914, opened 2023-10-31
Solution

Sibunas:


mosowiecka (approved):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let stateCopy = { ...state };
+
+  for (const action of actions) {
+    stateCopy = { ...stateCopy };
+
+    switch (action.type) {
+      case 'addProperties':
```
JoyCoffeeAddict, line 17: add your cases to consts

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let stateCopy = { ...state };
+
+  for (const action of actions) {
+    stateCopy = { ...stateCopy };
```
JoyCoffeeAddict, line 14: you don't need to copy the stateCopy

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
+        break;
+      case 'clear':
+        stateCopy = {};
+        break;
+    }
```
JoyCoffeeAddict, line 28: You miss default case in switch. There you should display the error in the console `console.error("your message")`

JoyCoffeeAddict (changes requested):

src/transformStateWithClones.js
Sibunas: Without it, it doesn't pass my tests, this is so that it assigns each time to a new object and not to the reference of the same one. I'm not sure if I understand it correctly, but I was looking for a solution and I found this one.

JoyCoffeeAddict: how about doing the only copy in the `for` instead of above it.

JoyCoffeeAddict (changes requested):
Resolve previous issues.

src/transformStateWithClones.js
Sibunas: First, I create the first copy so as not to modify the input array, then I make a copy in the loop each time so that modifications in one loop do not affect the result added to result through references, without it it fails 2 tests.

  × Should handle multiple types (5ms)                                                                                                                                                                        
  × Should handle a long list of types (1ms)  
  
  If you add "let stateCopy = { ...state };" to the loop, each object in the result array would be identical. If I add this code to the loop, it also fails the tests.

mvjl000 (changes requested):
Resolve previous issues. In case of troubles please go to a QnA session

src/transformStateWithClones.js
Sibunas: I was sure I made it, i have it like this in my loptop, pushed it one more time.

src/transformStateWithClones.js
JoyCoffeeAddict: @Sibunas resolve.

JoyCoffeeAddict (changes requested):
Suggestions left on slack

mbruhler (approved):
Approved

PR #2916, opened 2023-10-31
Solution

mosowiecka:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,53 @@
...
+
+  const copy = {};
+
+  for (const key in obj) {
+    copy[key] = deepCopy(obj[key]);
+  }
+
+  return copy;
+}
+
```
WiktoriaBielec, line 26: wow, that's great that you did deepCopy here, I haven't even thought about it!

src/transformStateWithClones.js
```diff
@@ -6,8 +6,53 @@
...
 function transformStateWithClones(state, actions) {
   // write code here
+  const result = [];
+  const state2 = deepCopy(state);
+
+  for (const a of actions) {
+    if (a.type === 'addProperties') {
```
WiktoriaBielec, line 33: Everything is great but I think we need to use switch/case command here. It was in this (stupid) checklist in readme file:
It takes 5 minutes, hope it won't bother you much.

WiktoriaBielec: I meant switch instead of all those if (a.type ===.......)
Good luck!

WiktoriaBielec (changes requested):

src/transformStateWithClones.js
mosowiecka: Thx, added fixes. :)

src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
...
   // write code here
+  const result = [];
+  const stateCopy = { ...state };
+
+  for (const a of actions) {
+    switch (a.type) {
+      case 'addProperties':
+        for (const d in a.extraData) {
+          stateCopy[d] = a.extraData[d];
+        }
```
mvjl000, line 19: Instead of loop here, you could do this:
```suggestion
  Object.assign(stateCopy, a.extraData);
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const result = [];
+  const stateCopy = { ...state };
+
+  for (const a of actions) {
+    switch (a.type) {
```
mvjl000, line 15: If I were you, I would destructure action properties.

Second thing. I want you to add more descriptive names for variables. `a` -> `action` etc

mvjl000 (changes requested):
Good job!
I just have some small improvements ;)

mvjl000 (changes requested):
I can't see any changes since my last review, by the way I noticed that you are missing one point from the checklist

src/transformStateWithClones.js
mosowiecka: > If I were you, I would destructure action properties.
Could you explain?

mvjl000: @mosowiecka sure!

I meant something like this:
```
const { type } = action;
```

Obviously you can destructure more than just `type`

PR #2917, opened 2023-10-31
transformSWC done

Mantserov:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,47 @@
...
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+
+  const copy = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
VeDono, line 16: 'if' seems must be implement through 'switch case' (In [checklist](https://github.com/mate-academy/js_stateful-clones/blob/master/checklist.md) it's also includes 👀)
I'm like you, don't adapted for this things 😅

VeDono (approved):
This code get 12 green test of 12, so it's works. 🫶
But i think that mentor will be hold in hand 'Not-approveHammer' 👀

src/transformStateWithClones.js
```diff
@@ -6,8 +6,47 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+
+  const copy = { ...state };
```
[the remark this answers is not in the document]
roman-mirzoian: Good point!

src/transformStateWithClones.js
roman-mirzoian: Good poin!

roman-mirzoian (changes requested):

BudnikOleksii (approved):
Good job 👍

PR #2920, opened 2023-10-31
solution in switch case

YehorHeneralov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    let newState = { ...currentState };
```
polosanya, line 14: Do you really need to create one more copy?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+
+    switch (action.type) {
+      case 'addProperties':
+        const { extraData } = action;
```
polosanya, line 18: Nice! You can destructure all the needed props of the action at the beginning of the loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+
+        for (const key in extraData) {
+          newState[key] = extraData[key];
```
polosanya, line 21: Better use Object.assign instead

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
polosanya, line 10: Would be great to use more meaningful names

polosanya (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    let newState = { ...currentState };
+
+    switch (action.type) {
+      case 'addProperties':
+        const { extraData } = action;
+
+        newState = Object.assign({}, newState, extraData);
```
```suggestion
        Object.assign(newState, extraData);
```

polosanya (approved):
👍

PR #2921, opened 2023-10-31
Solution

MarcinLigmann:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let status = Object.assign({}, state);
```
mvjl000, line 10: Assign it to a constant. Also I would like to see some better naming
```suggestion
  const stateCopy = Object.assign({}, state);
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+  const result = [];
+
+  actions.forEach(action => {
+    let newState = {};
```
mvjl000, line 14: You don't need another object. You can do all acions on the `stateCopy`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+
+    if (action.type === 'addProperties') {
+      newState = Object.assign({}, status, action.extraData);
+      Object.assign({}, status, action.extraData);
```
mvjl000, line 18: With those simple change above, you can just do like this:
```suggestion
      Object.assign(stateCopy, action.extraData);
```

mvjl000: Btw, you could destructure properties of the `action` at the top of the loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let status = Object.assign({}, state);
+  const result = [];
+
+  actions.forEach(action => {
+    let newState = {};
+
+    if (action.type === 'addProperties') {
```
mvjl000, line 16: It would be better to use switch for this. See the checklist ;)

mvjl000 (changes requested):

szymanski-marcin:
totally agreed with @mvjl000 💯

src/transformStateWithClones.js
MarcinLigmann: ready to check :)

src/transformStateWithClones.js
```diff
@@ -7,32 +7,37 @@
...
+        Object.assign(stateCopy, extraData);
+        break;
 
-      action.keysToRemove.forEach(key => {
-        if (Object.hasOwnProperty.call(status, key)) {
-          delete newState[key];
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          delete stateCopy[key];
         }
```
Patrykmclaren14, line 24: In this task, we should avoid nested loops and instead use forEach. This will be clearer and more professional.

src/transformStateWithClones.js
```diff
@@ -7,32 +7,37 @@
...
+      case 'addProperties':
+        Object.assign(stateCopy, extraData);
+        break;
 
-      action.keysToRemove.forEach(key => {
-        if (Object.hasOwnProperty.call(status, key)) {
-          delete newState[key];
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          delete stateCopy[key];
```
Patrykmclaren14, line 23: In my opinion, using forEach will make the code clearer and look better.

src/transformStateWithClones.js
```diff
@@ -7,32 +7,37 @@
...
-      });
-    } else if (action.type === 'clear') {
-      newState = {};
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
```
Patrykmclaren14, line 31: also in this case

Patrykmclaren14 (approved):
Code looks good and complies with the intended goals. My comments are only suggestions for your solution.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+
+      default:
+        throw new Error('Unknown Action');
```
JoyCoffeeAddict, line 34: Usually it's a bad idea to throw without any context, because this means that consumer of this function (You in the different part of the code, or someone else in the team) has to know that this function is not safe and has to check if it will not throw.
I would change it to console.error('Your message')

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const stateHistory = [];
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
```
JoyCoffeeAddict, line 17: move your cases to consts 
`const ADD_PROPERTIES = addProperties`

JoyCoffeeAddict (changes requested):

JoyCoffeeAddict (approved):

PR #2923, opened 2023-10-31
task solution

VeDono:
Sorry for lose deadlines, but that logical tasks is very hard for me.

Seems that not for my mind ¯\_(ツ)_/¯

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let stateCopy = { ...state };
+  const result = []
+
+  for (const action of actions) {
+    switch(action.type) {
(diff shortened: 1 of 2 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let stateCopy = { ...state };
+  const result = []
+
+  for (const action of actions) {
+    switch(action.type) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        case 'addProperties':
+          stateCopy = { ...stateCopy, ...action.extraData}
```
polosanya, line 16: Better use Object.assign

polosanya (approved):
Good job!

PR #2924, opened 2023-10-31
done task

olegnovikov10:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+      for (const keys of item.keysToRemove) {
+        delete newState[keys];
+      }
+    }
+
+    if (item.type === clear) {
+      for (const keys in newState) {
+        delete newState[keys];
+      }
+    }
```
roman-mirzoian, line 35: image

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const res = [];
+  let newState = {
+    ...state,
+  };
+  const addProp = 'addProperties';
+  const removeProp = 'removeProperties';
+  const clear = 'clear';
```
roman-mirzoian, line 16: It's cool that you decided to put the type names into separate constants. 🔥 
But then they should be taken out of the function definition altogether (so that they are not created anew every time when the function is called) and called in uppercase style.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const res = [];
+  let newState = {
+    ...state,
+  };
```
roman-mirzoian, line 13: Good naming should be in all parts of the code
image
```suggestion
  const transformedStates = [];
  const stateCopy = { ...state };
```

roman-mirzoian (changes requested):

roman-mirzoian:
@ArtemSydun you left the right comment - the use of the `switch` design is in the checklist, and therefore it is a condition of the task itself

src/transformStateWithClones.js
```diff
@@ -6,8 +6,39 @@
...
+          delete stateCopy[keys];
+        }
+        break;
+
+      case clear:
+        for (const keys in stateCopy) {
+          delete stateCopy[keys];
+        }
+        break;
+    }
```
BudnikOleksii, line 36: Checklist:
- [CODE STYLE]: switch/case should always have default case for error handling.

BudnikOleksii (changes requested):
Good work 👍 
Just one moment need to be fixed

BudnikOleksii (approved):
👍

PR #2926, opened 2023-11-01
js_stateful-clones

IvanLugovskiy:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,56 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        addP(tmpState, action.extraData);
+
+        const obj1 = {};
+
+        Object.assign(obj1, tmpState);
+        arrRes.push(obj1);
```
lerastarynets, line 24: this logic is reusable, so pls extract it to separate func or const and reuse here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,56 @@
...
+        Object.assign(obj3, tmpState);
+        arrRes.push(obj3);
+        break;
+    }
+  }
+
+  return arrRes;
+}
+
+function addP(tmpState, extraData) {
```
lerastarynets, line 50: pls mind your naming, `addP` is not clear as well as following `removeP`

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

src/transformStateWithClones.js
IvanLugovskiy: I don't quite understand ( could you explaine more detail ?

IvanLugovskiy: if I right understand...)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,60 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const tmpState = {};
+
+  Object.assign(tmpState, state);
```
VitaliyBodnar, line 12: can be less
```suggestion
  const tmpState = {...state};
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,60 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        addProperties(tmpState, action.extraData);
+
+        const obj1 = {};
+
+        assignObj(obj1, tmpState, arrRes);
+        break;
```
VitaliyBodnar, line 24: don't need to write a lot of extra logic, it can be simpler 
check other cases
```suggestion
      case 'addProperties':
        Object.assign(tmpState, extraData);
        break;
```

VitaliyBodnar (changes requested):

nazarmatsevych (approved):
Good work!

PR #2927, opened 2023-11-01
Develop

WiktoriaBielec:


michal-reszka (approved):

src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
...
+            delete obj[key];
+          }
+        }
+        break;
+      case 'clear':
+        for (const key in obj) {
+          delete obj[key];
+        }
+        break;
+      default:
```
mvjl000, line 31: Checklist point 3. - [CODE STYLE]: switch/case should always have default case for error handling.

src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const arr = [];
+  const obj = { ...state };
+
+  for (const action of actions) {
```
mvjl000, line 14: If I were you, I would destructure action at the top of the loop

mvjl000 (changes requested):
Very good job! You just forgot about a very small detail ;)

mvjl000 (approved):

PR #2929, opened 2023-11-01
Develop

szymanski-marcin:
Task done

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -6,8 +6,35 @@
  *
  * @return {Object[]}
  */
+//
+//
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clonedState = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(clonedState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete clonedState[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in clonedState) {
+          delete clonedState[key];
+        }
+        break;
+    }
+    result.push({ ...clonedState });
+  }
+
+  return result;
 }
 
 module.exports = transformStateWithClones;
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -25,7 +25,7 @@ function transformStateWithClones(state, actions) {
         }
         break;
 
-      case 'clear':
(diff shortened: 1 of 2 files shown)


src/transformStateWithClones.js
```diff
@@ -6,8 +6,35 @@
...
+        Object.assign(clonedState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete clonedState[key];
+        }
+        break;
+
+      default:
```
Timi2424, line 28: Default case should log an error refering to improper input (read this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw). In previous commit you correctly named this case "clear", however you can simplify your code->

case 'clear':
      clonedState = {};
      break;

kate-prakofyeva: clear case = yours default
default case in switch:
`throw new Error('Unknown action type');`

PS: I don’t know why the system give me to do yours code review again

kate-prakofyeva (changes requested):

HarryPotter-4000 (approved):

kate-prakofyeva (approved):

src/transformStateWithClones.js
```diff
@@ -6,8 +6,38 @@
...
+        break;
+
+      case 'clear':
+        for (const key in clonedState) {
+          delete clonedState[key];
+        }
+        break;
+
+      default:
+        throw new Error('Unknown action type');
```
JoyCoffeeAddict, line 35: Usually it's a bad idea to throw without any context, because this means that consumer of this function (You in the different part of the code, or someone else in the team) has to know that this function is not safe and has to check if it will not throw.

src/transformStateWithClones.js
```diff
@@ -6,8 +6,38 @@
...
+//
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clonedState = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```
JoyCoffeeAddict, line 18: Create consts for your cases.

JoyCoffeeAddict (changes requested):

src/transformStateWithClones.js
HarryPotter-4000: Przepraszam, to była moja rekomendacja, a można więcej o tym, ze niebiezpieczne?

mbruhler (approved):
Approved

PR #2930, opened 2023-11-01
solution

aureliagorzna:


adrian200002 (approved):
In my opinion it's fine.

src/transformStateWithClones.js
```diff
@@ -6,8 +6,36 @@
...
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClones = [];
+  const recentState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```
JoyCoffeeAddict, line 16: add your cases to consts 
`cosnt ADD_PROPERTIES =  'addProperties'`

src/transformStateWithClones.js
```diff
@@ -6,8 +6,36 @@
...
+        });
+        break;
+
+      case 'clear':
+        for (const key in recentState) {
+          delete recentState[key];
+        }
+        break;
+      default:
+        throw new Error();
```
JoyCoffeeAddict, line 32: Usually it's a bad idea to throw without any context, because this means that consumer of this function (You in the different part of the code, or someone else in the team) has to know that this function is not safe and has to check if it will not throw.

JoyCoffeeAddict: Also, if you are already throwing, always throw with some meaningful message like: 
```throw new Error(`Unknown action of type: ${action.type}`);```

JoyCoffeeAddict (changes requested):

mvjl000 (approved):
Looks good! 🚀

PR #2932, opened 2023-11-01
task done

Timi2424:


web3-sorcerer (approved):
You should use uppercase letters for your constants, but other than that, everything looks great and all tests have passed.

mvjl000 (approved):
All good!

PR #2933, opened 2023-11-01
Develop

michal-reszka:


src/transformStateWithClones.js
```diff
@@ -7,29 +7,35 @@
...
+        };
+        break;
 
-  for (const element of actions) {
-    switch (element.type) {
       case 'removeProperties':
-        for (const item of element.keysToRemove) {
-          delete copyOfState[item];
+        for (const key of action.keysToRemove) {
+          delete currentState[key];
```
Patrykmclaren14, line 23: We want to avoid nested loops, so here you can use the `forEach` construction.

Patrykmclaren14 (approved):
The rest of your code looks great, so making changes to the loops would be excellent.

src/transformStateWithClones.js
mvjl000: @Patrykmclaren14 keep in mind that `forEach` is also a type of loop ;)
Time complexity is the same

mvjl000 (approved):
Good job!

PR #2934, opened 2023-11-02
add task solution

VladimirKolesnikov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+      case 'addProperties':
+        const { extraData } = action;
+
+        stateCopy = {
+          ...stateCopy,
+          ...extraData,
+        };
+        break;
+      case 'removeProperties':
+        const { keysToRemove } = action;
```
maxim2310, line 26: you  could simplify it
```suggestion
    const { type, extraData, keysToRemove  } = action;

    switch (type) {
      case 'addProperties':
      
        stateCopy = {
          ...stateCopy,
          ...extraData,
        };
        break;
      case 'removeProperties':
```

maxim2310 (approved):
Good work!!!

src/transformStateWithClones.js
VladimirKolesnikov: Thank you))

PR #2939, opened 2023-11-02
solution

DanyloSemenov:


mbruhler (changes requested):
mate refresh

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+  const result = [];
+  const stateCopy = { ...state };
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
+        for (const data in extraData) {
+          stateCopy[data] = extraData[data];
```
Barosz30, line 19: avoid nested loops. Instead of for () you can use `Object.assign`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          delete stateCopy[key];
```
Barosz30, line 25: You should also me it work only if State copy actualy has the property you wanna delete.
"if (stateCopy.hasOwnProperty(key)"
Otherwise, it could potentialy cause some problem. Overall, realy good code bro ;)

Barosz30 (changes requested):

src/transformStateWithClones.js
mbruhler: this is true, you can just use `Object.assign`

src/transformStateWithClones.js
mbruhler: yeah, that's true, as those are keys pointed by programmer / user of the code.

mbruhler (changes requested):
Please add changes from @Barosz30

mbruhler (changes requested):
changes are needed

mbruhler (approved):
Approved

NOTE: for clearing you can change `stateCopy` to `let` and then just `stateCopy = {}`

PR #2940, opened 2023-11-02
Develop

RadoslawBielski:


src/transformStateWithClones.js
```diff
@@ -1,13 +1,36 @@
...
- *
- * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let currentState = { ...state };
+  const resultStates = [];
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```
JoyCoffeeAddict, line 11: Add your cases to consts

src/transformStateWithClones.js
```diff
@@ -1,13 +1,36 @@
...
+          delete currentState[key];
+        }
+        break;
+
+      case 'clear':
+        currentState = {};
+        break;
+
+      default:
+        throw new Error('Unknown action');
```
JoyCoffeeAddict, line 28: Usually it's a bad idea to throw without any context, because this means that consumer of this function (You in the different part of the code, or someone else in the team) has to know that this function is not safe and has to check if it will not throw.

JoyCoffeeAddict: You can disable linting for this line only.

JoyCoffeeAddict (changes requested):

RadoslawBielski:
I hope is okey now

mvjl000 (approved):

PR #2942, opened 2023-11-02
Solution

ykhmaruk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let firstCopy = { ...state };
+
+  for (const action of actions) {
+    const copy = { ...firstCopy };
```
sarakhmen, line 14: do you really need both 'firstCopy' and 'copy' variables?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+
+    if (action.type === 'addProperties') {
+      for (const key in action.extraData) {
+        copy[key] = action.extraData[key];
+      }
```
sarakhmen, line 19: consider using Object.assign() here instead

sarakhmen (changes requested):

PR #2943, opened 2023-11-02
add task solution

maks2809:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const results = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
maxim2310, line 14: better to use switch case here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      currentState = {
+        ...currentState,
+        ...action.extraData,
+      };
+      results.push({ ...currentState });
```
maxim2310, line 19: you do it it every cases, move this line to the and of loop

maxim2310 (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+        currentState = { ...newState };
+        break;
+      case 'clear':
+        currentState = {};
+        newState = { ...currentState };
+        break;
+      default:
+        newState = { ...currentState };
+    }
+    results.push(newState);
```
etojeDenys, line 39: you can just make like this
```suggestion
    results.push({...newState});
```

etojeDenys (changes requested):
image
still not fixed

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        break;
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          delete STATE_COPY[key];
+        }
+        break;
+      case 'clear':
+        for (const key in STATE_COPY) {
+          delete STATE_COPY[key];
+        }
```
maxim2310, line 28:
```suggestion
        for (const key in STATE_COPY) {
          STATE_COPY = {}
        }
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
-  // write code here
+  const STATE_COPY = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
+        Object.assign(STATE_COPY, extraData);
```
maxim2310, line 18:
```suggestion
        STATE_COPY = {...STATE_COPY, ...extraData}```

maxim2310 (approved):
Good work!!!

src/transformStateWithClones.js
maks2809: When I fixed this, I got errors
  28:20  error  'key' is assigned a value but never used  no-unused-vars
  
  `'use strict';

/**
 * @param {Object} state
 * @param {Object[]} actions
 *
 * @return {Object[]}
 */
function transformStateWithClones(state, actions) {
  let STATE_COPY = { ...state };
  const result = [];

  for (const action of actions) {
    const { type, extraData, keysToRemove } = action;

    switch (type) {
      case 'addProperties':
        STATE_COPY = {
          ...STATE_COPY, ...extraData,
        };
        break;
      case 'removeProperties':
        for (const key of keysToRemove) {
          delete STATE_COPY[key];
        }
        break;
      case 'clear':
        for (const key in STATE_COPY) {
          STATE_COPY = {};
        }
        break;
    }

    result.push({
      ...STATE_COPY,
    });
  }

  return result;
}

module.exports = transformStateWithClones;
`

PR #2948, opened 2023-11-04
feat [js-adv]: implement object`s state transformer

lexzane:


src/transformStateWithClones.js
```diff
@@ -1,13 +1,37 @@
...
  * @param {Object[]} actions
  *
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let temp = { ...state };
+  const states = [];
+
+  actions.forEach(action => {
```
SerhiiKirik, line 17: Would be better to use `for of` loop instead forEach, don't use forEach, it's a bad practice

src/transformStateWithClones.js
```diff
@@ -1,13 +1,37 @@
...
+    if (action.type === ADD_ACTION) {
+      temp = Object.assign({}, temp, action.extraData);
+    }
+
+    if (action.type === REMOVE_ACTION) {
+      temp = Object.fromEntries(Object.entries(temp)
+        .filter(entry => !action.keysToRemove.includes(entry[0])));
```
SerhiiKirik, line 24: Implement it with `delete` method,
```
const obj = { x: 1 };
delete obj.x
```

src/transformStateWithClones.js
```diff
@@ -1,13 +1,37 @@
 'use strict';
 
+const ADD_ACTION = 'addProperties';
+const REMOVE_ACTION = 'removeProperties';
+const CLEAR_ACTION = 'clear';
```
SerhiiKirik, line 5: let's combine these actions to one object

SerhiiKirik (changes requested):
Well done 🔥 

PR #2951, opened 2023-11-05
first commit transform state with clones

dima91020:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+
+      keysToRemove.forEach(key => {
+        delete currentClone[key];
+      });
+    } else if (action.type === 'clear') {
+      for (const key in currentClone) {
+        delete currentClone[key];
+      }
+      currentClone = {};
+    }
```
etojeDenys, line 31: better to use switch/case instead of the multiple if/else

etojeDenys (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+  const clonesArray = [];
+  const cloneState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          cloneState[key] = action.extraData[key];
+        }
+        clonesArray.push({ ...cloneState });
```
maxim2310, line 19: you are doing this 3 times, would be better to move it in the end of loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        for (const key of action.keysToRemove) {
+          delete cloneState[key];
+        }
+        clonesArray.push({ ...cloneState });
+        break;
+
+      case 'clear':
+        for (const key in cloneState) {
+          delete cloneState[key];
+        }
```
maxim2310, line 32:
```suggestion
        cloneState = {}
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
-  // write code here
+  const clonesArray = [];
+  const cloneState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          cloneState[key] = action.extraData[key];
+        }
```
maxim2310, line 18:
```suggestion
        cloneState = {...cloneState, ...action.extraData}
```

maxim2310 (changes requested):

maxim2310 (approved):
now looks much better, Good job!!!

PR #2952, opened 2023-11-06
completed task

Stepan-R:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const cloneOfStates = [];
+  const cloneState = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
VitaliyBodnar, line 14: It's better to use switch case statement instead of multiple if else blocks

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+      Object.assign(cloneState, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete cloneState[key];
+      }
+
+      Object.assign(cloneState);
```
VitaliyBodnar, line 23: Redundant here

VitaliyBodnar (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+    }
+
+    if (action.type === 'clear') {
+      for (const key of Object.keys(cloneState)) {
+        delete cloneState[key];
+      }
+
+      Object.assign(cloneState);
```
VitaliyBodnar, line 31: Redundant here

VitaliyBodnar (changes requested):
.

Stepan-R:
@VitaliyBodnar Код не хоче працювати тепер, що не так ?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        }
+        break;
 
+      case 'clear':
+        for (const key of Object.keys(cloneState)) {
+          delete cloneState[key];
+        }
+        break;
+
+      default:
```
Viktor-Kost, line 32: it is a good practice when you add a default behavior for switch statement. But don't let it be empty. If you have no default case you can just throw an Error with text like "unexpected action"

Viktor-Kost (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        for (const key of action.keysToRemove) {
+          delete cloneState[key];
+        }
+        break;
 
+      case 'clear':
+        for (const key of Object.keys(cloneState)) {
+          delete cloneState[key];
+        }
+        break;
```
Viktor-Kost, line 30: emm... why did you delete `default` case? You could just set a logic what to do when you've got an unexpected action
```suggestion
        break;
      
      default:
        throw new Error('Unexpected action!');
```

Viktor-Kost (approved):
Approved. Yet please apply recommendation I left with a comment

src/transformStateWithClones.js
Stepan-R: добавив, я просто не знав як добавляти такі `Error`

PR #2953, opened 2023-11-06
add task solution

xXPaRaZorXx:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        delete newState[key];
+      }
+      result.push({ ...newState });
+      currentState = { ...newState };
+    } else if (action.type === 'clear') {
+      const newState = {};
+
+      result.push({ ...newState });
+      currentState = { ...newState };
+    }
```
etojeDenys, line 34: better to use switch/case instead of the if/else

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      const newState = {
+        ...currentState, ...action.extraData,
+      };
+
+      result.push({ ...newState });
+      currentState = { ...newState };
```
etojeDenys, line 20: you can do it only one time, after all switch/cases

etojeDenys (changes requested):

IvanFesenko (approved):
Well done

PR #2954, opened 2023-11-07
add task solution

mykola-kharin:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+      for (const key of act.keysToRemove) {
+        delete stateCopy[key];
+      }
+    }
+
+    if (act.type === 'clear') {
+      for (const key in stateCopy) {
+        delete stateCopy[key];
+      }
+    }
```
etojeDenys, line 28: better to use switch/case instead of the multiple if/else

etojeDenys (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,55 @@
...
+//     if (act.type === 'clear') {
+//       for (const key in stateCopy) {
+//         delete stateCopy[key];
+//       }
+//     }
+//     transformState.push({ ...stateCopy });
+//   }
+
+//   return transformState;
+// }
```
loralevitska, line 59: Remove comment. Try it write obvious code that explain all we need without comments)
```suggestion
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,55 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const transformState = [];
+  let stateCopy = { ...state };
+
+  for (const act of actions) {
+    switch (act.type) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,55 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const transformState = [];
+  let stateCopy = { ...state };
+
+  for (const act of actions) {
```
loralevitska, line 13: Add more meaningful variable name
```suggestion
  for (const action of actions) {
```

loralevitska (changes requested):
Almost perfect) Let's improve

src/transformStateWithClones.js
mykola-kharin: I didn't get what you mean?

etojeDenys (approved):
awesome

PR #2960, opened 2023-11-11
add task soulution

Kozubowicz:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete tmp[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      tmp = {};
+    }
```
nazarmatsevych, line 28: Good work but let's rewrite it to switch/case statement since we rely on single action.type value

nazarmatsevych (changes requested):

nazarmatsevych (approved):
👍

PR #2963, opened 2023-11-12
add task solution

pawelnowicki87:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  const result = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
volodymyr-soltys97, line 14: Use a switch case instead of of

volodymyr-soltys97, line 14: Use a destruction for `actions`
const { type, .............. } = action;

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+
+    if (action.type === 'removeProperties') {
+      for (const keyToRemove of action.keysToRemove) {
+        if (currentState.hasOwnProperty(keyToRemove)) {
+          delete currentState[keyToRemove];
+        }
+      }
+    }
 
+    if (action.type === 'clear') {
```
volodymyr-soltys97, line 28: Create a variable for this magic words  and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code 👍

volodymyr-soltys97 (approved):
Great work 🔥

PR #2965, opened 2023-11-13
task finished

Deemacr:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+          delete actionToSave[key];
+        }
+        stateActionsArray.push(Object.assign({}, actionToSave));
+      }
+
+      if (value === 'clear') {
+        stateActionsArray.push({});
+        actionToSave = {};
+      }
+    }
```
nazarmatsevych, line 35: Good work, let's rewrite it to switch/case statement since we rely on a single action.type value

nazarmatsevych (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,47 @@
...
+          stateActionsArray.push({});
+          actionToSave = {};
+
+          break;
+        }
+      }
+    }
+  }
+
+  if (stateActionsArray.length === 0) {
```
lerastarynets, line 46:
```suggestion
  if (!stateActionsArray.length) {
```

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #2968, opened 2023-11-14
solution

Dmytro-Kretov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,26 @@
...
+      Object.assign(clone, obj.extraData);
+    } else if (obj.type === 'removeProperties') {
+      for (const remove of obj.keysToRemove) {
+        delete clone[remove];
+      }
+    } else if (obj.type === 'clear') {
+      for (const data in clone) {
+        delete clone[data];
+      }
+    }
```
etojeDenys, line 24: let's use switch/case instead of the multiple if/else

etojeDenys (changes requested):

maxim2310 (approved):
Good work!!!

PR #2969, opened 2023-11-14
add stateful clones

AnoriFrell:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
...
+        newArray.push({ ...copyState });
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(copyState)) {
+          delete copyState[key];
+        }
+        newArray.push({ ...copyState });
+        break;
+    }
```
VitaliyBondarenko1982, line 34:
```suggestion
        break;

      case 'removeProperties':
        for (const key of action.keysToRemove) {
          delete copyState[key];
        }
        break;

      case 'clear':
        for (const key of Object.keys(copyState)) {
          delete copyState[key];
        }
        break;
    }
    
    newArray.push({ ...copyState });
```
you can push to newArray after switch case. it will reduce amount of code.
also add default case (for example throw some errow for unexpected type)

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comment.

PR #2979, opened 2023-11-16
added task solution

AlNeon:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+
+    if (actions[i].type === 'addProperties') {
+      Object.assign(nextState, actions[i].extraData);
+    } else if (actions[i].type === 'removeProperties') {
+      for (let j = 0; j < actions[i].keysToRemove.length; j++) {
+        delete nextState[actions[i].keysToRemove[j]];
+      }
+    } else {
+      Object.keys(nextState).forEach(key => delete nextState[key]);
+    }
```
etojeDenys, line 29: let's use switch/case instead of the multiple if/else

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+  const newActions = [];
+
+  for (let i = 0; i < actions.length; i++) {
+    const nextState = {};
+
+    if (i > 0) {
+      Object.assign(nextState, newActions[i - 1]);
+    } else {
+      Object.assign(nextState, state);
+    }
```
etojeDenys, line 19: you can just create the copy object before the loop

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #2982, opened 2023-11-18
js_stateful-clones solution

serjgrm:


nazarmatsevych (changes requested):
Good work, let's make some improvements

PR #2989, opened 2023-11-21
transformStateWithClones

baraban2003:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        stateCopy = {
+          ...stateCopy,
+          ...action.extraData,
+        };
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
```
volodymyr-soltys97, line 23: You can use a destruction action
```suggestion
  for (const action of actions) {
    const { type, extraData, keysToRemove } = action;
    switch (type) {
      case 'addProperties':
        stateCopy = {
          ...stateCopy,
          ...extraData,
        };
        break;

      case 'removeProperties':
        for (const key of keysToRemove) {
```

volodymyr-soltys97 (approved):
Great work 🔥

PR #2994, opened 2023-11-24
solution

oskushnir:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(currentState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (currentState.hasOwnProperty(key)) {
```
anastasiiavorobiova, line 21:
```suggestion

```
There is no need to add an additional check here, the delete doesn't throw errors

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Awesome!

PR #2996, opened 2023-11-27
the solution transformState

gboreyko:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,38 @@
...
+      for (let i = 0; i < removeKeys.length; i++) {
+        delete copyState[removeKeys[i]];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in copyState) {
+        delete copyState[key];
+      }
+    }
```
etojeDenys, line 35: let's use switch/case instead of the multiple if/else

etojeDenys (changes requested):

src/transformStateWithClones.js
```diff
@@ -6,8 +6,40 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
+  const arrayStates = [];
+
+  for (let k = 0; k < actions.length; k++) {
+    const action = actions[k];
+
+    switch (action.type) {
+      case 'addProperties':
```
OleziO, line 18: Move all cases strings to the object

OleziO (approved):

PR #2999, opened 2023-11-27
add task solution

taninska:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+        delete modified[item];
+      }
+    }
+
+    if (actions[i].type === 'clear') {
+      for (const key of Object.keys(modified)) {
+        delete modified[key];
+      }
+    }
+
```
etojeDenys, line 29: let's use switch/case instead of the multiple if/else statements

etojeDenys (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      case 'removeProperties':
+        for (const item of actions[i].keysToRemove) {
+          delete modified[item];
+        }
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(modified)) {
+          delete modified[key];
+        }
```
maxim2310, line 28:
```suggestion
        modified = {}
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+    }
+
+    resultArr.push(Object.assign({}, modified));
```
```suggestion
    resultArr.push({...modified});
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const modified = Object.assign({}, state);
```
maxim2310, line 10:
```suggestion
  const modified = {...state};
```

maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

src/transformStateWithClones.js
taninska: strange, but before this option did not erase all the keys. 👎 
Therefore, I decided to manually erase them each through a cycle :-/

etojeDenys (approved):
good job

PR #3003, opened 2023-11-29
Solution

artemvlasiuk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultArray = [];
+  const stateClone = { ...state };
```
```suggestion
  let stateClone = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+
+      const removed = { ...stateClone };
+
+      resultArray.push(removed);
+    }
+
+    if (action.type === 'clear') {
+      for (const key in stateClone) {
+        delete stateClone[key];
+      }
```
etojeDenys, line 35:
```suggestion
      stateClone = {};
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete stateClone[key];
+      }
+
+      const removed = { ...stateClone };
+
+      resultArray.push(removed);
+    }
```
etojeDenys, line 30: let's use switch/case instead of the multiple if/else

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+
+    if (action.type === 'clear') {
+      for (const key in stateClone) {
+        delete stateClone[key];
+      }
+
+      const cleared = { ...stateClone };
+
+      resultArray.push(cleared);
```
etojeDenys, line 39: you can push only time after all cases in switch

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #3012, opened 2023-12-01
State has been cloned

maks-n:


maxim2310 (approved):
Good work!!!
But i think its kind of overengineering creating separate func on such simple action here, and you change external object that come to you sub func as attribute, now its not a problem but can lead to unexpected behavior later

PR #3023, opened 2023-12-05
solution

Phetryska:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const result = [];
+
+  for (const i of actions) {
```
maxim2310, line 13:
```suggestion
  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+    switch (i.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, i.extraData);
+        result.push(Object.assign({}, stateCopy));
```
maxim2310, line 17: its common action move it after all switch case
```suggestion
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        for (const key of i.keysToRemove) {
+          delete stateCopy[key];
+        }
+        result.push(Object.assign({}, stateCopy));
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
```
maxim2310, line 30:
```suggestion
        stateCopy = {}
```

maxim2310 (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        break;
+    }
+
+    result.push(Object.assign({}, stateCopy));
```
VitaliyBodnar, line 32: it's better to use spread operator for making copy
```suggestion
    result.push({...stateCopy}));
```

src/transformStateWithClones.js
VitaliyBodnar: not fixed

VitaliyBodnar (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+          delete stateCopy[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(stateCopy)) {
+          delete stateCopy[key];
+        }
+        break;
+    }
```
Viktor-Kost, line 30: it is a good practice when you add a default behavior for switch statement. Yet, if you don't have such one - just throw an Error with text like "unexpected action"

Viktor-Kost (approved):

PR #3027, opened 2023-12-07
Solution

MetatelKruka:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
```
OleziO, line 10: Remove all comments from the code.

OleziO (changes requested):

etojeDenys (changes requested):
i cannot find logic with the return value
image

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = { ...state };
+  const stateVersions = [];
+
+  for (const action of actions) {
+    switch (action.type) {
```

loralevitska (approved):
Well done)

PR #3029, opened 2023-12-08
Solution

SerhiiBabskyi:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,46 @@
...
+    Object.assign(changingObject, extraData);
+  }
+
+  function removeProperties(changingObject, keysToRemove) {
+    for (const key of keysToRemove) {
+      delete changingObject[key];
+    }
+  }
+
+  function clearProperties(changingObject) {
```
TarasHoliuk, line 45: as for a separate function with such name, it's fine logic
but in this particular case you might be able to do it:

image

because you anyway do copy after `swith` block

src/transformStateWithClones.js
```diff
@@ -7,7 +7,46 @@
...
+        clearProperties(STATE_CLONE);
+        break;
+    }
+
+    RESULT_ARR.push({ ...STATE_CLONE });
+  }
+
+  return RESULT_ARR;
+
+  function addProperties(changingObject, extraData) {
```
TarasHoliuk, line 35: generally, you really use it only once, but if you move the logic to the function, you better declare the function outside the `transformStateWithClones` function

PR #3032, opened 2023-12-09
Solution

AntonDuchenko:


vSainiuk (approved):
Looks good!

src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
...
+
+  function addingActionResultToArray() {
+    actionsResultArr.push({ ...stateCopy });
+  }
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        addingActionResultToArray();
```
SerhiiKirik, line 22: DRY
you can just put it after the `switch/case`
```suggestion
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const stateCopy = { ...state };
+  const actionsResultArr = [];
+
+  function addingActionResultToArray() {
```
SerhiiKirik, line 14: you can do it without function, so, remove it
Also, would be better if your function:
- Produce No Side Effects
A pure function produces no side effects, which means that it can’t alter any external state.

- Immutability
JavaScript’s object arguments are references, which means that if a function were to mutate a property on an object or array parameter, that would mutate state that is accessible outside the function. Pure functions must not mutate external state.

SerhiiKirik (changes requested):
Well done 🔥 

SerhiiKirik (approved):
LGTM 🔥

PR #3034, opened 2023-12-09
solution

vSainiuk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        for (const value of action.keysToRemove) {
+          delete initialObject[value];
+        }
+        break;
+
+      case 'clear':
+        for (const key in initialObject) {
+          delete initialObject[key];
+        }
+        break;
```
rtkachuk-moon, line 29: 3.[CODE STYLE]: switch/case should always have default case for error handling.

rtkachuk-moon (approved):
Good work) check my comments)

src/transformStateWithClones.js
dinaminahi: Agree with @rtkachuk-moon, let`s add default case

dinaminahi (changes requested):
Great job! Let`s do just one fix

PR #3038, opened 2023-12-11
Solution

nabokok:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const newObj = { ...state };
```
artem8746, line 11: mb better stateCopy?)

src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
...
+  const result = [];
+
+  for (const obj of actions) {
```
artem8746, line 14: mb better `action` instead of `obj`

src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
...
+    switch (obj.type) {
+      case 'addProperties':
+        Object.assign(newObj, obj.extraData);
+        break;
+
+      case 'removeProperties':
+        for (let i = 0; i < obj.keysToRemove.length; i++) {
+          delete newObj[obj.keysToRemove[i]];
+        }
```
artem8746, line 23: i think it would be better to create functions for all cases

src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
...
+        break;
+
+      case 'clear':
+        for (const key in newObj) {
+          delete newObj[key];
+        }
+        break;
+
+      default:
```
artem8746, line 32: 3. [Checklist] switch/case should always have default case for error handling. (add `throw Error` in `default`)

artem8746 (changes requested):
Excellent job!!!
Just need a few fixes

src/transformStateWithClones.js
nabokok: I don't think so. It takes more place. This code is easy enough to read

src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
...
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (let i = 0; i < action.keysToRemove.length; i++) {
```
SerhiiKirik, line 21: In `clear` action you make delete with `for of` loop, but here with simple loop)
Just make it in one style, with `for of`

SerhiiKirik (approved):
Nice 🔥

PR #3040, opened 2023-12-11
Solution

avakiel:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,26 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let myState = { ...state };
```
ssaviak, line 11: The strange name of the current state...

ssaviak (approved):
Good job! The next time just remember - the variables have to be named what they mean and their target. The variable name mystate looks so strange... Anyway you can use the name "current state"

src/transformStateWithClones.js
```diff
@@ -7,7 +7,26 @@
...
+      stateCopy = {};
+      result.push({});
+    } else if (actions[i].type === 'addProperties') {
+      stateCopy = {
+        ...stateCopy,
+        ...actions[i].extraData,
+      };
+      result.push({ ...stateCopy });
+    } else if (actions[i].type === 'removeProperties') {
+      actions[i].keysToRemove.forEach(e => delete stateCopy[e]);
```
DanilWeda, line 24: try to use another method, which can return object from array ( nor forEach )

src/transformStateWithClones.js
```diff
@@ -7,7 +7,26 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let stateCopy = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'clear') {
+      stateCopy = {};
+      result.push({});
+    } else if (actions[i].type === 'addProperties') {
```
DanilWeda, line 17: You can write all values in constant. 
const ACTION_CASES = {
        add: 'addProperties',
}

src/transformStateWithClones.js
```diff
@@ -7,7 +7,26 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let stateCopy = { ...state };
```
DanilWeda, line 11: try not to use let in this case.

DanilWeda (changes requested):
Good job! 
Check some improvements, 
Many thanks!

DanilWeda (approved):
Great. 
Approved.

DanilWeda (changes requested):
Check test suites. 
You don't need to use construction for example  Object.defineProperty or Object.hasOwn. You can, but the best way use object.assigne and check the key property inside the object. Please think about a simpler realization.

avakiel:
> Good! Check test suites. You don't need to use construction for example Object.defineProperty or Object.hasOwn. You can, but the best way use object.assigne and check the key property inside the object. Please think about a simpler realization. Many thanks!

test

PR #3041, opened 2023-12-11
Develop

Manankin:
All tests passed, task done successfully.

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -17,28 +17,25 @@ function transformStateWithClones(state, actions) {
         for (const key in item.extraData) {
           stateCopy[key] = item.extraData[key];
         };
-        // пушимо через деструктурізацію, зберігаючи модифікований об'єкт
-        result.push({ ...stateCopy });
         break;
 
       case 'removeProperties':
         for (const key of item.keysToRemove) {
           delete stateCopy[key];
         };
-        // пушимо через деструктурізацію, зберігаючи модифікований об'єкт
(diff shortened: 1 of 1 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        // пушимо через деструктурізацію
+        result.push({ ...stateCopy });
+        break;
+
+      default: break;
```
Xilston, line 40: Good practice is to return a default value, such as 'Unknown action'.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+  const result = [];
+
+  for (const item of actions) {
+    switch (item.type) {
+      case 'addProperties':
+        for (const key in item.extraData) {
+          stateCopy[key] = item.extraData[key];
+        };
+        // пушимо через деструктурізацію, зберігаючи модифікований об'єкт
+        result.push({ ...stateCopy });
```
Xilston, line 21: Here and in the next cases you repeat the same line ```result.push({ ...stateCopy }); ``` - you can do it after case block. Checklist 4. [DONT REPEAT YOURSELF] - If you perform same action in all `switch` cases - do it just once afterwards.

Xilston (changes requested):
Nice one! Good job!

YegorKorenkov (approved):
Good job! Please check one remark

PR #3048, opened 2023-12-12
add task solution

sasholk:
[checklist](https://github.com/mate-academy/js_stateful-clones/blob/master/checklist.md)

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -13,7 +13,7 @@ function transformStateWithClones(state, actions) {
     clear: 'clear',
   };
   const result = [];
-  let stateCopy = { ...state };
+  const stateCopy = { ...state };
 
   for (const action of actions) {
     const { type, extraData, keysToRemove } = action;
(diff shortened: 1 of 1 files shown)


avakiel:
Nice job!

EdenParadise (approved):
Looks solid! I like that!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let stateCopy = { ...state };
```
DanilWeda, line 11: Try to not use let in this case.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```
DanilWeda, line 15: Add cases in const.
For example const ACTION_TYPES = {
add: 'addProperties',
}

DanilWeda (changes requested):
Good job!
Take a look at some comments

SerhiiKirik (approved):
Nice 🔥

PR #3049, opened 2023-12-12
add solution

kostrubitskii:


src/transformStateWithClones.js
```diff
@@ -7,7 +8,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let newObject = { ...state };
```
DanilWeda, line 11: Try not to use let in this case. Search for another way of object assignment.

src/transformStateWithClones.js
```diff
@@ -7,7 +8,39 @@
...
+  const result = [];
+
+  actions.forEach(element => {
+    switch (element.type) {
+      case 'addProperties':
```
DanilWeda, line 16: Describe all cases in constant. 
For example:
const ACTION_CASES = {
  add: 'addProperties',
}

DanilWeda (changes requested):
Good job! 
Check some comments for improvement)

src/transformStateWithClones.js
```diff
@@ -7,7 +8,44 @@
...
+  let newObject = { ...state };
+  const result = [];
+  const ACTION_CASES = {
+    add: 'addProperties',
+    remove: 'removeProperties',
+    clear: 'clear',
+  };
+
+  actions.forEach(element => {
+    switch (element.type) {
```
olya-shyian, line 20: in order to make it more readable you can use destructuring :

`actions.forEach({ type, extraData, keysToRemove } => { `
 `const { add, remove, clear } = ACTION_CASES;`

olya-shyian (changes requested):
Nicely done!😊

DanilWeda (changes requested):
Please check the last my comment and fix it. Use Object.assign

DanilWeda (approved):
Great! 
Approved! 
Thanks!

PR #3050, opened 2023-12-12
add task solution

HeySleare:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [];
+  let tempClone = cloneObj(state);
+
+  for (const obj of actions) {
```
MaxKalalashnyk02, line 13: const action will be more appropriate name than obj

MaxKalalashnyk02 (changes requested):
Great work! Make a little improvement

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [];
+  let tempClone = cloneObj(state);
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        tempClone = Object.assign({}, tempClone, action.extraData);
+        states.push(cloneObj(tempClone));
```
SerhiiKirik, line 17: why do you need to `cloneObj` in each case?
You repeat yourself 3 times
What if we do it after your  `switch / case` ?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+
+      default:
+        break;
+    }
+  }
+
+  return states;
+}
+
+function cloneObj(obj) {
```
SerhiiKirik, line 40: `clone` or `copy`?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [];
+  let tempClone = cloneObj(state);
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        tempClone = Object.assign({}, tempClone, action.extraData);
```
SerhiiKirik, line 16: Can you do it in another way but with `Object.assign`? Try it

SerhiiKirik (changes requested):
Well done 🔥 
LEt's make some improvements 🧑‍💻

YegorKorenkov (changes requested):
Hey!
Please correct the errors that the mentor wrote about in the previous review. If you have any questions, feel free to ask them in the chat!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [];
+  let tempClone = cloneObj(state);
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        tempClone = Object.assign(tempClone, action.extraData);
```
YegorKorenkov, line 16: You can use cloneObj function in this case if you add first parametr with default value = {}

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        break;
+    }
+
+    states.push(cloneObj(tempClone));
+  }
+
+  return states;
+}
+
+function cloneObj(obj) {
```
YegorKorenkov, line 39: copyObj will be more correct name for this function

YegorKorenkov (approved):
Good job! Please check my notes in comments!

PR #3053, opened 2023-12-12
Solution

Yur1yBubl1k:


yaneznayou (approved):
Good job

src/transformStateWithClones.js
```diff
@@ -4,10 +4,38 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const finalArr = [];
+  const newState = Object.assign({}, state);
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(newState, action.extraData);
+        finalArr.push({ ...newState });
```
SerhiiKirik, line 17: Can you do it just one time? after `switch / case`?

src/transformStateWithClones.js
```diff
@@ -4,10 +4,38 @@
  * @param {Object} state
  * @param {Object[]} actions
  *
- * @return {Object[]}
+ * @return {Object[]}\
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const finalArr = [];
```
SerhiiKirik, line 10: It would be better
```suggestion
  const result = [];
```

SerhiiKirik (changes requested):
Well done 🔥 
Don't forget about DRY, let's fix it :D

src/transformStateWithClones.js
```diff
@@ -4,10 +4,36 @@
...
  *
- * @return {Object[]}
+ * @return {Object[]}\
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const newState = Object.assign({}, state);
+
+  for (const action of actions) {
```
olya-shyian, line 13: try to use destructuring  here
`const { type, extraData, keysToRemove } = action;`

use these variables instead

olya-shyian (changes requested):
I like it! Just a small update🙂

DanilWeda (approved):
Great job! 
Thanks for the request! 
Approved!

PR #3060, opened 2023-12-13
solution

KirillStetsenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+  for (const action of actions) {
+    const type = action.type;
+    let nextState;
+
+    /* Here I checked action condition for every case
+    and implement what I need to do
+    проверяю состояние действия для каждого случая
+    и выполняю действия в зависимости от условия */
+
+    if (type === 'addProperties') {
```
DanilWeda, line 23: Describe all variants of cases in constant.
const ACTION_CASES = {
add: 'addProperties',
}

src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  // Here I clone the given state / создаю объект текущего состояния
+  let currentState = { ...state };
```
DanilWeda, line 12: Try not to use let in this case (Mean search is another way to transform the currentState object).

src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+
+    if (type === 'addProperties') {
+      nextState = {
+        ...currentState, ...action.extraData,
+      };
+    }
+
+    if (type === 'removeProperties') {
+      nextState = { ...currentState };
+      action.keysToRemove.forEach((key) => delete nextState[key]);
```
DanilWeda, line 31: Try not to use forEach

DanilWeda (changes requested):
Thanks for the good job! 
Check some improvements.

DanilWeda (changes requested):
Check previous comments. They are not fixed.

PR #3065, opened 2023-12-14
Solution

yepolotn1ak:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      }
+    } else if (action.type === 'removeProperties') {
+      for (const keyRemove of action.keysToRemove) {
+        delete newState[keyRemove];
+      }
+    } else if (action.type === 'clear') {
+      for (const key of Object.keys(newState)) {
+        delete newState[key];
+      }
+    }
```
anastasiia-tilikina, line 28: replace nested if else with switch case structure

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resMass = [];
```
anastasiia-tilikina, line 10: it's not a good practice to name variables like this since it is not self-descriptive.
It's better to use something like result or resultArray or stateVersions or clonesArray etc.

anastasiia-tilikina (changes requested):
well done! 
just a couple of fixes

l4st1m0za (approved):
Great job!

PR #3072, opened 2023-12-18
state clone has been done

u9on:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const initialStateCopy = JSON.parse(JSON.stringify(state));
```
Viktor-Kost, line 10: why do you need to parse and stringify this object? You can just use `spread` oerator
```suggestion
  const initialStateCopy = {...state};
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+
+  const resultArray = [];
+
+  const applyAction = (currentState, action) => {
+    const currentStateCopy = JSON.parse(JSON.stringify(currentState));
```
Viktor-Kost, line 15: this is redundant
```suggestion
```

Viktor-Kost (changes requested):

Andriy-Shymkiv (approved):

PR #3076, opened 2023-12-19
Solution

LadyLiza:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
+    const action = actions[i];
+    let newState = { ...currentState };
+
+    if (action.type === 'addProperties') {
```
maxim2310, line 17: better to use switch case here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
```
maxim2310, line 13: why do you need index here, need to use for... of

src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+      }
+    }
+
+    if (action.type === 'removeProperties') {
+      const keysToRemove = action.keysToRemove;
+
+      for (let j = 0; j < keysToRemove.length; j++) {
+        const key = keysToRemove[j];
+
+        if (!(newState[key] === undefined)) {
```
maxim2310, line 29: don't need because if there will not be a key its just will not deleted

maxim2310 (changes requested):

nazarmatsevych (approved):
🔥

PR #3079, opened 2023-12-19
The task is completed

BoGoBooM:
I completed the assignment, it was interesting, but I'm kicking myself for often looking for a solution to a problem on the Internet, including chatGPT, instead of sitting and thinking for hours about the task and figuring out how to do it and how it would be better...


etojeDenys (approved):
good job

PR #3085, opened 2023-12-25
add solution

Diliradon:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+  let stateCopy = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      stateCopy = {
+        ...stateCopy,
+        ...action.extraData,
+      };
+
+      stateArray.push({ ...stateCopy });
```
maxim2310, line 20: dont repeat it in every case, move it after switch case

maxim2310 (changes requested):

src/transformStateWithClones.js
etojeDenys: image
still not fixed

etojeDenys (changes requested):
let's use switch/case instead of the multiple if else

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateArray = [];
+  let stateCopy = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```
volodymyr-soltys97, line 14: Use a destruction for `action`
```suggestion
  for (const action of actions) {
    const { type, .......... } = action;
    
    switch (type) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+      case 'addProperties':
```
volodymyr-soltys97, line 15: Create a variable for these words and use it here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+        stateCopy = {
+          ...stateCopy,
+          ...action.extraData,
+        };
```
volodymyr-soltys97, line 19: You can use Object.assign method here

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

Viktor-Kost (changes requested):
.

volodymyr-soltys97 (approved):
Great work 🔥

PR #3101, opened 2024-01-03
transformStateWithClones

Yaroslav65:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let currentState = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    let newState = { ...currentState };
```
Viktor-Kost, line 14: redundant
```suggestion
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(newState, action.extraData);
+        result.push(newState);
```
Viktor-Kost, line 19: you shouldn't push at each case. use it after switch
```suggestion
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete newState[key];
+        }
+        result.push(newState);
```
Viktor-Kost, line 26:
```suggestion
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+
+      case 'clear':
+        newState = {};
+        result.push(newState);
```
Viktor-Kost, line 31:
```suggestion
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+    }
+
+    currentState = { ...newState };
```
Viktor-Kost, line 35:
```suggestion
    result.push({ ... currentState });
```

Viktor-Kost (changes requested):

src/transformStateWithClones.js
Yaroslav65: redundant variable?

Viktor-Kost: yes, since I've written below, that you can just use it at `push`
https://github.com/mate-academy/js_stateful-clones/pull/3101/files#r1440814282

Viktor-Kost (changes requested):

src/transformStateWithClones.js
Yaroslav65: exactly!!)) that's right, thanks for the hint, I didn't notice it right away, it simplified the code)))

lerastarynets (approved):
You're on the right track!

PR #3107, opened 2024-01-06
solve task

Kredinec:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+      for (const key of action.keysToRemove) {
+        delete newState[key];
+      }
+      result.push(newState);
+    }
+
+    if (action.type === 'clear') {
+      result.push({});
+    }
```
volodymyr-soltys97, line 35: Use a switch case here instead of if for better readability

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
volodymyr-soltys97, line 13: Use a destruction for `action`
```suggestion
  for (const action of actions) {
    const { type, ........ } = action;
  
    if (type === 'addProperties') {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+      const newState = result.length
+        ? { ...result[result.length - 1] } : { ...state };
+
+      for (const key in action.extraData) {
+        newState[key] = action.extraData[key];
+      }
+      result.push(newState);
+    }
+
+    if (action.type === 'removeProperties') {
```
volodymyr-soltys97, line 23: Move this word 'removeProperties' to variable and use it here, it will be better

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

l4st1m0za (approved):
Good job here!

PR #3109, opened 2024-01-07
Transform State with Clones function | JS

julia-dronova:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
-  // write code here
+  const newState = { ...state };
+  const allStates = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(newState, action.extraData);
+
+        allStates.push({ ...newState });
```
etojeDenys, line 18: you can push only one time after switch/case

etojeDenys (changes requested):
small comment

etojeDenys (approved):
good job

PR #3111, opened 2024-01-07
Solution

lLiashko:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,28 @@
...
+      };
+    } else if (action.type === 'removeProperties') {
+      currentState = { ...currentState };
+
+      for (const key of action.keysToRemove) {
+        delete currentState[key];
+      }
+    } else if (action.type === 'clear') {
+      currentState = {};
+    }
```
etojeDenys, line 28: better to use switch/case instead of the multiple if/else

etojeDenys (changes requested):

lerastarynets (approved):
You're making excellent progress!

PR #3115, opened 2024-01-08
Develop

Beliyandr:
My first version!)

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,7 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  const newState = Object.assign({}, state);
+  const stateCopy = Object.assign({}, state);
   const allSteps = [];
 
   function doAction(modState, types, action) {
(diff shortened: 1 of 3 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,51 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = Object.assign({}, state);
+  const allSteps = [];
+
+  function doAction(modState, types, action) {
+    switch (types) {
+      case 'addProperties':
+        addProperties(modState, action.extraData);
+        allSteps.push({ ...modState });
```
etojeDenys, line 17: you can push only one time after switch/case

etojeDenys (changes requested):

src/transformStateWithClones.js
Beliyandr: if I understand, I fixed it !)

volodymyr-soltys97 (approved):
Good job 👍

PR #3117, opened 2024-01-09
add task solution

leravlasova:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+      for (const key of action.keysToRemove) {
+        if (stateCopy.hasOwnProperty(key)) {
+          delete stateCopy[key];
+        }
+      }
+    }
 
+    if (action.type === 'clear') {
+      stateCopy = {};
+    }
```
etojeDenys, line 28: let's use switch/case instead of the multiple if/else

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #3121, opened 2024-01-10
feat: provides with solution

kkrasnoshchok:


src/transformStateWithClones.js
```diff
@@ -3,11 +3,36 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClones = [];
+  let stateCopy = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        stateClones.push({ ...stateCopy });
```
maxim2310, line 16: this action you are doing in all cases, you could move it after switch case

maxim2310 (changes requested):

src/transformStateWithClones.js
kkrasnoshchok: sure, my bad. Fixing it now, thanks a lot

src/transformStateWithClones.js
```diff
@@ -3,11 +3,34 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+      case 'removeProperties':
+        const filteredEntries = Object.entries(stateCopy).filter(
+          ([key]) => !action.keysToRemove.includes(key)
+        );
+
+        stateCopy = Object.fromEntries(filteredEntries);
```
etojeDenys, line 22: you can just write the "for in" loop

etojeDenys (changes requested):
try to simplify it

src/transformStateWithClones.js
kkrasnoshchok: did the update, thanks for suggestion

lerastarynets (approved):
lgtm 🚀

PR #3122, opened 2024-01-10
transformStateWithClones implemented

Greshch:
Please check it.Thank`s

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -8,6 +8,51 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const ACTION_ADD = 'addProperties';
+  const ACTION_REM = 'removeProperties';
+  const ACTION_CLR = 'clear';
+  const copyState = [];
+  let obj = { ...state };
+
+  for (let i = 0; i < actions.length; ++i) {
+    const curOper = actions[i].type;
+
+    if (i > 0) {
+      obj = { ...copyState[i - 1] };
+    }
+
+    switch (curOper) {
+      case ACTION_ADD:
+        const extraData = actions[i].extraData;
+
+        for (const key in extraData) {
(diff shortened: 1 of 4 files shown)


src/transformStateWithClones.js
```diff
@@ -8,6 +8,51 @@
...
+
+    switch (curOper) {
+      case ACTION_ADD:
+        const extraData = actions[i].extraData;
+
+        for (const key in extraData) {
+          obj[key.toString()] = extraData[key.toString()];
+        }
+
+        copyState.push(obj);
```
etojeDenys, line 32: you can push only one time after all cases

src/transformStateWithClones.js
```diff
@@ -8,6 +8,51 @@
...
+  const ACTION_CLR = 'clear';
+  const copyState = [];
+  let obj = { ...state };
+
+  for (let i = 0; i < actions.length; ++i) {
+    const curOper = actions[i].type;
+
+    if (i > 0) {
+      obj = { ...copyState[i - 1] };
+    }
```
etojeDenys, line 22: looks redundant

src/transformStateWithClones.js
etojeDenys, line 32: we can just copy the object here
```suggestion
        copyState.push({...obj});
```

etojeDenys (changes requested):

src/transformStateWithClones.js
Greshch: I've just done it

src/transformStateWithClones.js
Greshch: I have no idea how to fix it now. But I agree with you.

src/transformStateWithClones.js
Greshch: I've done it.

PR #3124, opened 2024-01-12
add task solution

Marina-Semipyatnaya:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,47 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let currentState = { ...state };
+
+  const resultArray = [];
+
+  for (const action of actions) {
+    let newState = { ...currentState };
```
volodymyr-soltys97, line 15: Do you need to copy state again?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,47 @@
...
+
+    switch (action.type) {
```
volodymyr-soltys97, line 17: Use a destruction for `action`
```suggestion
  for (const action of actions) {
    let newState = { ...currentState };
    const { type, .......... } = action;

    switch (type) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,47 @@
...
+      case 'addProperties':
+
+        if (action.extraData) {
+          Object.assign(newState, action.extraData);
+        }
+        break;
+
+      case 'removeProperties':
```
volodymyr-soltys97, line 25: Consider will be better if you create the variable for this word and use it here

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code.

etojeDenys (approved):
lgtm

PR #3131, opened 2024-01-15
Demian Zolotukhin Add solution

DemianZank:
Add solution

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,47 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = {};
+
+  Object.assign(stateCopy, state);
+
+  const resultStateCopy = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        addProperties(stateCopy, action.extraData);
+        break;
+
+      case 'removeProperties':
+        removeProperties(stateCopy, action.keysToRemove);
+        break;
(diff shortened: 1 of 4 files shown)


DemianZank:
Всі юзають рест оператори, я його ніяк не освою

src/transformStateWithClones.js
```diff
@@ -7,7 +7,47 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = {};
+
```
PolinaRomanenko, line 11: You don't need to create an empty object and then use Object.assign. Instead, you can directly create a shallow copy of the initial state using the spread operator (...).
  let stateCopy = { ...state };

src/transformStateWithClones.js
```diff
@@ -7,7 +7,47 @@
...
 
+function removeProperties(stateCopy, keysToRemove) {
+  for (const key of keysToRemove) {
+    delete stateCopy[key];
+  }
+}
+
+function clearProperties(stateCopy) {
+  for (const key in stateCopy) {
+    delete stateCopy[key];
```
PolinaRomanenko, line 49: Instead of having separate functions for addProperties, removeProperties, and clearProperties, you can handle each case directly within the switch statement. This simplifies your code and avoids unnecessary function calls.

PolinaRomanenko (approved):
I approve, as there are no harsh mistakes, but please, take a look at my comments and try not to do the same mistakes again)

PR #3138, opened 2024-01-20
add task solution

Kapidastra:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in stateCopy) {
+        delete stateCopy[key];
+      }
+    }
+    result.push({ ...stateCopy });
+  }
```
anastasiia-tilikina, line 30: let's use switch case here

anastasiia-tilikina (changes requested):
Well done!
Just one thing to improve

Esceype (changes requested):
I'm sorry for the double rejection, pay attention only to the previous comment.

VitaliyBondarenko1982 (approved):
GJ!

PR #3140, opened 2024-01-20
add task solution

AnastasiaChaikovskaya:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    if (type === 'addProperties') {
+      Object.assign(tempObj, extraData);
+
+      const resultAdd = { ...tempObj };
+
+      stateHistory.push(resultAdd);
```
etojeDenys, line 21: you can move it out of the cases and write only one time

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  const tempObj = { ...state };
+  const stateHistory = [];
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    if (type === 'addProperties') {
```
etojeDenys, line 16: better to use switch case here instead of multiple if/else

etojeDenys (changes requested):

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Awesome work!

PR #3148, opened 2024-01-23
Solution

rwlp:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,30 @@
...
+      case 'clear':
+        for (const t in arrayHistory[i]) {
+          delete arrayHistory[i][t];
+        }
+        break;
+    }
+    arrayHistory.push({ ...arrayHistory[i] });
+  }
+
+  arrayHistory.pop();
```
EdPirro, line 32: You could change your logic to copy the previous object before applying the changes and then pushing instead of pushing and then changing it to avoid this `.pop()`

EdPirro (approved):
Awesome work, @rwlp! Clean code and good implementation, keep it up! 🚀

PR #3154, opened 2024-01-25
add a solution

ShvetsOlha:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = { ...state };
```
witflash, line 11: In general, you don't need this additional variable.
You can use inside your cycle the next way:
```
newState = {
  ...state,
  ..extraData,
}
// etc.
```
and just push directly the newState:
```
result.push(newState);
```

It would be the same but without additional variables.

witflash (approved):
Good work!

Pay attention to my comment for your future work

PR #3156, opened 2024-01-25
solution

arturgorniak:


src/transformStateWithClones.js
```diff
@@ -7,7 +9,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateVersions = [];
+
+  actions.reduce((currentState, action) => {
+    let nextState;
+
```
volodymyr-soltys97, line 16: Use the destruction for `action`
```suggestion
  actions.reduce((currentState, action) => {
    let nextState;
    const { type, ........ } = action; 
```

src/transformStateWithClones.js
```diff
@@ -7,7 +9,36 @@
...
+    switch (action.type) {
+      case 'addProperties':
```
volodymyr-soltys97, line 18: Consider will be better if you create the variables for strings and use it here

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/transformStateWithClones.js
arturgorniak: Could u please explain why i should use it? I dont really understand why it will be better after changes you sugested

PR #3158, opened 2024-01-26
solution

KaterinaZhlobinskaya:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,25 @@
...
+
+    if (action.type === 'addProperties') {
+      Object.assign(clone, currentState, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      for (const key in currentState) {
+        if (!action.keysToRemove.includes(key)) {
+          clone[key] = currentState[key];
+        }
+      }
+    }
```
nazarmatsevych, line 24: Good work but let's rewrite it to the switch/case statement since we rely on a single action.type value

nazarmatsevych (changes requested):
If you have any questions - feel free to ask them in fe_chat

src/transformStateWithClones.js
KaterinaZhlobinskaya: rewrote

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const clones = [];
+
+  for (const action of actions) {
```
loralevitska, line 12: Destructure `action`

loralevitska (approved):
Well done)

PR #3162, opened 2024-01-27
Solution(stateful-clones)

dkolinska:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+      for (const key of action.keysToRemove) {
+        delete nextState[key];
+      }
+      result.push(nextState);
+      currentState = nextState;
+    } else if (action.type === 'clear') {
+      result.push({});
+      currentState = {};
+    }
+  }
```
nazarmatsevych, line 33: Good work but let's rewrite it to switch/case statement since we rely on a single action.type value

nazarmatsevych (changes requested):

etojeDenys (approved):
lgtm

PR #3163, opened 2024-01-27
task accomplished

dmkopylets:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateHistory = [];
+
+  for (const action of actions) {
```
volodymyr-soltys97, line 12: Use destruction for `action`
```suggestion
  for (const action of actions) {
    const { type, ...... } = action;
  
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,41 @@
...
+
+    if (action.type === 'clear') {
+      newState = {};
+    } else {
+      const lastState = stateHistory[stateHistory.length - 1];
+
+      newState = JSON.parse(JSON.stringify(lastState || state));
+    }
+
+    switch (action.type) {
```
volodymyr-soltys97, line 23: Now the code is hard to read with these conditions, try using only `switch`

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

PR #3165, opened 2024-01-28
Solution

Yehorf21:
Completed the task

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -6,8 +6,34 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  let newState = { ...state };
+  const output = [];
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      newState = {
+        ...newState,
+        ...action.extraData,
+      };
+      output.push({ ...newState });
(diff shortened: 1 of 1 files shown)


src/transformStateWithClones.js
```diff
@@ -6,8 +6,34 @@
...
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  let newState = { ...state };
+  const output = [];
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
anastasiiavorobiova, line 15: Consider using a switch case here

src/transformStateWithClones.js
```diff
@@ -6,8 +6,34 @@
...
+      newState = {
+        ...newState,
+        ...action.extraData,
+      };
+      output.push({ ...newState });
```
anastasiiavorobiova, line 20: You can make a push once after the switch

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/transformStateWithClones.js
Yehorf21: I got so used to if / else if that I completely forgot about switch. Thank you

anastasiiavorobiova (approved):
Awesome work!

PR #3176, opened 2024-01-30
Solution

MOstapchyk:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,39 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const stateCopy = [];
+  const transformedState = { ...state };
+  const addProperties = 'addProperties';
+  const removeProperties = 'removeProperties';
+  const clear = 'clear';
```
Viktor-Kost, line 15: this variables is redundant here. You can use `enum` instead

Viktor-Kost (approved):
GJ!

src/transformStateWithClones.js
MOstapchyk: Ok. Thanks.

PR #3180, opened 2024-01-31
solution task transformStateWithClones(state, actions)

MikiMikiRage:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,30 @@
...
+let doubleCopyState = { ...copyState};
+stansAfterAction.push(doubleCopyState);
+}
+if (action.type === 'removeProperties') {
+  action.keysToRemove.forEach(key => delete copyState[key]);
+  let doubleCopyState = { ...copyState};
+  stansAfterAction.push(doubleCopyState);
 }
+if (action.type === 'clear') {
+  Object.keys(copyState).forEach(key => delete copyState[key]);
```
etojeDenys, line 26: you check in each if `action.type` values, better to use switch case for such cases

etojeDenys: and looks like you forgot to add spaces and tabulation

etojeDenys (changes requested):

src/transformStateWithClones.js
MikiMikiRage: ok, thx

etojeDenys (approved):
lgtm

PR #3185, opened 2024-02-01
add task solution

ShaoLei2:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,45 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const historyState = [];
+  const stateCopy = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        addProperty(stateCopy, action.extraData);
+        historyState.push({ ...stateCopy });
```
etojeDenys, line 17: you can move it after all cases and write only one time

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #3186, opened 2024-02-01
Solution js

vhrubyi:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,23 @@
...
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(cloneState, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      action.keysToRemove.forEach(key => {
+        delete cloneState[key];
+      });
+    } else if (action.type === 'clear') {
+      cloneState = {};
+    }
```
etojeDenys, line 22: you check `action.type` in each if, better to use switch/case in this case

etojeDenys (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
...
+      case 'addProperties':
+        addProperties(cloneState, action.extraData);
+        break;
+
+      case 'removeProperties':
+        removeProperties(cloneState, action.keysToRemove);
+        break;
+
+      case 'clear':
+        clearProperties(cloneState);
```
maxim2310, line 24:
```suggestion
        cloneState = {};
```

maxim2310 (approved):
Good work!!!

PR #3188, opened 2024-02-02
Develop

VerJane23:
Sorry, I first solved the task, then read the checklist and remade it in accordance with the conditions, that’s why there are two committees.


anastasiiavorobiova (approved):
Great work!

PR #3192, opened 2024-02-02
Solution

kravets111:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let NEXT_STATE = { ...state };
+  const RESULT = [];
```
Andriy-Shymkiv, line 11: you don't really need uppercase here:
```suggestion
  let stateCopy = { ...state };
  const result = [];
```

Andriy-Shymkiv (approved):

PR #3194, opened 2024-02-03
solution

mfcastilho:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,37 @@
...
+    }
+
+    if (action.type === 'addProperties') {
+      currentState = {
+        ...currentState,
+        ...action.extraData,
+      };
+    }
+
+    if (action.type === 'removeProperties') {
```
EdPirro, line 26: Use `else if` or early `return` to avoid unnecessary comparisons

EdPirro (approved):
Awesome job, @mfcastilho! Your code is clean and straightforward, it follows the requirements flawlessly! Keep it up! 🚀

PR #3199, opened 2024-02-05
js_stateful-clones solution

petrushinsky-kirill:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateHistory = [];
+  let stateCopy = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
Viktor-Kost, line 14: it's better to use `switch` statement here instead of `if`

Viktor-Kost (approved):
GJ!
Just apply a recommendation I left for you

PR #3201, opened 2024-02-06
created funciton for multiply actions with object

Zubyk-Yaroslav:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,37 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const prevState = { ...state };
+
+  for (const val of actions) {
+    switch (val.type) {
+      case 'addProperties':
+        Object.assign(prevState, val.extraData);
+        result.push({ ...prevState });
```
anastasiiavorobiova, line 24: You can push once after the switch case to avoid repeating code

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/transformStateWithClones.js
Zubyk-Yaroslav: Yes, you are right, thank you!

PR #3205, opened 2024-02-07
add task solution

valentynavik:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,68 @@
...
+}
+
+function clear(state) {
+  const newState = { ...state };
+
+  for (const key in newState) {
+    delete newState[key];
+  }
+
+  return newState;
```
maxim2310, line 71:
```suggestion
  return {};
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,68 @@
...
+
+  for (const key of keysToRemove) {
+    if (key in newState) {
+      delete newState[key];
+    }
+  }
+
+  if (Object.keys(newState).length === 0) {
+    return newState;
+  }
```
maxim2310, line 59: its redundant condition
```suggestion

```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,68 @@
...
+  const result = [];
+  let newState = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
+    const action = actions[i];
+
+    switch (action.type) {
+      case 'addProperties':
+        newState = addProperties(newState, action.extraData);
+        result.push(newState);
```
maxim2310, line 19: you doing this in every case, move it after switch case

maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

PR #3210, opened 2024-02-09
add task solution

Rusik-best86:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    let newState = { ...currentState };
```
anastasia1383, line 14: why do you need this variable?

anastasia1383 (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
roman-mirzoian, line 10: Don't use such generic naming,
```suggestion
  const clonedState = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+      case 'removeProperties':
+        currentState = { ...currentState };
+
+        for (const key of action.keysToRemove) {
+          if (currentState.hasOwnProperty(key)) {
+            delete currentState[key];
+          }
+        }
+        break;
+      default:
```
roman-mirzoian, line 32: Do not leave the default block empty.
For example, you can throw an error that the action is not correct.

roman-mirzoian (changes requested):

roman-mirzoian (approved):
Well done 👍

PR #3213, opened 2024-02-10
solution

Nastyahumeniuk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
...
+      currentState = {
+        ...currentState, ...action.extraData,
+      };
+    } else if (action.type === 'removeProperties') {
+      const keysToRemove = action.keysToRemove || [];
+
+      for (const key of keysToRemove) {
+        delete currentState[key];
+      }
+    }
```
VitaliyBondarenko1982, line 26: Try to rewrite if...else on switch..case

VitaliyBondarenko1982 (approved):
Good job.
~~Suggest some refactoring.~~
~~Please check comment.~~
Left comment but then accidently press approve. So can just read comment and refactor without re-request))

PR #3223, opened 2024-02-13
Solution

objelov:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,26 @@
...
+        delete cloneState[value];
+      }
+    } else if (action.type === 'clear') {
+      for (const key in cloneState) {
+        delete cloneState[key];
+      }
+    }
+
+    statesArray.push({ ...cloneState });
+  }
```
4ksandrey, line 28: if else construction isn't good
please rewrite your code with switch-case construction

4ksandrey (changes requested):
Check the comment

src/transformStateWithClones.js
objelov: Fixed

vlasiuksasha (approved):
well done

PR #3225, opened 2024-02-13
add

AntonSakh:
1. [CODE STYLE] - don't mutate object or arrays - it will cause unexpected results later on. You should make copy using `Object.assign` or `spread` operator
2. [CODE STYLE]: Use switch statement if you have limited amount of conditions.
4. [DONT REPEAT YOURSELF] - If you perform same action in all `switch` cases - do it just once afterwards.
5. [NAMING] - use proper names for object copy

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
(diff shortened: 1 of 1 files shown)


dima8528 (approved):
Looks good to me. Well done!

roman-mirzoian (approved):
Well done 👍

PR #3233, opened 2024-02-13
Clone state

DenNice-cloud:


Math8585 (changes requested):
In checklist:

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const STATE_HISTORY = [];
```
BudnikOleksii, line 10: It's not constant in general as you add values to it, so it should be in camelCase

BudnikOleksii (approved):
Good job 👍

PR #3234, opened 2024-02-13
add solution

yaGodya:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      case 'addProperties':
+        Object.assign(nextState, action.extraData);
+        break;
+      case 'removeProperties':
+        action.keysToRemove.forEach((key) => {
+          if (nextState.hasOwnProperty(key)) {
+            delete nextState[key];
+          }
+        });
+        break;
```
Nirvacsh, line 29: Better to use helper functions in switch structure (it's already not good for readability) => create clear, addProperty, removeProperty functions

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateHistory = [];
+  let currentState = { ...state };
```
Nirvacsh, line 11: Not a best practice to use let object / array

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+
+  actions.forEach((action) => {
+    let nextState = { ...currentState };
```
Nirvacsh, line 14: I think you can alter states only using one object

Nirvacsh (approved):
Good job, i wrote some comments for you to consider.

src/transformStateWithClones.js
dffUqp, line 14: I'm not sure that you need this variable. You already made a copy of state on 11 lines of code

src/transformStateWithClones.js
dffUqp: I agree with this comment

src/transformStateWithClones.js
dffUqp, line 29: use curly braces everywhere in switch construction, if you have more than 1 line of logic

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+
+    switch (action.type) {
```
dffUqp, line 16: We should always have `default` in our switch

dffUqp (changes requested):
Almost done🙂

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+      case 'addProperties': {
+        Object.assign(nextState, action.extraData);
+        break;
+      }
+
+      case 'removeProperties': {
+        action.keysToRemove.forEach((key) => {
+          if (nextState.hasOwnProperty(key)) {
+            delete nextState[key];
+          }
```
Moroz-Dmytro, line 31: You don't need this `if` here

Moroz-Dmytro (approved):
Good work
But check comment

PR #3238, opened 2024-02-14
Solution

Math8585:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newArr = [];
+  let newState = { ...state };
```
dffUqp, line 11: Let's rename these variables

```suggestion
  const stateHistory = [];
  let stateCopy = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+
+  for (let i = 0; i < actions.length; i++) {
```
dffUqp, line 13: We can use `for...of` here

dffUqp (changes requested):
Almost done🙂

roman-mirzoian (approved):
Well done 👍

PR #3239, opened 2024-02-14
solution v1

Nirvacsh:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        addProperties(copiedState, action.extraData);
+        break;
+      case 'removeProperties':
+        removeProperties(copiedState, action.keysToRemove);
+        break;
+      case 'clear':
+        clear(copiedState);
+        break;
```
HolyRaveOS, line 23: 3. [CheckList] switch/case should always have default case for error handling.
there have to be something like this 
default:
        throw new Error('action.type not found');

HolyRaveOS (changes requested):
Good job!
There is just one small thing to fix

roman-mirzoian (changes requested):
Well done 👍

roman-mirzoian (approved):

PR #3241, opened 2024-02-14
Solution

DmitryKot96:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,45 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copyState = {
+    ...state,
+  };
+
+  const history = [];
```
hattoriKanto, line 14: About variables naming. That is a comment from one mentor:

This variable should be named in camelCase
UPPER_CASE -> constants(as above, lines 31-35)
camelCase -> calculated value

So, I think you need to use  this type of vars - UPPER_CASE.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,45 @@
...
+};
+
+function removeProperties(copyState, keysToRemove) {
+  for (const key of keysToRemove) {
+    delete copyState[key];
+  }
+}
+
+function clear(copyState) {
+  Object.keys(copyState).forEach(key => delete copyState[key]);
```
hattoriKanto, line 48: If I remember there is a permission to use object's methods like map, forEach etc. But I don't know, it looks legit for me.

hattoriKanto (approved):
Nice, I have some suggestions, but good job.

src/transformStateWithClones.js
BudnikOleksii: > About variables naming. That is a comment from one mentor:
> 
> This variable should be named in camelCase UPPER_CASE -> constants(as above, lines 31-35) camelCase -> calculated value
> 
> So, I think you need to use this type of vars - UPPER_CASE.

It's not constant in this case, add dynamically new values to it. Constants are variables that never changes

src/transformStateWithClones.js
BudnikOleksii: > If I remember there is a permission to use object's methods like map, forEach etc. But I don't know, it looks legit for me.

It's ok

BudnikOleksii (approved):
Good work 👍

PR #3243, opened 2024-02-14
add task solution

IRONH4MMER:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,34 @@
...
+      res.push({ ...clonedState });
+    }
+
+    if (action.type === CLEAR) {
+      for (const key in clonedState) {
+        delete clonedState[key];
+      }
+      res.push({});
+    }
+  }
```
HolyRaveOS, line 36: 2. [CheckList]  Use switch statement if you have limited amount of conditions.

HolyRaveOS (changes requested):
Good job! But try to use switch...case as required in checkList.

PR #3245, opened 2024-02-14
Solution

hattoriKanto:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const CLONE_STATE = { ...state };
+  const HISTORY_STATE = [];
+
+  for (let i = 0; i < actions.length; i++) {
+    const CURRENT_ACTION = actions[i];
+    const ACTION_TYPE = CURRENT_ACTION.type;
```
BudnikOleksii, line 15: All these variables are not constants, so they should be in camelCase. Constants are variables that are static and never change.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const CLONE_STATE = { ...state };
+  const HISTORY_STATE = [];
+
+  for (let i = 0; i < actions.length; i++) {
```
BudnikOleksii, line 13: Use `for of` instead for better readability

BudnikOleksii (changes requested):
Good work 👍 
Just a few minor fixes are required

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const cloneState = { ...state };
+  const historyState = [];
+
+  for (const currentAction of actions) {
+    const actionType = currentAction.type;
+
+    switch (actionType) {
```
roman-mirzoian, line 16: image

roman-mirzoian (changes requested):

roman-mirzoian (approved):
Well done 👍

PR #3250, opened 2024-02-15
solved the task

ogerenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,45 @@
...
+
+      if (action[key] === 'removeProperties') {
+        removeProperties(stateClone, action.keysToRemove);
+        historyOfState.push({ ...stateClone });
+      }
+
+      if (action[key] === 'clear') {
+        clearProperties(stateClone);
+        historyOfState.push({ ...stateClone });
+      }
```
etojeDenys, line 28: better to use the switch case here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,45 @@
...
+
+      if (action[key] === 'removeProperties') {
+        removeProperties(stateClone, action.keysToRemove);
+        historyOfState.push({ ...stateClone });
+      }
+
+      if (action[key] === 'clear') {
+        clearProperties(stateClone);
+        historyOfState.push({ ...stateClone });
```
etojeDenys, line 27: you can write it only one time after all cases

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #3252, opened 2024-02-16
Solution

JustinNowoczek:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+  actions.forEach((actionObj) => {
+    switch (actionObj.type) {
+      case 'addProperties':
+        Object.assign(currentState, { ...actionObj.extraData });
+        break;
+      case 'removeProperties':
+        actionObj.keysToRemove.forEach((key) => {
+          delete currentState[key];
+        });
+        break;
```
choeqq, line 22: you're creating 2 loops - both in `removeProperties` and `clear`, I think it would be worth to create a separate function - just for deleting object fields, and then just pass the keys you want to delete

choeqq (changes requested):

TideTwo:
I'm not sure if I was supposed to pass the object that I modify into the helper function? Since I only ever work on 1 object, I referenced that one and defined the entire helper inside of the main function to give it access to the specific object.

PR #3259, opened 2024-02-18
Implemented a `transformStateWithClones` function 

YarynaPuhach:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
-}
+  const stateHistory = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    let nextState = { ...currentState };
+
+    switch (action.type) {
+      case 'clear':
+        nextState = {};
```
choeqq, line 18: this also works, but I think we would rather want to `delete` the fields from an object, so as you did in `removeProperties`, we don't really want to reassign empty object, we want to modify the existing one

choeqq (changes requested):

choeqq (changes requested):
-

asakevych (approved):
Good job! :rocket: :fire:

PR #3266, opened 2024-02-19
Solution

weswojciech:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
-  // write code here
+  const history = [];
+  const stateClone = Object.assign({}, state);
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(stateClone, action.extraData);
+      history.push({ ...stateClone });
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
```
TideTwo, line 18: I suggest, as was suggested to me by a mentor, that you create a function that can iterate over an array of keys, and delete them from your stateClone, so that you dont have to have 2 loops in both your `clear` and `removeProperties` action types.

TideTwo (changes requested):

src/transformStateWithClones.js
weswojciech: Could you help me understand how it's supposed to be done? To be honest I don't have an idea of what are you talking about.

TideTwo (approved):

asakevych (changes requested):
Good job! just check checklist once more :wink:

src/transformStateWithClones.js
```diff
@@ -11,24 +11,30 @@ function transformStateWithClones(state, actions) {
...
+      case 'clear':
+        const keys = Object.keys(stateClone);
 
-      iterate(keys);
-      history.push({ ...stateClone });
+        deleteKeysFromObj(keys);
+        history.push({ ...stateClone });
+        break;
+      default:
+        throw Error('Error');
```
jv-aquino, line 37: For better user experience (very important in real apps) you should always specify the error, in this case you could say something like "invalid action type"

jv-aquino (approved):
Good, clean code and  great naming conventions overall! Nice use for switch, it's the perfect place for one

PR #3267, opened 2024-02-19
add task solution

OksanaNastenko:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,30 @@
...
+    if (action.type === 'addProperties') {
+      for (const key in action.extraData) {
+        currentState[key] = action.extraData[key];
+      }
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete currentState[key];
+      }
+    } else if (action.type === 'clear') {
+      currentState = {};
```
etojeDenys, line 26: better to use switch/case here instead of multiple if/else

etojeDenys (changes requested):
small comment

src/transformStateWithClones.js
OksanaNastenko: done

l4st1m0za (approved):
Good job!

PR #3273, opened 2024-02-20
Develop

Maksym-Atanasov:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+  let stateClone = { ...state };
+
+  for (const action of actions) {
+    for (const par in action) {
+      switch (action[`${par}`]) {
+        case 'addProperties':
+          for (const data in action.extraData) {
+            stateClone[`${data}`] = action.extraData[`${data}`];
+          }
+          stateHistory.push({ ...stateClone });
```
etojeDenys, line 20: you can write it only one time, after all cases

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateHistory = [];
+  let stateClone = { ...state };
+
+  for (const action of actions) {
+    for (const par in action) {
```
etojeDenys, line 14: you don't need this loop, just use `action.type` or `action.extraData`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+      switch (action[`${par}`]) {
+        case 'addProperties':
+          for (const data in action.extraData) {
+            stateClone[`${data}`] = action.extraData[`${data}`];
```
etojeDenys, line 18:
```suggestion
          for (const data in action.extraData) {
            stateClone[data] = action.extraData[data];
```

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #3277, opened 2024-02-21
Develop

marinatea:


pawelkloc (approved):
It's looking good. Great work! :)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateHistory = [];
+  let CURRENT_STATE = { ...state };
```
choeqq, line 11: it's not a constant value like (40 or 50), please use `camelCase` 
```suggestion
  const CURRENT_STATE = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
+        CURRENT_STATE = {};
```
choeqq, line 16: please use `delete` instead of reassigning new object

choeqq (changes requested):

PR #3282, opened 2024-02-22
solution

sonyasapsan:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,45 @@
...
+  let index = 0;
+  let copyState;
+
+  for (const action of actions) {
+    if (index === 0) {
+      copyState = { ...state };
+    } else {
+      copyState = { ...stateResults[index - 1] };
+    }
+    index++;
```
sarakhmen, line 21: it's much easier to use some temp variable for example 'tempState' and operate with it. But then at the end of `for loop` you can add to the stateResults a copy of the tempState

src/transformStateWithClones.js
```diff
@@ -6,8 +6,45 @@
...
+
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(copyState, action.extraData);
+        break;
+      case 'removeProperties':
+        copyState = deleteProperties(copyState, action.keysToRemove);
+        break;
+      case 'clear':
+        copyState = deleteProperties(copyState, Object.keys(copyState));
```
sarakhmen, line 31:
```suggestion
        copyState = {};
```

sarakhmen (changes requested):

PR #3291, opened 2024-02-23
Solution

Zima220410:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      }
+      resultOfStates.push({ ...newState });
+    }
+
+    if (actions[i].type === 'clear') {
+      for (const key in newState) {
+        delete newState[key];
+      }
+      resultOfStates.push({ ...newState });
+    }
```
l4st1m0za, line 31: This part is too complex, it would be better to maintain it with the switch/case

l4st1m0za (changes requested):
Good job, some refactoring needed

l4st1m0za (approved):
Good job!

PR #3300, opened 2024-02-26
Develop

valerahajko:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,40 @@
...
+          delete nextState[key];
+        }
+        break;
+
+      default:
+        nextState = currentState;
+    }
+
+    stateHistory.push({ ...nextState });
+    currentState = { ...nextState };
```
DarkMistyRoom, line 41: no need to use `nextState`. You can modify `currentState` and push its copy to the array

DarkMistyRoom (approved):
Good job! Only one suggestion

PR #3303, opened 2024-02-27
solved task

KatsuboDmytro:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+
+  actions.forEach((value, index) => {
+    const newState = { ...newArr[index] };
+
+    switch (value.type) {
+      case 'addProperties':
+        for (const key in value.extraData) {
+          newState[key] = value.extraData[key];
+        }
+        newArr.push(newState);
```
etojeDenys, line 20: you can write it only one time after all cases

etojeDenys (approved):
good job

PR #3305, opened 2024-02-28
stateful clones 

svyatoslavPy:
# Cloning and transforming state

Write a `transformStateWithClones` function that takes a `state` object and an `actions` array,
applies each action to the previos `state` to calculate the next state,
and returns an array with states recieved after each action.


lerastarynets (approved):
Keep up the great work!

PR #3307, opened 2024-02-28
Solution

lishvets:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+      for (const key of action.keysToRemove) {
+        delete copy[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in copy) {
+        delete copy[key];
+      }
+    }
```
etojeDenys, line 28: better to use switch/case instead of multiple if/else

etojeDenys (changes requested):

OleziO (approved):

PR #3316, opened 2024-03-01
solution added

happytravels38:
https://github.com/happytravels38/js_stateful-clones.git

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -8,7 +8,7 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
-  let listOfStates = [];
+  const listOfStates = [];
   let newState = { ...state };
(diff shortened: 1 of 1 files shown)


src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
...
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          delete newState[key];
+        }
+        listOfStates.push({ ...newState });
+        break;
+
+      case 'clear':
+        newState = {};
+        listOfStates.push({ ...newState });
```
sTorba24, line 32: .push action is the same for every case, so move it out of switch case and use only once. DRY

sTorba24 (approved):
Good work! Pay attention to my comment

PR #3317, opened 2024-03-02
added solution

willwildred:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const copy = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```
volodymyr-soltys97, line 14: You can use destruction for `action`
```suggestion
  for (const action of actions) {
    const { type, .......... } = action;
  
    switch (type) {
```

volodymyr-soltys97 (approved):
Good job 👍

PR #3319, opened 2024-03-03
solution task

boikoua:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      for (const elem in action.extraData) {
```
anastasiiavorobiova, line 15: It's better to use Object.assign() here instead of the cycle

anastasiiavorobiova (approved):
Great work!

PR #3322, opened 2024-03-05
solutionV1

kgGugala:


yevhenii-pyl (approved):
Good job! 👍 

To make it even cleaner you can: 

1. Create an object to store action types to prevent typos and make it easier to extend, like:
```
const actionTypes = {
   clear: 'clear',
   ...
}
```

2. Create helper functions for each action type, like:
```
function handleAddProperties(state, data) {...}
```

PR #3325, opened 2024-03-06
over the task

true-binar-White-Rabbit:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,56 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const FINAL_ARR = [];
+  let MOD_OBJ = { ...state };
```
etojeDenys, line 11: we only name constants in uppercase, ordinary vars should be named in camelCase

src/transformStateWithClones.js
```diff
@@ -7,7 +7,56 @@
...
+
+  for (const obj of actions) {
+    switch (obj.type) {
+      case 'addProperties':
+        MOD_OBJ = addProps(MOD_OBJ, obj.extraData);
+        FINAL_ARR.push(MOD_OBJ);
```
etojeDenys, line 17: we can push it only one time after all cases

etojeDenys (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,54 @@
...
+function addProps(data, extraData) {
+  const OBJ = { ...data };
+
+  Object.assign(OBJ, extraData);
+
+  return OBJ;
+}
+
+function removeProps(data, keyToRemove) {
+  const OBJ = { ...data };
```
TarasHoliuk, line 41: image

It meant **really** constant values, not values defined using `const` keyword - FYI

TarasHoliuk (approved):

PR #3332, opened 2024-03-08
Add solution

olafchuszno:


yevhenii-pyl (approved):
Good job 👍 
As the only possible improvement, action types can be extracted into constant to avoid typos, like:

```
const actions = Object.freeze({
  'addProperties': 'addProperties'
})
```

Also, as a shortcut to Object.assign you can use spread operator { ...state } with the same result, but it's mostly up to you.

PR #3333, opened 2024-03-09
Solution

antonmolchanov18:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const res = [];
+  let obj = Object.assign({}, state);
+
+  for (const action of actions) {
+    obj = Object.assign({}, obj, action.extraData);
+
+    if (action.type === 'addProperties') {
+      res.push(obj);
```
Viktor-Kost, line 17: 1. push after conditions
2. just use either Object.assign or spread operator here
```suggestion
      Object.assign(obj, extraData)
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const res = [];
+  let obj = Object.assign({}, state);
+
+  for (const action of actions) {
+    obj = Object.assign({}, obj, action.extraData);
```
Viktor-Kost, line 14: redundant. You'd better use object destructing
```suggestion
    const { type, extraData, keysToRemove } = action;
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+
+    if (action.type === 'addProperties') {
```
Viktor-Kost, line 16: use switch case instead

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+        delete obj[removeElement];
+      }
+
+      res.push(obj);
+    }
+
+    if (action.type === 'clear') {
+      obj = {};
+      res.push(obj);
+    }
```
Viktor-Kost, line 31: so after condition you can use `push`
```
    res.push({ ... obj });
```

Viktor-Kost (changes requested):

src/transformStateWithClones.js
etojeDenys: image
still not fixed

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      case 'removeProperties':
+        for (const removeElement of keysToRemove) {
+          delete obj[removeElement];
+        }
+        res.push(Object.assign({}, obj));
+        break;
+
+      case 'clear':
+        obj = {};
+        res.push(Object.assign({}, obj));
```
etojeDenys, line 31: you can push only one time after all cases

etojeDenys (changes requested):

src/transformStateWithClones.js
etojeDenys: image
you should fix it everywhere

etojeDenys (changes requested):
please, fix all my comments

if you have some problems, then don't hesitate to ask for some help in the chat

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      case 'removeProperties':
+        for (const removeElement of keysToRemove) {
+          delete obj[removeElement];
+        }
+        res.push({ ...obj });
+        break;
+
+      case 'clear':
+        obj = {};
+        res.push({ ...obj });
```
maxim2310, line 31: yes, but you are doing it in every cases so move it after all cases by DRY principle

maxim2310 (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        break;
+
+      case 'clear':
+        obj = {};
+        pushClonedObject();
+        break;
+    }
+  }
+
+  return res;
```
maxim2310, line 40:
```suggestion
  for (const action of actions) {
    const { type, extraData, keysToRemove } = action;

    switch (type) {
      case 'addProperties':
        obj = { ...obj, ...extraData };
        break;

      case 'removeProperties':
        for (const removeElement of keysToRemove) {
          delete obj[removeElement];
        }
        break;

      case 'clear':
        obj = {};
        break;
    }
    res.push({ ...obj });
  }

  return res;
```

maxim2310 (changes requested):
better to do in this way

maxim2310 (approved):
well done

PR #3334, opened 2024-03-09
add solution

metrakonon:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,32 @@
...
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  let currentState = [{ ...state }];
+  const stateHistory = [];
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      currentState[0] = { ...currentState[0], ...action.extraData };
```
polosanya, line 16: You can use Object.assign instead

src/transformStateWithClones.js
```diff
@@ -6,8 +6,33 @@
...
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete currentState[key];
+        }
+        stateHistory.push({ ...currentState });
+        break;
+      case 'clear':
+        currentState = {};
+        stateHistory.push({ ...currentState });
```
polosanya, line 28: image

src/transformStateWithClones.js
```diff
@@ -6,8 +6,33 @@
...
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  let currentState = { ...state };
+  const stateHistory = [];
+
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 15: You can destructure all the needed props of action

polosanya (changes requested):
Almost perfect! 🙏 
A few small but important moments left

src/transformStateWithClones.js
```diff
@@ -6,8 +6,35 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  let currentState = { ...state };
```
DanilWeda, line 11: Also, you can try to use const in this case (optional), for objects and arrays better way not use reassigns)
Thanks)

DanilWeda (approved):
Great! 
Approved)

PR #3335, opened 2024-03-09
Working function

Pashahu1:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arr = [];
+  const copyState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const remove of action.keysToRemove) {
+          delete copyState[remove];
+        }
+        break;
+
+      case 'clear':
+        Object.keys(copyState).forEach(key => delete copyState[key]);
```
polosanya, line 26: `for` loop can be used instead

polosanya (approved):
Nice! 👍

PR #3339, opened 2024-03-11
add task solution

Flokuss:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,24 @@
...
+  for (const action of actions) {
+    if (action.type === 'clear') {
+      currentState = {};
+    } else if (action.type === 'addProperties') {
+      currentState = { ...currentState, ...action.extraData };
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete currentState[key];
+      }
+    }
```
nazarmatsevych, line 22: Let's rewrite it to switch/case since we rely on a single action value

nazarmatsevych (changes requested):
Almost done, please check the comment below

lerastarynets (approved):
Keep up the great work!

PR #3341, opened 2024-03-11
add solution

AnastasiiaCagle:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,34 @@
...
+
+    if (action.type === 'addProperties') {
+      nextState = { ...currentState, ...action.extraData };
+    }
+
+    if (action.type === 'removeProperties') {
+      nextState = { ...currentState };
+      for (const keyToRemove of action.keysToRemove) {
+        delete nextState[keyToRemove];
+      }
```
etojeDenys, line 29: better to use switch case here instead of the if else

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #3342, opened 2024-03-12
Solution

danil-maksimenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const transformStateResult = [];
+  const stateClone = { ...state };
+
+  for (const action of actions) {
+    if (action.hasOwnProperty('type') && action.type === 'clear') {
```
anastasiiavorobiova, line 14:
```suggestion
    if (action.type === 'clear') {
```
Consider moving strings 'clear', 'addProperties', etc. into variables. And it's better to use `switch` here

anastasiiavorobiova (changes requested):
Great work! Almost done!

lerastarynets (approved):
You're on the right track!

PR #3344, opened 2024-03-13
VB solution for transformStateWithClones

vbuldenko:


lerastarynets (approved):
You're doing brilliantly, keep it up!

PR #3345, opened 2024-03-13
add task solution

AlexandeeerG:

IMPORTANT! DON'T modify the initial state object in any way!

[CODE STYLE] - don't mutate object or arrays - it will cause unexpected results later on. You should make copy using Object.assign or spread operator

[DONT REPEAT YOURSELF] - If you perform same action in all switch cases - do it just once afterwards.

BAD EXAMPLE:

const copy = { ...state }

GOOD EXAMPLE:

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopies = [];
+  let newState = { ...state };
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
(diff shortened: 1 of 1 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+  const stateCopies = [];
+  let newState = { ...state };
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'clear':
+        newState = {};
+        stateCopies.push({ ...newState });
```
l4st1m0za, line 19: You can directly push the newState as it is already a new object.

l4st1m0za (approved):
Good job!

PR #3352, opened 2024-03-15
task solution

YaroslavHor:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+        delete currentState[key]; // Видаляємо ключ, якщо він існує
+      });
+    } else if (action.type === 'clear') {
+      // Очищуємо стан
+      currentState = {};
+    }
+    stateHistory.push(currentState);
+  });
+
+  return stateHistory;
```
etojeDenys, line 33: Remove comment. Try to write obvious code that explains all we need without comments)

etojeDenys (changes requested):

PR #3358, opened 2024-03-17
Solution

Allanpt:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,48 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const historyArray = [];
+  const copyState = { ...state };
+
+  for (const act in actions) {
+    const typeAction = Object.values(actions[act]);
+    const commandData = typeAction[1];
+
+    if (typeAction[0] === 'addProperties') {
```
EdPirro, line 23: You should simply use `actions[act].type` for the type and `actions[act].extraData` (or `actions[act].keysToRemove` if type is `removeProperties`) this is far more readable. Also your current implementation depends on the order of definition of the objects properties

src/transformStateWithClones.js
```diff
@@ -6,8 +6,48 @@
...
+      for (const data in commandData) {
+        copyState[data] = commandData[data];
+      }
```
EdPirro, line 26: You could use object spread here [Reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#:~:text=prototype.map().-,Spread%20in%20object%20literals,-Copying%20and%20merging)

EdPirro (changes requested):
Great job, @Allanpt! I suggested just two minor improvements; if you have any questions about them please ask in the qna channel!

src/transformStateWithClones.js
```diff
@@ -6,8 +6,44 @@
...
+      copyStateAndPush(copyState, historyArray);
+      continue;
+    }
+
+    if (actions[act].type === 'clear') {
+      copyState = {};
+      copyStateAndPush(copyState, historyArray);
+      continue;
+    }
+  }
```
EdPirro, line 44: Using a `switch/case` would be better than multiple `if`s on the same variable with well defined options

src/transformStateWithClones.js
```diff
@@ -6,8 +6,44 @@
...
+
+  arr.push(copyObj);
+}
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const historyArray = [];
+  let copyState = { ...state };
+
+  for (const act in actions) {
```
EdPirro, line 19: Could also use a `for...of` here since `act` is always only used to access `actions`

EdPirro (approved):
Great job, @Allanpt! 🚀 I suggested two optional changes, if you have the time, please try and implement them as I'm sure it'll be a valuable experience for future tasks

PR #3360, opened 2024-03-18
task solution using switch

MatheusStefan:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,33 @@
...
+        newState = {};
+        break;
+      case 'addProperties':
+        Object.assign(newState, action.extraData);
+        break;
+      case 'removeProperties':
+        if (Array.isArray(action.keysToRemove)) {
+          action.keysToRemove.forEach((key) => delete newState[key]);
+        }
+        break;
```
jv-aquino, line 30: you should also prevent errors when there's no action type / empty type

src/transformStateWithClones.js
```diff
@@ -6,8 +6,33 @@
...
+
+    switch (action.type) {
+      case 'clear':
+        newState = {};
+        break;
+      case 'addProperties':
+        Object.assign(newState, action.extraData);
+        break;
+      case 'removeProperties':
+        if (Array.isArray(action.keysToRemove)) {
```
jv-aquino, line 27: to avoid errors you should prevent the case where there's no property action.keysToRemove and also action.extraData

jv-aquino (changes requested):
Awesome job, for real!! Clean and concise js code, also nice use of switch and ternary operators! 

There's just a small thing missing that I wish you implemented and that is: **error handling**. It's not strictly necessary nor will make the tests fail but it would make your code safer and well-designed!

src/transformStateWithClones.js
```diff
@@ -6,8 +6,45 @@
...
+          }
+          break;
+        default:
+          throw new Error(`Unsupported action type: ${action.type}`);
+      }
+    } else {
+      throw new Error('Action type must be provided and non-empty!');
+    }
+
+    return [...history, { ...newState }];
```
pedro-ruas, line 46: This line here dismantles the history array and the newState object, here are some hints:
- *history* is the returning result, or accumulator, of your reduce. It doesn't need to be reassigned in this case, you could use the same array and use a `.push` function;
- *newState* is being dismantled into a new object that is being passed to the new history array. This is also not necessary, as this is the last line of your return, and newState will be recreated in the next reduce iteration, there's no need to also dismantle it here.

With this in mind, this line could be replaced for:
```
  history.push(newState);   //this adds newState to the history array without recreating any of them
  
  return history;    //this passes history to the next iteration of your reduce method without recreating it
```

pedro-ruas (approved):
Very nice use of the `.reduce` method! Great error handling, well structured code.
Awsome!

PR #3361, opened 2024-03-18
Solution

nnalinn:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+  const result = [];
+  let currentState = { ...state };
+
+  actions.forEach(action => {
+    switch (action.type) {
+      case 'addProperties':
+        currentState = {
+          ...currentState, ...action.extraData,
+        };
+        result.push({ ...currentState });
```
etojeDenys, line 19: you can push it only one time after cases

etojeDenys (changes requested):

Viktor-Kost (approved):
Nice!

PR #3367, opened 2024-03-20
Solution

HigorPedroso:


pedro-ruas (approved):
Great job! I really like the fact that you used const destructuring. You could also improve it by handling possible errors. 
Imagine if you get a different action type that wasn't defined, or no type at all. 
Or if the type is *addProperties*, but no *extraData* is passed. In a real application, it's good to think of those cases.
In any way, your code looks great!

PR #3371, opened 2024-03-21
adding solution to script

vinasxz:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+
+      case 'clear':
+        for (const key in newState) {
+          delete newState[key];
+        }
+        break;
+
+      default:
+        continue;
```
jv-aquino, line 32: nice vision for error handling! not in this case, as you did the right call, just a ui/ux thing to remember: when there's visual stuff generally try print the problem to the user

jv-aquino (approved):
Great job, clever and concise solution! Nice use of structures and methods, keep up this good job!

PR #3380, opened 2024-03-25
Solution

jkopan:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      for (const property of action.keysToRemove) {
+        delete transformedState[property];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in transformedState) {
+        delete transformedState[key];
+      }
+    }
```
etojeDenys, line 30: better to use a switch case here, 'cause you check everywhere action.type values

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #3384, opened 2024-03-25
Solution

ropanefarias:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let currentState = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    if (action.type === 'clear') {
```
EdPirro, line 14: Use a `switch/case` structure instead of multiple `if`s against the same variable with well defined values

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      currentState = {};
+    }
+
+    if (action.type === 'addProperties') {
```
EdPirro, line 18: Avoid impossibly true evaluations by using `else if` (would also be fixed if changed to `switch/case`)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      currentState = {
+        ...currentState,
+        ...action.extraData,
+      };
```
EdPirro, line 22: Could use `Object.assign` instead of spreading here (already creating a copy of the object before `.push`ing)

EdPirro (changes requested):
Awesome job, @ropanefarias! You're almost there 👏. The core logic of your solution seems to be great, I suggested just a few changes to make if flawless! 
If you have any questions, just let me know!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (currentState.hasOwnProperty(key)) {
+            delete currentState[key];
+          }
+        }
+        break;
+      default:
+        break;
```
jv-aquino, line 29: you could also add an error message here - it's a good practice to have error handling where you can, specially in our area of web dev and UI/UX

jv-aquino (approved):
Great job! With these changes you now have a very clean and modern code! Keep up your hard work!

PR #3385, opened 2024-03-26
Solution

yuliaPel:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,47 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const arrStates = [];
+
+  for (const action of actions) {
+    switch (action.type) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,47 @@
...
+        break;
+    }
+
+    arrStates.push({ ...stateCopy });
+  }
+
+  return arrStates;
+}
+
+function addProperties(state, extraData) {
```
polosanya, line 37: It is a good practice to create a function before use

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments

PR #3387, opened 2024-03-26
Solution

GrigoryyKach:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateHistory = [];
+  let stateCopy = { ...state };
```
metrakonon, line 11: You can try to use const in this case (optional), for objects and arrays better way not use reassigns)
Thanks)

metrakonon (changes requested):
Well done! But lets make some improvement.

metrakonon (approved):
Awesome!

mrkovaliv (approved):
Great job 👍 👍 👍

PR #3389, opened 2024-03-26
done with clones

sevenlays:


PrytulaAlexandr (approved):
Greate job!!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,46 @@
...
+function removeProperties(newState, keysToRemove) {
+  for (const key of keysToRemove) {
+    delete newState[key];
+  }
+}
+
+function clearProperties(newState) {
+  for (const k in newState) {
+    delete newState[k];
+  }
```
alexander-ignatow, line 49: It is better to move these functions up (declare them before using)

PR #3393, opened 2024-03-26
add solution

ordila:


TetianaVeremchuk (changes requested):
Good job! But according to checklist you need to use switch/case in this task😊

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in currentState) {
+          delete currentState[key];
+        }
+        break;
+      default:
+        throw new Error(`Invalid action type - ${action.type}`);
```
alexander-ignatow, line 32: 👍  Very nice
It is a good habit always handling the `default` case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete currentState[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in currentState) {
+          delete currentState[key];
```
alexander-ignatow, line 28: Minor thing - it would be more efficient to simply assign an empty object (but you will need to update your logic to do so)

alexander-ignatow (approved):
Looks good!

PR #3395, opened 2024-03-26
Solution

lmmaarriiaa:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
iAlfreX, line 10: A very general name, you can call it 'states'. If the function will be larger in size it may be a problem.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+      case 'clear':
+        stateClone = {};
+        break;
+
+      default:
+        break;
+    }
+
+    result.push({ ...stateClone });
```
iAlfreX, line 35: Just result.push(stateclone)

iAlfreX (approved):
Good Job!!!:)

src/transformStateWithClones.js
alexander-ignatow: Copying here makes sense, as `stateClone` can be modified on each iteration

alexander-ignatow (approved):
Good solution

Just one thing - please don't leave PR comments without any reaction, even if you don't accept them. It is always a good idea to leave a brief note with an explanation why you don't accept the comment

lmmaarriiaa:
> Good solution
> 
> Just one thing - please don't leave PR comments without any reaction, even if you don't accept them. It is always a good idea to leave a brief note with an explanation why you don't accept the comment

Hi! I was absent today, so I saw the comments only now( Thank you for the review!

PR #3396, opened 2024-03-26
Solution

TetianaVeremchuk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arrayClone = [];
+  const stateClone = { ...state };
+
+  for (const action of actions) {
```
Olek-91, line 13: you can use destructuring of objects:  `const { type, extraData, keysToRemove } = action;`

after switch case will be : 
'    switch (type) {
      case 'addProperties':
        Object.assign(stateClone, extraData);
        break;

      case 'removeProperties':
          delete stateClone[key];
        }
        break;

      case 'clear':
          delete stateClone[key];
        }
        break;

      default:
        break;
    }' 

or you can write this sorter (I known about this today at the lesson) '
  for (const { type, extraData, keysToRemove } of actions) {
    // const { type, extraData, keysToRemove } = action;'

Olek-91 (approved):

src/transformStateWithClones.js
Olek-91: GOOD JOB!!!!

Olek-91:
GOOD JOB!!!!

alexander-ignatow (approved):
Good! 👍

PR #3397, opened 2024-03-27
create transformStateWithClones func

iAlfreX:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,49 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [];
+  let stateCopy = { ...state };
```
polosanya, line 11: Do you really need to reassign it later? Consider using const instead of let

src/transformStateWithClones.js
```diff
@@ -7,7 +7,49 @@
...
+
+  for (const action of actions) {
+    stateCopy = { ...stateCopy };
+
+    switch (action.type) {
+      case 'addProperties':
+        addProperties(stateCopy, action.extraData);
```
polosanya, line 18: You can destructure all the needed props of action

src/transformStateWithClones.js
```diff
@@ -7,7 +7,49 @@
...
+        throw new Error('Action not found');
+    }
+
+    states.push(stateCopy);
+  }
+
+  return states;
+}
+
+function addProperties(state, extraData) {
```
polosanya, line 39: It is a good practice not to use functions before they were created

polosanya (changes requested):
Good job! Let's try to improve your code a bit)

iAlfreX:
Done :)

alexander-ignatow (approved):
👍

PR #3398, opened 2024-03-27
My solution

Bogdan-Pohilko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newObj = {};
+  const result = [];
+
+  Object.assign(newObj, state);
```
GUSILLUS, line 13: You can just write `const newObj = { ...state };`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+  for (let i = 0; i < actions.length; i++) {
+    const type = actions[i].type;
+
+    if (type === 'addProperties') {
+      const extraData = actions[i].extraData;
+
+      for (const key in extraData) {
+        newObj[key] = extraData[key];
+      }
+      result.push(Object.assign({}, newObj));
```
GUSILLUS, line 24: Just push one time at the end of the iteration after if block

GUSILLUS (changes requested):
GJ! But:

src/transformStateWithClones.js
Bogdan-Pohilko: agree

src/transformStateWithClones.js
Bogdan-Pohilko: if I make an output only once, the last two tests will not pass because they expect an array of objects

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+      const remove = actions[i].keysToRemove;
+
+      for (const key in remove) {
+        delete newObj[remove[key]];
+      }
+      // result.push(Object.assign({}, newObj));
+    } else if (type === 'clear') {
+      Object.keys(newObj).forEach((key) => delete newObj[key]);
+    }
+    result.push(Object.assign({}, newObj));
```
GUSILLUS, line 32: For future, you can re-write it like this or use a switch case
```suggestion
    if (type === 'addProperties') {
      const extraData = actions[i].extraData;

      for (const key in extraData) {
        newObj[key] = extraData[key];
      }
    } 
    
    if (type === 'removeProperties') {
      const remove = actions[i].keysToRemove;

      for (const key in remove) {
        delete newObj[remove[key]];
      }
      // result.push(Object.assign({}, newObj));
    } 
    
    if (type === 'clear') {
      Object.keys(newObj).forEach((key) => delete newObj[key]);
    }
    
    result.push(Object.assign({}, newObj));
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+      for (const key in extraData) {
+        newObj[key] = extraData[key];
+      }
+    } else if (type === 'removeProperties') {
+      const remove = actions[i].keysToRemove;
+
+      for (const key in remove) {
+        delete newObj[remove[key]];
+      }
+      // result.push(Object.assign({}, newObj));
```
GUSILLUS, line 28: Remove redundant comments

GUSILLUS (changes requested):
GJ! Lets fix this comment and i will approve the task

GUSILLUS (approved):
Approve

PR #3400, opened 2024-03-27
Solution

Follder:


D-Kondratiev (approved):
Nice work!

alexander-ignatow (approved):
Good job!

Minor thing - it is a good idea always have `default` in the `switch` block

Follder:
> Good job!
> 
> Minor thing - it is a good idea always have `default` in the `switch` block

Ok, thanks!

PR #3403, opened 2024-03-27
Add task solution

Esther-Guimaraes:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete currentState[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      currentState = {};
+    }
```
EdPirro, line 28: Use `switch/case` structure instead of multiple `if`s against the same variable with well defined values; (also, in situations such as these, adding a `default` fallback option for when the supplied `type` is invalid is a good practice)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      Object.assign(currentState, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete currentState[key];
+      }
+    }
+
+    if (action.type === 'clear') {
```
EdPirro, line 26: Avoid impossibly true evaluations with `else if` (In this case using `switch/case` with well positioned `break`s should do)

EdPirro (changes requested):
Great effort, @Esther-Guimaraes, you're almost there 👏! I suggested just one change that I believe will improve your solution. Please ask if you have any questions!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        break;
+      case 'clear':
+        currentState = {};
+        break;
+      default:
+        throw new Error('Tipo de ação inválido: ' + action.type);
+    }
+    result.push(currentState);
+    previousState = currentState;
+  }
```
DarkMistyRoom, line 33:
```suggestion
  let currentState = { ...state };

  for (const action of actions) {
    switch (action.type) {
      case 'addProperties':
        Object.assign(currentState, action.extraData);
        break;
      case 'removeProperties':
        for (const key of action.keysToRemove) {
          delete currentState[key];
        }
        break;
      case 'clear':
        currentState = {};
        break;
      default:
        throw new Error('Tipo de ação inválido: ' + action.type);
    }
    result.push({ ...currentState });
  }
```

DarkMistyRoom: You can do it just like this. It looks a bit simpler

DarkMistyRoom (approved):
Good job!

PR #3404, opened 2024-03-27
Solution

denyssheremeta:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateHistory = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    let newState = { ...currentState };
+
+    if (action.type === 'addProperties') {
```
stepanyukvova200, line 16: [CODE STYLE]: Use switch statement if you have limited amount of conditions.

stepanyukvova200 (changes requested):
Good job )
Please change 'if else' to 'switch case', so that the code will follow the correct style. 😉

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateHistory = [];
+  let currentState = { ...state };
```
undeletable, line 11: NOTE: this is not critical, just a suggestion for a small optimization.

You're making double copying of `state` object - on lines 11 and 14.
Copying is certainly needed on line 14 - because on previous iteration we don't do it (line 31). But on line 11, before the loop, we can just do `let currentState = state`.

undeletable (approved):
The solution looks nice! I just have a suggestion for a small optimization - please check the comment.

stepanyukvova200 (approved):

PR #3410, opened 2024-03-28
Solution

luyx0:


Voloshyn-Oleg (approved):
Good job! 
2. [Checklist] I would skip such a test, but it would be better to do it through the switch method

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+    } else if (action.type === 'addProperties') {
+      currentState = { ...currentState, ...action.extraData };
+
+    } else if (action.type === 'removeProperties') {
+
+      const keysToRemove = action.keysToRemove || [];
+      for (const key of keysToRemove) {
+        delete currentState[key];
+      }
+    }
```
alexander-ignatow, line 28: `else / if` works here, but `switch / case` would be way simpler and readable solution
Please rewrite it with a switch

alexander-ignatow (changes requested):
Good, but some refactor is needed

undeletable (approved):
Well done!

PR #3411, opened 2024-03-28
Solution

my-git-username:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+    if (action.type === 'addProperties') {
+      Object.assign(stateClone, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const rKey of action.keysToRemove) {
+        delete stateClone[rKey];
+      }
+    }
+
```
etojeDenys, line 29: better to use switch case here instead of the multiple if else

etojeDenys (changes requested):

PR #3412, opened 2024-03-28
Develop

yanamomot:


alexander-ignatow (changes requested):
Good 👍 

One thing - please add `default` case to the `switch` block and handle it

alexander-ignatow (changes requested):
Good 👍

One thing - please add `default` case to the `switch` block and handle it

src/transformStateWithClones.js
```diff
@@ -6,8 +6,50 @@
...
+        stateHistory.push(Object.assign({}, newObj));
+        break;
+
+      case 'clear':
+        clearProperties(newObj);
+        stateHistory.push(Object.assign({}, newObj));
+        break;
+
+      default:
+        return 'some error';
```
undeletable, line 32: If action's type is not one of known ones (`addProperties`, `removeProperties`, `clear`), it makes sense to return the current state itself - so that the next actions would work correctly.

undeletable (approved):
Well done! I just have a small comment about handling edge case - please consider that during your future work.

src/transformStateWithClones.js
```diff
@@ -6,8 +6,50 @@
...
+    }
+  }
+
+  return stateHistory;
+}
+
+function addProperties(newObj, extraData) {
```
undeletable, line 39: Defining separate functions for each action - well done!

PR #3415, opened 2024-03-29
add task solution

tarassamchenko:


alexander-ignatow (changes requested):
Please add `default` case to the `switch` and handle it

undeletable (approved):
Well done

PR #3417, opened 2024-03-29
Solution

Voloshyn-Oleg:


odunchyk (commented):
Good job!
LGTM

MarchAlbion:
GJ

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+      case 'addProperties':
+        Object.assign(stateClone, extraData);
+        break;
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          delete stateClone[key];
+        }
+        break;
+      case 'clear':
+        for (const key in stateClone) {
```
ArtemZakhar, line 26: I have made the same mistake: took the code from the previous task.
And was told that this part should be written with an empty object.
Smth like this: stateClone = {};

But still, the job is done great!)

ArtemZakhar (approved):
Good job)
Have a nice evening!)

src/transformStateWithClones.js
alexander-ignatow: Sounds reasonable, but probably not critical here

alexander-ignatow (approved):
👍

PR #3418, opened 2024-03-30
Solution

Pidsvynok:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      case action.type === 'removeProperties':
+        for (const keyToRemove of action.keysToRemove) {
+          delete newState[keyToRemove];
+        }
+        break;
+
+      case action.type === 'clear':
+        newState = {};
+        break;
+    }
```
vadiimvooo, line 28:
```suggestion
    switch (action.type) {
      case 'addProperties':
        Object.assign(newState, action.extraData);
        break;

      case 'removeProperties':
        for (const keyToRemove of action.keysToRemove) {
          delete newState[keyToRemove];
        }
        break;

      case 'clear':
        newState = {};
        break;
    }
```

vadiimvooo (changes requested):
Almost done. Just one change required:

anastasiiavorobiova (approved):
Awesome!

PR #3422, opened 2024-04-02
Solution

oleksandr-kovalchuk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      for (const key of action.keysToRemove) {
+        delete stateCopy[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in stateCopy) {
+        delete stateCopy[key];
+      }
+    }
```
anastasiia-tilikina, line 28: let's use switch case structure instead of if here

anastasiia-tilikina (changes requested):
GJ

just one comment

vadiimvooo (approved):
LGTM

PR #3423, opened 2024-04-02
Solution

Markelko:


djtyrf312 (approved):
I am not sure that it is your solution. But you study for yourself)

PR #3434, opened 2024-04-05
solution

SerhiiYarenko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateHistory = [Object.assign({}, state)];
+
+  for (const action of actions) {
+    const newState = Object.assign({}, stateHistory[stateHistory.length - 1]);
+
+    switch (action.type) {
```
```suggestion
    const { type, ............ } = action;

    switch (type) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+      case 'clear':
```
volodymyr-soltys97, line 16: Consider will be better if you create a variable for words

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

anastasiiavorobiova (approved):
Great work!

PR #3443, opened 2024-04-10
js_stateful-clones SOLVED

IryB-k:
Hi, here is the solution. 
Thanks a lot!


anastasiia-tilikina (changes requested):
Good job!

let's just make one small improvement)

etojeDenys (approved):
awesome

PR #3450, opened 2024-04-12
added task solution

mariamiborul:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,31 @@
...
+
+    if (action.type === 'addProperties') {
+      Object.assign(stateCopy, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete stateCopy[key];
+      }
+    }
```
etojeDenys, line 29: better to rewrite it using the switch/case instead of the if/else

etojeDenys (changes requested):

src/transformStateWithClones.js
mariamiborul: done

anastasiiavorobiova (approved):
Awesome work!

PR #3451, opened 2024-04-12
Solution

StepanenkoA:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      }
+    } else if (action.type === 'addProperties') {
+      Object.assign(newState, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      action.keysToRemove.forEach((key) => {
+        if (newState.hasOwnProperty(key)) {
+          delete newState[key];
+        }
+      });
+    }
```
andriy-shymkiv, line 28: looks like it should work, but for better developer experience I'd rather use switch/case instead

andriy-shymkiv (changes requested):

lerastarynets (approved):
lgtm 🚀

PR #3452, opened 2024-04-12
added solution

syavaYki:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+        newState[key] = val;
+      }
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete newState[key];
+      }
+    } else if (action.type === 'clear') {
+      if (Object.keys(state).length < 1) {
+        return {};
+      }
```
volodymyr-soltys97, line 25: Will be better if you use switch case here instead of if else

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [];
+  const newState = { ...state };
+
+  actions.forEach((action) => {
+    if (action.type === 'addProperties') {
```
```suggestion
    const { type, ....... } = action;

    if (type === 'addProperties') {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

volodymyr-soltys97 (changes requested):
.

andriy-shymkiv (approved):

PR #3453, opened 2024-04-12
Added solution

vadym-valchak:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,52 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let stateCopy = Object.assign({}, state);
+  const stateHistory = [];
+
+  for (const action of actions) {
+    switch (action.type) {
```
```suggestion
  for (const action of actions) {
    const { type, ........ } = action;
    
    switch (type) {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/transformStateWithClones.js
vadym-valchak: Fixed

vadiimvooo (approved):
Well done!)

PR #3456, opened 2024-04-15
Solution

hatevadym:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,30 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const stateHistory = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```
```suggestion
  for (const action of actions) {
    const { type, ........ } = action;
    
    switch (type) {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

volodymyr-soltys97 (approved):
Well done! 🔥

PR #3458, opened 2024-04-16
add task solution with few problems

pakhomovalex:
pr for Q&A JS

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -8,6 +8,33 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  let stateCopy = { ...state };
+  const finalArray = [];
+
+  actions.forEach((action) => {
+    if (action.type === 'addProperties') {
+      stateCopy = Object.assign(stateCopy, action.extraData);
+      finalArray.push(stateCopy);
+    }
+
+    if (action.type === 'removeProperties') {
+      action.keysToRemove.forEach((key) => {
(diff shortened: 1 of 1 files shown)


src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
...
+      finalArray.push(stateCopy);
+    }
+
+    if (action.type === 'clear') {
+      for (const property in stateCopy) {
+        delete stateCopy[property];
+      }
+
+      finalArray.push(stateCopy);
+    }
```
DarkMistyRoom, line 34: If you have predetermined amount of options it is better to use switch/case structure instead
of multiple if/else blocks

src/transformStateWithClones.js
```diff
@@ -8,6 +8,33 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  let stateCopy = { ...state };
+  const finalArray = [];
+
+  actions.forEach((action) => {
+    if (action.type === 'addProperties') {
+      stateCopy = Object.assign(stateCopy, action.extraData);
+      finalArray.push(stateCopy);
```
DarkMistyRoom, line 17: no need to `push` after each case. you can do it once at the end of each iteration

DarkMistyRoom: and you should push the copy of stateCopy if you don't want to get an array with the same object on all positions

DarkMistyRoom (changes requested):

PR #3478, opened 2024-04-24
Solution

ashatskyy:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,41 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const rez = [];
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'clear') {
+      rez.push({});
+    }
+
+    if (actions[i].type === 'addProperties' && rez.length > 0) {
```
anastasiiavorobiova, line 18:
```suggestion
    if (actions[i].type === 'addProperties') {
```

anastasiiavorobiova: Consider using the switch here

anastasiiavorobiova: Also, you can use a helper object to change it instead of checking the size of the array

anastasiiavorobiova (changes requested):
Great work! Almost done

lerastarynets (approved):
Keep up the great work!

PR #3496, opened 2024-05-01
Stateful Clones

kozlovoleksii:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
...
+
+    if (obj.type === 'removeProperties') {
+      for (const keyToRemove of obj.keysToRemove) {
+        delete newObj[keyToRemove];
+      }
+    }
+
+    if (obj.type === 'clear') {
+      newObj = {};
+    }
```
etojeDenys, line 26: better to use switch/case here

etojeDenys (changes requested):

etojeDenys (approved):
amazing

PR #3500, opened 2024-05-04
GG Mate academy

Teslaanazuma:

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let localState = { ...state };
+  const result = [];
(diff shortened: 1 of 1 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let localState = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
```
```suggestion
  for (const action of actions) {
    const { type, ...... } = action;
    
    switch (type) {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

l4st1m0za (approved):
Good job!

PR #3503, opened 2024-05-07
add task solution

yursprava:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,26 @@
...
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
etojeDenys, line 27: better to use the switch/case here instead of multiple if/else

src/transformStateWithClones.js
```diff
@@ -7,7 +7,26 @@
...
+  }
+  return state;
```
etojeDenys, line 29: please, read the description of this task, you need to return the array

etojeDenys (changes requested):
feel free to ask for some help in the chat

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let result = [];
```
volodymyr-soltys97, line 10:
```suggestion
  const result = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+
+  for (const action of actions) {
+    const newState = structuredClone(state);
+
+    switch (action.type) {
```
```suggestion
  for (const action of actions) {
    const newState = structuredClone(state);
    const { type, .......... } = action;

    switch (type) {
```

volodymyr-soltys97 (changes requested):
Almost done!
Let's make your code better

vadiimvooo (approved):
LGTM

PR #3512, opened 2024-05-09
Solution

Alexejklimov:
Add some solution. I hope it`s not so bad)

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const result = [];
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
+      for (const key in actions[i].extraData) {
+        newState[key] = actions[i].extraData[key];
+      }
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      for (const key of actions[i].keysToRemove) {
+        delete newState[key];
+      }
+    }
+
+    if (actions[i].type === 'clear') {
+      const arrKeys = Object.keys(newState);
+
+      for (const key of arrKeys) {
+        delete newState[key];
+      }
+    }
+    result.push({ ...newState });
+  }
+
+  return result;
 }
 
 module.exports = transformStateWithClones;


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      }
+    }
+
+    if (actions[i].type === 'clear') {
+      const arrKeys = Object.keys(newState);
+
+      for (const key of arrKeys) {
+        delete newState[key];
+      }
+    }
```
volodymyr-soltys97, line 32: Rewrite this conditions with switch case construction, it will be better

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const result = [];
+
+  for (let i = 0; i < actions.length; i++) {
```
volodymyr-soltys97, line 13: Indexes are not needed here, so it will be better to use the for of loop.
Also use destruction for `action` after
```suggestion
  for (const action of actions) {
    const { type, ........... } = action;
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
```
```suggestion
  for (const action of actions) {
    const { type, extraData, keysToRemove} = action;

    switch (type) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+      case 'addProperties':
+        for (const key in action.extraData) {
+          newState[key] = action.extraData[key];
```
volodymyr-soltys97, line 17:
```suggestion
        for (const key in extraData) {
          newState[key] = extraData[key];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
```
volodymyr-soltys97, line 22:
```suggestion
        for (const key of keysToRemove) {
```

volodymyr-soltys97 (changes requested):
Almost done!
Let's improve your code

volodymyr-soltys97 (approved):
Great! 🔥

PR #3516, opened 2024-05-10
Solution

SSDoizha:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,24 @@
...
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(stateClone, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete stateClone[key];
+      }
+    } else {
+      stateClone = {};
+    }
```
volodymyr-soltys97, line 23: Use a switch case construction instead of if else, it will be better

src/transformStateWithClones.js
```diff
@@ -8,6 +8,24 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  let stateClone = { ...state };
+  const stateHistory = [];
+
+  for (const action of actions) {
```
```suggestion
  for (const action of actions) {
    const { type, ......... } = action;
  
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/transformStateWithClones.js
volodymyr-soltys97: Still not fixed from the previous review 
image

PR #3517, opened 2024-05-11
add task solution

hsvirina:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      Object.assign(resultObject, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      const keys = action.keysToRemove;
+
+      for (const key of keys) {
+        delete resultObject[key];
+      }
+    }
```
volodymyr-soltys97, line 31: Use the switch case construction instead of if conditions, it will be better

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultObject = { ...state };
+
+  const resultArray = [];
+
+  for (const action of actions) {
+    if (action.type === 'clear') {
```
```suggestion
  for (const action of actions) {
    const { type, ....... } = action;
  
    if (type === 'clear') {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

volodymyr-soltys97 (approved):
Well done 🔥

PR #3521, opened 2024-05-15
Solution  / js_stateful-clones

MariCodec:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const currentState = { ...state };
+  const stateHistory = [];
+
+  for (const obj of actions) {
+    if (obj.type === 'addProperties') {
```
dinaminahi, line 15: Let's use switch case instead of if statement here for better readability

src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
...
+      for (const date in obj.extraData) {
+        currentState[date] = obj.extraData[date];
+      }
```
dinaminahi, line 18: Think of how you can use here Object.assign instead of loop

src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const currentState = { ...state };
+  const stateHistory = [];
+
+  for (const obj of actions) {
```
dinaminahi, line 14: obj is not a clear name, let's think of better one

src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
...
+
+    if (obj.type === 'removeProperties') {
+      for (const date in obj.keysToRemove) {
+        for (const key in currentState) {
+          if (key === obj.keysToRemove[date]) {
+            delete currentState[key];
+          }
+        }
+      }
+    }
```
dinaminahi, line 29: Instead of nested loops for removing properties, you can use delete directly if the property exists.

dinaminahi (changes requested):
Good job 👍 
Let's improve it a bit

src/transformStateWithClones.js
MariCodec: there is more than one simple check in the code, is switch case more suitable than the if statement?

src/transformStateWithClones.js
MariCodec: thank you! Corrected

src/transformStateWithClones.js
```diff
@@ -8,6 +8,28 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
```
roman-mirzoian, line 10:
```suggestion
```

roman-mirzoian (approved):
Well done!

PR #3523, opened 2024-05-16
add soulution

HumenVAsya:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
```
```suggestion
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
...
+
+  const array = [];
```
roman-mirzoian, line 12: Not the best name for an array. 
The name of the variable to which an array is assigned should give a clear description of what the array holds.

roman-mirzoian (approved):
Well done, and please pay attention to the comment.

PR #3532, opened 2024-05-20
Develop

ikocherovets:
As you can see, I've written some comments after several lines, it is just an explanation for myself. 
I hope it won't bother you :)

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let stateAtTheMoment = { ...state }; // clone the state at the beginning
+  const stateHistory = []; // const, because an array is an object in JS.
+
+  for (const action of actions) {
+    let newState = { ...stateAtTheMoment };
+    // Clone the current state to the new state
+
+    if (action.type === 'clear') {
+      newState = {};
+    }
+
+    if (action.type === 'addProperties') {
+      Object.assign(newState, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete newState[key];
+      }
+    }
+
+    stateHistory.push(newState);
+    // push the new state to the empty array, where the result is collected
+
+    stateAtTheMoment = newState; // update the current state to the new state
+  }
+
+  return stateHistory;
 }
 
 module.exports = transformStateWithClones;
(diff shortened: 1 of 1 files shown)


FaiHamid (approved):
Well done!))

roman-mirzoian (changes requested):
Good work, but there is room for improvement:
1) Although the current approach works, using ``switch`` might be more efficient and readable.

2) Avoid reassigning stateAtTheMoment to newState at the end of each loop iteration.

Also pay attention to the AI comments.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
-  // write code here
+  let stateAtTheMoment = { ...state };
+  const stateHistory = [];
+
+  for (const action of actions) {
+    let stateCopy = { ...stateAtTheMoment };
+
+    switch (action.type) {
+      case 'clear':
+        stateCopy = {};
```
dinaminahi, line 18: Better way here will be or to use delete in loop for all properties or to use Object.assign to an empty object

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let stateAtTheMoment = { ...state };
+  const stateHistory = [];
+
+  for (const action of actions) {
+    let stateCopy = { ...stateAtTheMoment };
```
dinaminahi, line 14: Instead of creating 2 copies of state, let's keep one and when you are pushing it to the history use spread operator `stateHistory.push({...stateCopy})`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let stateAtTheMoment = { ...state };
```
dinaminahi, line 10:
```suggestion
  const stateCopy = { ...state };
```

dinaminahi (changes requested):
Good job 👍 But pay attention to the comments

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateHistory = [];
+  const stateCopy = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```

polosanya (approved):
Nice! 👍

PR #3540, opened 2024-05-21
made transformStateWithClones

StsKushnir:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+
+  for (const act of actions) {
+    switch (act.type) {
+      case 'addProperties':
+        Object.assign(newState, act.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const i of act.keysToRemove) {
+          delete newState[`${i}`];
```
roman-mirzoian, line 23: Deleting properties from an object does not require Template strings:
```suggestion
          delete newState[i];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
-  // write code here
+  const arrBack = [];
+  let newState = {};
+
+  Object.assign(newState, state);
+
+  for (const act of actions) {
+    switch (act.type) {
+      case 'addProperties':
+        Object.assign(newState, act.extraData);
```
roman-mirzoian, line 18: Using ``Object.assign(newState, state)`` works, but it's better to use destructuring to create a copy of the object:

roman-mirzoian (changes requested):
Good job, the code is generally correct, but there are a few things that could be improved.
Please pay attention to the comments:

1) The name ``arrBack`` is not very clear. Better to use a more descriptive name like ``stateHistory``.
  Instead of ``newState``, it is better to use ``currentState``, which more accurately reflects the purpose of the variable.

2) Handling default in switch: Better to add handling of unknown action types to generate an error.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
-}
+  const stateHistory = [];
+  let currentState = {};
+
+  Object.assign(currentState, state);
```
roman-mirzoian, line 13: You're already using Object.assign, but it's better to use destructuring to copy the initial state.

```suggestion
  let currentState = { ...state };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        }
+        break;
 
+      case 'clear':
+        currentState = {};
+        break;
+
+      default:
+        throw new Error();
+    }
```
roman-mirzoian, line 33: Adding an informative message to throw new Error() in the default case.
```suggestion
      default:
        throw new Error(`Unexpected action type: ${type}`);
    }
```

roman-mirzoian (approved):
Well done and please pay attention to this comments.

PR #3542, opened 2024-05-22
add task solution

Stasandrushchak:


OleksiiKnihin (approved):
Good job!

src/transformStateWithClones.js
```diff
@@ -8,6 +8,29 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
```
```suggestion
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,29 @@
...
+        currentState = {};
+        break;
+      case 'addProperties':
+        currentState = { ...currentState, ...action.extraData };
+        break;
+      case 'removeProperties':
+        action.keysToRemove.forEach((key) => delete currentState[key]);
+        break;
+      default:
+        break;
```
roman-mirzoian, line 27: In the case of ``default``, it is better to throw an error to handle unknown action types.

roman-mirzoian (changes requested):
Good job, but please pay attention to the comments.

PR #3544, opened 2024-05-22
Solution

sa4ok-1:


priemssonia (approved):
Great job:)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,47 @@
...
+  const newState = { ...state };
+
+  for (const key of keysToRemove) {
+    delete newState[key];
+  }
+
+  return newState;
 }
 
+function clearProperties(state) {
```
roman-mirzoian, line 50: ``clearProperties`` function does not need a ``state`` input parameter because it always returns an empty object.
```suggestion
function clearProperties() {
```

roman-mirzoian (approved):
Well done!

PR #3545, opened 2024-05-22
Solution

demkoveronika:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [];
+  let currentState = { ...state };
+
+  actions.forEach((action) => {
+    const nextState = { ...currentState };
```
dinaminahi, line 14: Instead of creating 2 copies of state, let's keep one and when you are pushing it to the states use spread operator states.push({...currentState})

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        break;
+
+      case 'clear':
+        for (const key in nextState) {
+          delete nextState[key];
+        }
+        break;
+
+      default:
+        break;
```
dinaminahi, line 34: Let's throw a descriptive error in default case

dinaminahi (changes requested):
Good job 👍 
I see that you deleted the default case from switch, let's return it and throw an error there telling that such action.type is invalid.
`throw new Error(`Invalid action type: ${action.type}`);`

src/transformStateWithClones.js
```diff
@@ -1,13 +1,42 @@
...
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
   // write code here
+  const states = [];
+  const currentState = { ...state };
+
+  actions.forEach((action) => {
+    switch (action.type) {
```
polosanya, line 15: You can destructure all the needed props of action

polosanya (approved):
Well done! 👍

PR #3546, opened 2024-05-22
Develop

pushyshyn:


V-Shut (approved):
Well done! Easy to read and nothing extra...

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+        currentState = {};
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete currentState[key];
+        }
+        break;
+      case 'addProperties':
+        Object.assign(currentState, action.extraData);
+        break;
```
dinaminahi, line 25: Let's add default case for you switch case and throw there an error with description 
`    throw new Error('Action type is invalid: ${action.type}!');
`

dinaminahi (changes requested):
Good job 👍 
Let's add one improvement

roman-mirzoian (approved):
Well done!

PR #3547, opened 2024-05-22
add solution transformStateWithClones

OleksiiKnihin:


roman-mirzoian (approved):
Good job. 
Your solution is quite close to the correct one, but there are a few important notes:

1) In the case of default, it is better to throw an error so that the developer is immediately notified of an incorrect action type.

2) Order of adding states to history: A copy of the current state should be added to stateHistory only after each action is processed to avoid code duplication.

(You added a copy of the state to ``stateHistory`` inside each case block in the switch statement. This leads to code duplication.
To avoid this duplication, you can call stateHistory.push({ ...cloneState }) once after the switch statement completes.)

PR #3548, opened 2024-05-23
Solution

mhryvko:


pushyshyn (approved):
Well done

roman-mirzoian (changes requested):
Good job but pay attention to the AI comments.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+
```
roman-mirzoian, line 11: Do not leave extra empty lines.
```suggestion
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    let stateCopy = { ...currentState };
```
roman-mirzoian, line 15: You create a copy of the current state at the start of each iteration. 
To reduce code duplication, we do this after processing the action.
```suggestion
```
``result.push({ ...currentState });``

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
```
roman-mirzoian, line 19: Instead of using `Object.assign` to add properties, it is better to use destructuring ``{ ...currentState, ...action.extraData }``, which is a more convenient and modern approach.

```suggestion
        currentState = { ...currentState, ...action.extraData };
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
+        break;
+
+      case 'clear':
+        stateCopy = {};
+        break;
```
roman-mirzoian, line 30:
```suggestion
      case 'removeProperties':
        for (const key of action.keysToRemove) {
          delete currentState[key];
        }
        break;

      case 'clear':
        currentState = {};
        break;
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+
+      default:
+        throw new Error(`Unknown action type: ${action.type}`);
+    }
+
+    result.push(stateCopy);
+    currentState = stateCopy;
```
roman-mirzoian, line 37: Add a copy of the current state to the result after processing each action.

```suggestion
  result.push({ ...currentState });
```

roman-mirzoian (changes requested):
Good job the code looks good and does the job correctly.
Here are some tips to improve your code.

roman-mirzoian (approved):
Well done!

PR #3562, opened 2024-05-28
add solution

akryto:

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -1,5 +1,5 @@
 'use strict';
-
+/* eslint-disable */
 /**
  * @param {Object} state
  * @param {Object[]} actions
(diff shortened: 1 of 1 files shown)


roman-mirzoian (changes requested):
Good start but let's make the code even better.
To begin with, pay attention to the comments.

Let's change the structure of the logic and replace the ``for-loop`` with ``for-of`` and instead of ``if-else'' we go to ``switch case'', is will make the code more concise and understandable.

src/transformStateWithClones.js
```diff
@@ -1,13 +1,39 @@
...
+      case 'removeProperties':
+        for (const fieldName of action.keysToRemove) {
+          delete newState[fieldName];
+        }
+        break;
+      case 'clear':
+        newState = {};
+        break;
+      default:
+        break;
```
roman-mirzoian, line 30: `default`: present, but it does not perform any actions. You can add a throw to handle errors.
Example: 
```suggestion

      default:
        throw new Error(`Unknown action type: ${action.type}`);
```

src/transformStateWithClones.js
```diff
@@ -1,13 +1,39 @@
...
+          ...newState,
+          ...action.extraData,
+        };
+        break;
+      case 'removeProperties':
+        for (const fieldName of action.keysToRemove) {
+          delete newState[fieldName];
+        }
+        break;
+      case 'clear':
```
roman-mirzoian, line 26: Adding an empty line after `break` in a `switch` is not mandatory, but it can improve code readability.

```suggestion
        break;
        
      case 'clear':
```

roman-mirzoian (changes requested):
Good job, just a little work to do.

PR #3575, opened 2024-06-06
[Feature] added transform state function

Arthur-Tolkachov:
## Description
Added transform state function

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -17,7 +17,7 @@ function transformStateWithClones(state, actions) {
   const history = [];
 
   for (const { type, extraData, keysToRemove } of actions) {
-    const stateCopy = { ...(history.at(-1) ?? state) };
+    const stateCopy = { ...(history[history.length - 1] ?? state) };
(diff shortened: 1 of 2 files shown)


src/transformStateWithClones.js
```diff
@@ -6,8 +6,48 @@
...
+  clear: 'clear',
+};
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const history = [];
+
+  for (const { type, extraData, keysToRemove } of actions) {
+    const stateCopy = { ...(history[history.length - 1] ?? state) };
+    let historyItem = {};
```
maxim2310, line 21: it's unnecessary variable  
```suggestion
```

maxim2310 (approved):
Good work

src/transformStateWithClones.js
Arthur-Tolkachov: AI said: `Pushing the state copy to the history array should be done after the switch block to avoid repetition.`
So I just move my value to the historyItem variable. Which solution could be better?

PR #3586, opened 2024-06-12
Stateful clonesss

KaterynaHubko:


PolinaRomanenko (approved):
This is nice, the only thing I don't like is your namings of the variables: actionss, stateNeww, sttates - these duplications are a bit confusing) mind it for the future and try not to write such namings))

PR #3587, opened 2024-06-13
add solution task

OleglVOl:
add solution task


lerastarynets (approved):
You're on the right track!

PR #3590, opened 2024-06-14
Solution

jacube1:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  const stateCopy = {};
+
+  Object.assign(stateCopy, state);
+
+  for (const action of actions) {
+    switch (action.type) {
```
natalia-klonowska, line 16: switch always need to have default

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const j of action.keysToRemove) {
+          delete stateCopy[j];
+        }
+        break;
+
+      case 'clear':
+        for (const k in stateCopy) {
```
natalia-klonowska, line 28: rename j and k for example to key

natalia-klonowska (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+          delete stateCopy[rem];
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
+    }
```
Zibi95, line 32: There is still missing default case. Often its just an case where you throw an error as the value of the variable that you check is unexpected.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+      case 'removeProperties':
+        for (const rem of action.keysToRemove) {
+          delete stateCopy[rem];
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
```
Zibi95, line 30: This is simply the same as
```suggestion
      stateCopy = {};
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+  Object.assign(stateCopy, state);
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const rem of action.keysToRemove) {
```
Zibi95, line 22: give more meaningful variable names for example here it could be `keyToRemove`

Zibi95 (changes requested):
Still missing default case.

BogdanMaliuta (changes requested):
Not bad, but make some changes in your switch/case, which was described by another reviewers

Zibi95 (approved):

PR #3598, opened 2024-06-18
creat function transformStateWithClones

oliaShymanska:


lerastarynets (approved):
You're on the right track!

PR #3601, opened 2024-06-20
add task solution

Kirill1908:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const historyArray = [];
+  let changedObject = { ...state };
```
TarasHoliuk, line 11: Good comment, BTW 🙂 
Let's fix it as well - naming is a very important part of code

TarasHoliuk (changes requested):
One tiny fix required

TarasHoliuk (approved):

PR #3605, opened 2024-06-25
Solution

AnastasiiaKorolko:


etojeDenys (changes requested):
looks good, but better to use switch/case instead of multiple if/else

lerastarynets (approved):
Keep up the great work!

PR #3616, opened 2024-07-01
Solution

roman-logos-frontend:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,23 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let currentState = { ...state };
+  const statesHistory = [];
+
+  for (const action of actions) {
+    if (action.type === 'clear') {
```
anastasiiavorobiova, line 14: Could you consider using the switch statement here to improve the readability? Also, it's better to create variables instead of using magic values https://softwareengineering.stackexchange.com/questions/365339/what-is-wrong-with-magic-strings

anastasiiavorobiova (changes requested):
Great work! Almost done!

etojeDenys (approved):
awesome

PR #3617, opened 2024-07-01
Solution

ansnihur:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+
+        for (const key of action.keysToRemove) {
+          delete newState[key];
+        }
+        currentState = newState;
+        break;
+      case 'clear':
+        currentState = {};
+        break;
+    }
```
TarasHoliuk, line 29: it's good practice to have `default` case in a switch as is said in the previous comment

If you don't know what to do in default case you can throw an error - it's common way to handle it

for example:

```suggestion
    }
    
    default:
      throw new Error(`Unknown action type: ${action.type}`);
```

TarasHoliuk (changes requested):

SanyaBratashchuk (approved):

PR #3621, opened 2024-07-02
Solution

YaroslavKolesnyk:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const currentState = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
```
ArturErmolenko, line 14: use destruction for action

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in currentState) {
+          delete currentState[key];
+        }
+        break;
+    }
+
```
ArturErmolenko, line 33: switch case construction should always have default case

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const currentState = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          currentState[key] = action.extraData[key];
```
ArturErmolenko, line 17: will be great to use Object.assign here but it's not critical

ArturErmolenko (changes requested):
Looks good 🔥 
Let's improve a bit

VenomMB (changes requested):
please add correct solution

BudnikOleksii (approved):
Good work 👍

PR #3642, opened 2024-07-05
Task complete

Artem-Brui:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopyList = [];
+
+  const oneAction = (task) => {
+    const stateCopy =
+      stateCopyList[0] !== undefined
+        ? { ...stateCopyList[stateCopyList.length - 1] }
+        : { ...state };
```
volodymyr-soltys97, line 16:
```suggestion
  const stateCopyList = [];
  const stateCopy = { ...state };

  const oneAction = (task) => {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+    switch (task.type) {
```
volodymyr-soltys97, line 18: Use destructuring for task
```suggestion
   const { type, ...... } = task;

    switch (type) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopyList = [];
+
+  const oneAction = (task) => {
```
volodymyr-soltys97, line 12: Also, you don't need to create a new function, you can just use for of for actions

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+    const stateCopy =
+      stateCopyList[0] !== undefined
+        ? { ...stateCopyList[stateCopyList.length - 1] }
+        : { ...state };
+
+    switch (task.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, task.extraData);
```
volodymyr-soltys97, line 20:
```suggestion
        Object.assign(stateCopy, extraData);
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+      case 'removeProperties':
+        task.keysToRemove.forEach((key) => delete stateCopy[key]);
```
volodymyr-soltys97, line 23:
```suggestion
        keysToRemove.forEach((key) => delete stateCopy[key]);
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/transformStateWithClones.js
Artem-Brui: Thanks, solved

src/transformStateWithClones.js
Artem-Brui: Thanks, solved

src/transformStateWithClones.js
Artem-Brui: Thanks, solved

src/transformStateWithClones.js
Artem-Brui: Thanks, solved

src/transformStateWithClones.js
Artem-Brui: Thanks, solved, but works only inside the loop, otherway array has the same object that times that we have rules in actions

GUSILLUS (approved):
Sorry missclick

PR #3644, opened 2024-07-07
Created the function transform state with clones

r-prykhodko:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [];
+
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    const stateCopy = { ...currentState };
```
BudnikOleksii, line 15: `currentState` variable is enough, you don't need this variable

BudnikOleksii (changes requested):
Good work 👍 
Approved, but get rid of the unnecessary variable

BudnikOleksii (approved):
👍

PR #3652, opened 2024-07-08
add solution

Mazurok001:


coldshein (approved):
Well done 🥳

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateHistory = [];
+  let stateCopy = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```
ArturErmolenko, line 14: use destruction for action

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const i of action.keysToRemove) {
```
ArturErmolenko, line 20: you iterate over the keys, so why you name it `i` ?)) I believe `key `suits better

let's learn to give good names to variables ;)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+          }
+        }
+        break;
+
+      case 'clear':
+        stateCopy = {};
+        break;
+
+      default:
+        throw Error;
```
ArturErmolenko, line 32: will be great to show some info in this error,because now your error will justmake App down and you won't have any information about what happened
```suggestion
        throw new Error('some error message with some info');
```

ArturErmolenko (changes requested):
GJ 👍 
Let's improve a bit

VenomMB (approved):
Good job!

PR #3654, opened 2024-07-09
solution

annton-shakur:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        delete obj[key];
+      });
+      stateHistory.push({ ...obj });
+    }
+
+    if (action.type === 'clear') {
+      Object.keys(obj).forEach((key) => {
+        delete obj[key];
+      });
+      stateHistory.push({ ...obj });
```
SarakhmanFTMentor, line 32: could you please refactor the code so that reducing the code duplication?

SarakhmanFTMentor (changes requested):

PR #3660, opened 2024-07-10
task js_stateful-clones completed

mfirka:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,45 @@
...
+
+      case 'removeProperties':
+        let tempState = { ...newState };
+
+        for (const key of action.keysToRemove) {
+          const { [key]: unusedVariable, ...rest } = tempState;
+
+          tempState = rest;
+        }
+        newState = tempState;
```
anastasiiavorobiova, line 37: You can use a delete here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete

anastasiiavorobiova (changes requested):
Great work! Almost done!

etojeDenys (approved):
lgtm

PR #3661, opened 2024-07-10
Solution task

SashaPas05:


iyehorova (approved):
Hi! Good job!
Everything is fine, but you created an extra variable `newState`.  Just replace it with `currentState` and everything will work exactly the same way

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateHistory = [];
+
+  let currentState = { ...state };
+
+  actions.forEach((action) => {
+    let newState;
```
BudnikOleksii, line 15: You don't this variable, you can use `currentState` instead

BudnikOleksii (approved):
Good work 👍

PR #3676, opened 2024-07-16
Solution

renata-aline:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        for (const key in stateClone) {
+          delete stateClone[key];
+        }
+        break;
+
+      default:
+        break;
+    }
+
+    const newStateClone = { ...stateClone };
```
pedro-ruas, line 35: I believe you don't need this variable as you could simply push `stateClone`, because it's already has been created using the spread synthax

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateClone[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateClone) {
```
pedro-ruas, line 26: If `stateClone` were created using a `let` instead of a `const`, you could simply assign it to a new empty object instead of iterating all it's keys

pedro-ruas (changes requested):
Minimal changes regarding performance improvement

src/transformStateWithClones.test.js
```diff
@@ -39,6 +39,7 @@ test('Should create a new object with multiple added properties', () => {
     .toEqual({});
 });
 
+
```
gitlherme, line 42: You can remove this empty blank line.

gitlherme (approved):
Hello, @renata-aline!

A little point about the empty line, but great job with the code!

PR #3679, opened 2024-07-17
task solution

SofiSolf:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+      for (const key of action.keysToRemove) {
+        delete stateCopy[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in stateCopy) {
+        delete stateCopy[key];
+      }
+    }
```
etojeDenys, line 31: better to rewrite it using the switch/case instead of multiple if/else

etojeDenys (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
-  // write code here
+  const stateCopy = {};
+
+  Object.assign(stateCopy, state);
+
+  const modificationsHistory = [];
+
+  for (const action of actions) {
+    switch (true) {
+      case action.type === 'addProperties':
```
etojeDenys, line 18:
```suggestion
    switch (action.type) {
      case 'addProperties':
```

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #3695, opened 2024-07-23
Solution

Ilona-Gogol:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+      for (const key of action.keysToRemove) {
+        delete stateClone[key];
+      }
+      result.push({ ...stateClone });
+    }
+
+    if (action.type === 'clear') {
+      for (const key in stateClone) {
+        delete stateClone[key];
+      }
```
etojeDenys, line 29: better to use switch/case here instead of the multiple if statements

etojeDenys (changes requested):

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateClone = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateClone, action.extraData);
+        result.push({ ...stateClone });
```
etojeDenys, line 17: you can write it only one time after all cases

etojeDenys (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #3697, opened 2024-07-24
Solution

OlenaZiniak:
Write a `transformStateWithClones` function that takes a `state` object and an `actions` array,
applies each action to the previos `state` to calculate the next state,
and returns an array with states recieved after each action.


lerastarynets (approved):
lgtm 🚀

PR #3700, opened 2024-07-25
Solution

shodmon:


shodmon:
Had to add

```
global.structuredClone = val => {
  return JSON.parse(JSON.stringify(val))
}
```

due to _**structuredClone is not defined**_ error

PR #3710, opened 2024-07-28
last version

vikaruda:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+      currentState = { ...currentState };
+
+      for (const key of action.keysToRemove) {
+        delete currentState[key];
+      }
+    }
 
+    if (action.type === 'clear') {
+      currentState = {};
+    }
```
etojeDenys, line 28: rewrite it using the switch/case instead of the multiple if/else

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #3711, opened 2024-07-28
task done

misulkin:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+      for (const key of action.keysToRemove) {
+        delete copyState[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const prop in copyState) {
+        delete copyState[prop];
+      }
+    }
```
etojeDenys, line 28: better to rewrite this logic using the switch/case instead of the multiple if/else

etojeDenys (changes requested):

etojeDenys (changes requested):
my prev comment still not fixed, feel free to ask for some help in the chat

src/transformStateWithClones.js
```diff
@@ -6,8 +6,47 @@
...
+        break;
+    }
+  }
+
+  return massiveToReturn;
+}
+
+function addProperties(massiveToReturn, stateCopy, extraData) {
+  Object.assign(stateCopy, extraData);
+  massiveToReturn.push({ ...stateCopy });
```
etojeDenys, line 35: you can push it only one time after all cases

etojeDenys (changes requested):

src/transformStateWithClones.js
```diff
@@ -6,8 +6,45 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateCopy = { ...state };
+  const massiveToReturn = [];
```
etojeDenys, line 12: image

etojeDenys (approved):
good job

PR #3721, opened 2024-08-02
solution

VitaliiNez:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+  let obj = { ...state };
+  const arr = [];
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
+        obj = { ...obj, ...extraData };
+        arr.push({ ...obj });
```
etojeDenys, line 19: you can push it only one time after all cases

etojeDenys (changes requested):

volodymyr-soltys97 (approved):
Well done! 🔥

PR #3722, opened 2024-08-03
Reply

Mr36Tomek:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,25 @@
...
+      currentState = { ...currentState, ...action.extraData };
+    } else if (action.type === 'removeProperties') {
+      currentState = { ...currentState };
+
+      action.keysToRemove.forEach((key) => {
+        delete currentState[key];
+      });
+    } else if (action.type === 'clear') {
+      currentState = {};
+    }
```
natalia-klonowska, line 24: [checklist](https://github.com/mate-academy/js_stateful-clones/blob/master/checklist.md): using switch will improve readability because you clearly just need to handle 3 cases ;)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,25 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  let currentState = Object.assign({}, state);
+  const states = [];
+
+  actions.forEach((action) => {
+    if (action.type === 'addProperties') {
+      currentState = { ...currentState, ...action.extraData };
+    } else if (action.type === 'removeProperties') {
+      currentState = { ...currentState };
```
natalia-klonowska, line 17: you don't need to make another copy of currentState

natalia-klonowska (changes requested):

PR #3724, opened 2024-08-04
states

MOODDDII:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [];
+  let currentState = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
+    const action = actions[i];
```
volodymyr-soltys97, line 14: Use for of loop and destructuring for action
```suggestion
  for (const action of actions) {
    const { type, ....... } = action;
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+      currentState = { ...currentState };
+
+      for (let k = 0; k < action.keysToRemove.length; k++) {
+        const key = action.keysToRemove[k];
+
+        delete currentState[key];
+      }
+    }
+
+    if (action.type === 'clear') {
```
volodymyr-soltys97, line 30: Use switch case instead of 3 if conditions, it will be better

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

vadiimvooo (approved):
Well done

PR #3728, opened 2024-08-05
Develop

2bmakohon:


etojeDenys (changes requested):
image

src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const stateCopy = { ...state };
+  const result = [];
+
+  for (const action of actions) {
+    switch (action.type) {
```
volodymyr-soltys97, line 15: Use destructuring for `action`
```suggestion
  for (const action of actions) {
    const { type, ..... } = action;
    
    switch (type) {
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
...
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
```
volodymyr-soltys97, line 17:
```suggestion
        Object.assign(stateCopy, extraData);
```

src/transformStateWithClones.js
```diff
@@ -8,6 +8,36 @@
...
+        break;
+
+      case 'removeProperties':
+        for (let i = 0; i < action.keysToRemove.length; i++) {
```
volodymyr-soltys97, line 21: Consider will be better if you use for of loop here

volodymyr-soltys97 (changes requested):
Almost done!

volodymyr-soltys97 (approved):
Great! 🔥

PR #3747, opened 2024-08-17
Added logic

CamilleHaus:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,27 @@
...
+  const newState = [];
+  let stateCopy = { ...state };
+
+  actions.forEach((action) => {
+    if (action.type === 'clear') {
+      stateCopy = {};
+    } else if (action.type === 'addProperties') {
+      stateCopy = { ...stateCopy, ...action.extraData };
+    } else if (action.type === 'removeProperties') {
+      stateCopy = { ...stateCopy };
```
pedro-ruas, line 20: This line is not needed, won't change your code and results, but is adding a unnecessary computation process

pedro-ruas (approved):
Great job Camille! I left only a suggestion about an unnecessary line, besides that, great work! 😄

PR #3751, opened 2024-08-20
Solution

inesshtepa:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,40 @@
...
-  // write code here
+  const stateHistory = [];
+  let stateNew = JSON.parse(JSON.stringify(state));
+
+  for (let key = 0; key < actions.length; ++key) {
+    if (actions[key].type === 'addProperties') {
+      const stateNew1 = Object.assign({}, stateNew, actions[key].extraData);
+
+      stateHistory.push(stateNew1);
+      stateNew = JSON.parse(JSON.stringify(stateNew1));
```
etojeDenys, line 19: image

etojeDenys (changes requested):
also better to rewrite it using the switch/case instead of the multiple if/else

inesshtepa:
Thanks for your comments. I fixed all issues. Please check again.

PR #3760, opened 2024-08-25
solved task

PikalovAndrey:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resArray = [];
+  let stateCopy = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
```
VolodymyrKirichenko, line 13: u need each element for this case. use `for of` loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+    if (actions[i].type === 'addProperties') {
+      const extraData = actions[i].extraData;
+
+      for (const key in extraData) {
+        stateCopy[key] = extraData[key];
+      }
```
VolodymyrKirichenko, line 19: `Object.assign`?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
...
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      const keysToRemove = actions[i].keysToRemove;
+
+      for (let j = 0; j < keysToRemove.length; j++) {
```
VolodymyrKirichenko, line 25: use `for of` loop also

src/transformStateWithClones.js
```diff
@@ -7,7 +7,34 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resArray = [];
+  let stateCopy = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
```
VolodymyrKirichenko, line 14: Use `switch case` instead of if conditions for better readability when checking multiple values

VolodymyrKirichenko (changes requested):
Nice work, but u can do it better!
Just do it 👍

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resArray = [];
```
BudnikOleksii, line 10: res/result/array/obj/object are bad names for variables

BudnikOleksii (approved):
👍

PR #3762, opened 2024-08-25
Solution

TomaszKrzywicki:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,29 @@
...
-  // write code here
+  const stateHistory = [];
+  let currentState = { ...state };
+
+  actions.forEach((action) => {
+    if (action.type === 'clear') {
+      currentState = {};
+    } else if (action.type === 'addProperties') {
+      currentState = { ...currentState, ...action.extraData };
+    } else if (action.type === 'removeProperties') {
```
natalia-klonowska, line 19: use switch instead of multiple if/else statements

src/transformStateWithClones.js
```diff
@@ -6,8 +6,29 @@
...
+      currentState = { ...currentState };
```
natalia-klonowska, line 20: you already created copy of state so you don't need to make another copy

natalia-klonowska (changes requested):

src/transformStateWithClones.js
TomaszKrzywicki: done

natalia-klonowska (approved):

PR #3771, opened 2024-08-27
add task solution

dokrod:
solution

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,27 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copy = [];
+  let stateCopy = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateCopy, action.extraData);
+        break;
+
+      case 'removeProperties':
+        action.keysToRemove.forEach((key) => delete stateCopy[key]);
+        break;
+
+      case 'clear':
+        stateCopy = {};
+        break;
+    }
+    copy.push({ ...stateCopy });
+  }
+
+  return copy;
 }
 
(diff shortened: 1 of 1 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const copy = [];
```
KrivorotkoYurii, line 10: Can you maybe name a variable in the other way? In your function you return an array with history of states after actions. That`s why it is better to name a variable like "stateHistory" or "historyOfStates".

KrivorotkoYurii (approved):
Nicely done!🔥🔥🔥

src/transformStateWithClones.js
Moroz-Dmytro, line 10: This is not `copy`, this is `stateHistory`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
...
+        Object.assign(stateCopy, action.extraData);
+        break;
+
+      case 'removeProperties':
+        action.keysToRemove.forEach((key) => delete stateCopy[key]);
+        break;
+
+      case 'clear':
+        stateCopy = {};
+        break;
```
Moroz-Dmytro, line 25: image

Moroz-Dmytro (changes requested):
Good work 👍 
But fix comments below

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+
+      default:
+        break;
```
Moroz-Dmytro, line 28: But you should throw an error here

Moroz-Dmytro (approved):
I'll approve it, but check comment below

PR #3773, opened 2024-08-27
Solution

hma-3:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,44 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const currentState = { ...state };
+  const stateHistory = [];
+
+  for (const action of actions) {
+    switch (action.type) {
```
VolodymyrKirichenko, line 14: u can use destructuring for ur `action`, but up to u. For this case it not necessary)

VolodymyrKirichenko (approved):
Well done 🔥

PR #3774, opened 2024-08-27
Solution

davicatarino:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,31 @@
...
+  actions.forEach((action) => {
+    switch (action.type) {
+      case 'clear':
+        currentState = {};
+        break;
+      case 'addProperties':
+        currentState = { ...currentState, ...action.extraData };
+        break;
+      case 'removeProperties':
+        currentState = { ...currentState };
```
pedro-ruas, line 22: This line is unnecessary, you're not changing the variable, just re-creating it with the same value, so there's no need to use it

pedro-ruas (approved):
Davi, good job, however there is an unnecessary line there, which I have pointed out.
Remove it and your code will continue to work as it is.

PR #3794, opened 2024-08-29
solution

mbulchak:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        stateCopy = { ...stateCopy, ...action.extraData };
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (stateCopy[key]) {
```
Ostkreuzzz, line 21: you can skip if and delete keys straightaway from keysToRemove

case 'removeProperties':
          delete stateCopy[key]

Ostkreuzzz (approved):
GJ!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        stateCopy = { ...stateCopy, ...action.extraData };
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateCopy[key];
+        }
```
igorkruz: use proper variable names in for of loop, please use something like 'property'

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+        break;
+
+      case 'clear':
+        stateCopy = {};
+        break;
+
+      default:
+        break;
```
igorkruz: you can throw an error if there is no existing case 

```suggestion
      default:
         throw new Error('invalid action type');
```

igorkruz (changes requested):
Good job 🔥
Let's improve your solution

PR #3805, opened 2024-09-02
Solution

myehor:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,37 @@
...
+      for (const key of ch.keysToRemove) {
+        delete stateCopy[key];
+      }
+    }
+
+    if (ch.type === 'clear') {
+      for (const key in stateCopy) {
+        delete stateCopy[key];
+      }
+    }
```
SanyaBratashchuk, line 33: it is better to use switch case in such cases

SanyaBratashchuk (approved):

src/transformStateWithClones.js
myehor: Yes, i agree, next time I`ll try to use switch case. Thank you!

PR #3810, opened 2024-09-04
solution

YuriyBilskiy:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newAction = [];
+  let stateClones = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```
volodymyr-soltys97: Use destructuring for `actions`
```suggestion
  for (const action of actions) {
    const { type, ........ } = action;
    
    switch (type) {
```

volodymyr-soltys97 (changes requested):
Good job!
Let's make your code better

volodymyr-soltys97 (approved):
Great work! 🔥

PR #3811, opened 2024-09-04
Solution

bardyuzha:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        workingState[newProperty] = action.extraData[newProperty];
+      }
+    }
+
+    // Remove property from the workingState
+    if (action.type === 'removeProperties') {
+      for (const removingKey of action.keysToRemove) {
+        delete workingState[removingKey];
+      }
+    }
```
volodymyr-soltys97: Use a switch case construction instead of if conditions

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const workingState = { ...state };
+  const stateHistory = [];
+
+  for (const action of actions) {
+    // Clear workingState
```
```suggestion
  for (const action of actions) {
    const { type, ...... } = action;
    // Clear workingState
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

anastasiiavorobiova (approved):
Great work!

PR #3815, opened 2024-09-07
Solution

Anton777148:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
...
 function transformStateWithClones(state, actions) {
-  // write code here
+  const resultStates = [];
+
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    let newState = { ...currentState };
+
+    if (action.type === 'addProperties') {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code, you need to fix your code by AI comments

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+
+    if (actionType === 'addProperties') {
+      stateCopy = { ...stateCopy, ...action.extraData };
+    } else if (actionType === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete stateCopy[key];
+      }
+    } else if (actionType === 'clear') {
+      stateCopy = {};
+    }
```
etojeDenys: better to rewrite it using the switch case instead of multiple if/else statements

etojeDenys (changes requested):

etojeDenys (approved):

PR #3830, opened 2024-09-11
Solution

SashaDmyt:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+      for (const key of action.keysToRemove) {
+        delete stateCopy[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in stateCopy) {
+        delete stateCopy[key];
+      }
+    }
```
etojeDenys: better to rewrite it using the switch case instead of multiple if statements

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #3834, opened 2024-09-11
finish transform state with clones task

italomagno:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newStates = [];
+
+  actions.forEach((action) => {
+    switch (action.type) {
+      case 'addProperties':
+        const StateBfAdd = { ...(newStates[newStates.length - 1] || state) };
```
pedro-ruas: Italo, a variable should start with a lowercase letter, we use uppercase for classes, types and such, but not variables.
Also, `stateBfAdd` isn't very intuitive, you could call it something like `previousState`.
The same advice works for other declarations

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+        newStates.push({ ...StateBfAdd, ...action.extraData });
+        break;
+      case 'removeProperties':
+        const stateBfRemove = { ...(newStates[newStates.length - 1] || state) };
+
+        action.keysToRemove.forEach((key) => {
+          if (Object.keys(stateBfRemove).includes(key)) {
```
pedro-ruas: You don't need to test with an `if` here, if the property doesn't exist, nothing will happen.
So `delete stateBfRemove[key]` to an unknown key will just keet `stateBfRemove` as is

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+
+  actions.forEach((action) => {
+    switch (action.type) {
+      case 'addProperties':
+        const StateBfAdd = { ...(newStates[newStates.length - 1] || state) };
+
+        newStates.push({ ...StateBfAdd, ...action.extraData });
+        break;
+      case 'removeProperties':
+        const stateBfRemove = { ...(newStates[newStates.length - 1] || state) };
```
pedro-ruas: I see you intended to use only `const` and so you need to access the previous state using this index.
You could make it in two other ways that are better for performance:
1. Using `array.at` method:
```
const previousState = newStates.at(-1) ?? state;
```
2. Using another const to store the previous state before:
```
const newStates = [];
const stateHistory = { state };

switch ...
  ...
  const previousState = { ...stateHistory.state };
 ... do the transformations in previousState as needed
 stateHistory.state = { ...previousState };
 break;
```

pedro-ruas (changes requested):

pedro-ruas (approved):
Greatt!!

PR #3837, opened 2024-09-12
Solution

associate2coder:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,48 @@
...
+          delete copy[key];
+        }
+
+        stateHistory.push(copy);
+        referenceState = { ...copy };
+        break;
+      }
+
+      case 'clear': {
+        stateHistory.push({});
```
etojeDenys: you can push it only one time after the switch case

etojeDenys (changes requested):

src/transformStateWithClones.js
associate2coder: Thanks! Fixed it ))

SanyaBratashchuk (approved):

PR #3838, opened 2024-09-13
Solution

Sonya-Miss:


src/transformStateWithClones.js
```diff
@@ -1,13 +1,37 @@
...
+    } else {
+      newState = { ...state };
+
+      if (action.type === 'addProperties') {
+        Object.assign(newState, action.extraData);
+      } else if (action.type === 'removeProperties') {
+        for (const key of action.keysToRemove) {
+          delete newState[key];
+        }
+      }
```
etojeDenys: better to rewrite it using the switch case here instead of multiple if/else

etojeDenys (changes requested):

src/transformStateWithClones.js
Sonya-Miss: why switch case is better here?

lerastarynets (approved):
lgtm 🚀

PR #3844, opened 2024-09-16
Solution

svitlanak31:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
...
+    if (action.type === 'removeProperties') {
+      currentState = { ...currentState };
+
+      for (const key of action.keysToRemove) {
+        delete currentState[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      currentState = {};
```
volodymyr-soltys97: Use a switch case construction instead of if conditions, it will be better

src/transformStateWithClones.js
```diff
@@ -7,7 +7,29 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let currentState = { ...state };
+  const statesHistory = [];
+
+  for (const action of actions) {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/transformStateWithClones.js
lerastarynets: not fixed

lerastarynets (changes requested):
Pls fix prev comments

etojeDenys (approved):

PR #3870, opened 2024-09-25
feat:implement solution

ab3MN:
Implement solution

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -27,22 +27,25 @@ const clearState = (state, res) => {
 
 function transformStateWithClones(state, actions) {
   const newState = { ...state };
-  const res = [];
+  const statesHistory = [];
 
   actions.forEach((action) => {
     switch (action.type) {
       case 'addProperties':
-        return addProperties(newState, action.extraData, res);
+        return addProperties(newState, action.extraData, statesHistory);
 
       case 'removeProperties':
-        return removeProperty(newState, action.keysToRemove, res);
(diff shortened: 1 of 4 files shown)


src/transformStateWithClones.js
```diff
@@ -6,8 +6,43 @@
...
+const clearState = (state, res) => {
+  Object.keys(state).forEach((key) => delete state[key]);
+
+  res.push({ ...state });
+};
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newState = { ...state };
+  const res = [];
```
```suggestion
  const statesHistory = [];
```
`res/result/arr/obj/el/elem/` are bad names, variable names should be descriptive

src/transformStateWithClones.js
```diff
@@ -6,8 +6,43 @@
...
+  actions.forEach((action) => {
+    switch (action.type) {
+      case 'addProperties':
+        return addProperties(newState, action.extraData, res);
+
+      case 'removeProperties':
+        return removeProperty(newState, action.keysToRemove, res);
+
+      default:
+        return clearState(newState, res);
```
BudnikOleksii: - You should clear only in case of action type is `clear`
- In the default case you can throw an error, so someone who use your function can understand that they pass incorrect type

BudnikOleksii (changes requested):
Good work 👍 
Just a few moments should be fixed

src/transformStateWithClones.js
ab3MN: The comments have been corrected.

BudnikOleksii (approved):
Good job 👍

PR #3887, opened 2024-10-02
Develop

Krykunov:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,32 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete newState[key];
+        }
+        result.push({ ...newState });
+        break;
+
+      case 'clear':
+        newState = {};
+        result.push({ ...newState });
```
BudnikOleksii: In each case you do `result.push({ ...newState });`, you can do it once

src/transformStateWithClones.js
```diff
@@ -6,8 +6,32 @@
...
+    }
```
BudnikOleksii, line 32: switch should have a default case, you can throw an error if action type is invalid

src/transformStateWithClones.js
```diff
@@ -6,8 +6,32 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  let newState = { ...state };
+  const result = [];
```
BudnikOleksii: `res/result/arr/obj/el/elem/etc` are bad names, the variable name should be informative, for example `statesHistory`

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions, check them

BudnikOleksii (approved):
Looks great 🚀

PR #3906, opened 2024-10-13
solution

kami2693:


src/transformStateWithClones.js
```diff
@@ -1,13 +1,30 @@
...
- * @param {Object[]} actions
- *
- * @return {Object[]}
- */
-function transformStateWithClones(state, actions) {
-  // write code here
+function transformStateWithClones(initialState, actions) {
+  return actions.reduce((stateHistory, action) => {
+    const prevState = stateHistory[stateHistory.length - 1] || initialState;
+    let newState;
```
OkMoroz, line 6: **Well done on completing the task! 👍**  In the checklist, it is mentioned that clear names should be used for object copies. For example, instead of just using _newState_, you can use _stateCopy_ to better describe that it is a copy of the state.

OkMoroz (approved):

src/transformStateWithClones.js
```diff
@@ -1,13 +1,30 @@
...
+
+    switch (action.type) {
```
polosanya: You can destructure all the needed props of action

src/transformStateWithClones.js
```diff
@@ -1,13 +1,30 @@
...
+      case 'clear':
+        newState = {};
+        break;
+      case 'addProperties':
+        newState = { ...prevState, ...action.extraData };
```
polosanya: Object.assign can be used instead

polosanya (changes requested):
Almost done! Let's improve the code a bit

BudnikOleksii (approved):

PR #3907, opened 2024-10-13
Solution

ihorkorshylivskyi:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,54 @@
...
+          for (const ch of arr) {
+            delete removeState[ch];
+          }
+
+          result.push(removeState);
+          changeState = removeState;
+        }
+        break;
+
+      case 'clear':
```
Pirgei, line 39: For "clear", it's easier to just create a new empty object instead of a copy of the previous one.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,54 @@
...
+      case 'addProperties':
+        {
+          const addState = Object.assign({}, changeState, action.extraData);
+
+          result.push(addState);
+          changeState = addState;
+        }
+        break;
+
+      case 'removeProperties':
```
Pirgei, line 25: For "removeProperties", you can simply delete the keys in the copy instead of creating a new object for each change

Pirgei (approved):
Overall, your code works, but I left comments on how it can be improved

src/transformStateWithClones.js
ihorkorshylivskyi: Got it, thanks

src/transformStateWithClones.js
ihorkorshylivskyi: Thanks, I'll take it into account

PR #3915, opened 2024-10-18
Solution

iirk1:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+
+    if (action.type === 'addProperties') {
+      currentState = { ...currentState, ...action.extraData };
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete currentState[key];
+      }
+    }
```
volodymyr-soltys97: Consider will be better if you use switch case construction here instead of if conditions

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const res = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

vadiimvooo (approved):
Well done

PR #3928, opened 2024-10-25
Solution

MakksymSly:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+
+    if (action.type === 'addProperties') {
+      newState = { ...currentState, ...action.extraData };
+    } else if (action.type === 'removeProperties') {
+      newState = { ...currentState };
+
+      for (const key of action.keysToRemove) {
+        delete newState[key];
+      }
+    } else if (action.type === 'clear') {
```
BudnikOleksii: [CODE STYLE]: Use switch statement if you have limited amount of conditions.

BudnikOleksii (changes requested):
Good work 👍 
Let's use switch case instead

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+
+      default:
+        newState = { ...currentState };
+        break;
+    }
+    states.push(newState);
+    currentState = newState;
+  }
+
+  return states;
```
VolodymyrKirichenko, line 41: u don't need an additional object. This is redundant
u can do the same actions with ur `currentState` object
```suggestion
  const states = [];
  let currentState = { ...state };

  for (const action of actions) {
    switch (action.type) {
      case 'addProperties':
        currentState = { ...currentState, ...action.extraData };
        break;

      case 'removeProperties':
        for (const key of action.keysToRemove) {
          delete currentState[key];
        }
        break;

      case 'clear':
        newState = {};
        break;

      default:
        break;
    }
    
    states.push(newState);
  }

  return states;
```

VolodymyrKirichenko (approved):
Good job, but plz, review my comment and fix it 👍

PR #3943, opened 2024-10-29
develop

k0tlik02:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+
+        for (const key of action.keysToRemove) {
+          delete currentState[key];
+        }
+        break;
+      case 'clear':
+        currentState = {};
+        break;
+    }
+    stateHistory.push(currentState);
```
schyrva, line 33: Good job here!
 But, move `stateHistory.push(currentState)` line outside the loop.

schyrva (changes requested):
Good job! 
But i see 2 things to fix here. 
3. [CODE STYLE]: switch/case should always have a default case for error handling: 
you need to add a `default` case. Something like this:
 `default:
  throw new Error('Error');`
  
4.[DONT REPEAT YOURSELF] - If you perform the same action in all switch cases, do it just once afterwards:
`stateHistory.push(currentState)` line is repeated in each iteration of the loop. Since this action is common for all switch cases, it would be better to move it outside the switch statement.

schyrva (approved):
Good job!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+
+        for (const key of action.keysToRemove) {
+          delete currentState[key];
+        }
+        break;
+      case 'clear':
+        currentState = {};
+        break;
+      default:
+        throw new Error('Error');
```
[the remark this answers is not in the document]
BudnikOleksii: > The error message 'Error' is too generic. Consider providing a more descriptive error message that includes the action type or additional context to help with debugging.

Good point

src/transformStateWithClones.js
```diff
@@ -7,7 +7,33 @@
...
+    }
+    // stateHistory must be in'for' couse we add currentState for saving history
```
BudnikOleksii, line 32: Remove all comments

BudnikOleksii (approved):
Good work 👍 
Approved, but left a few suggestions

PR #3945, opened 2024-10-29
solution

petro-boiarenko:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,42 @@
  *
  * @return {Object[]}
  */
+const addProperties = 'addProperties';
+const removeProperties = 'removeProperties';
+const clear = 'clear';
+
 function transformStateWithClones(state, actions) {
-  // write code here
+  const STATES = [];
```
Daniil-102: Avoid writing all variables in uppercase. This convention is generally reserved only for constants or environment variables.
Here, STATES and CURRENT_STATE could be better named in camelCase (e.g., states and currentState) to follow JavaScript naming conventions for variables. 
and for other ones

src/transformStateWithClones.js
```diff
@@ -6,8 +6,42 @@
...
+  let CURRENT_STATE = { ...state };
+
+  for (const action of actions) {
+    let NEXT_STATE;
```
Daniil-102: The NEXT_STATE variable is unnecessary, you can change your current state directly.

Daniil-102 (changes requested):
Good job, but you need to make some changes to your code.

src/transformStateWithClones.js
petro-boiarenko: done

src/transformStateWithClones.js
```diff
@@ -6,8 +6,42 @@
...
+      case removeProperties:
+        nextState = { ...currentState };
+
+        for (const key of action.keysToRemove) {
+          delete nextState[key];
+        }
+        break;
+
+      default:
+        continue;
```
BudnikOleksii, line 37:
```suggestion
        throw new Error(`Invalid action type: ${action.type}`);
```
It's better to throw an error in case someone passes invalid action

BudnikOleksii (approved):
Good job 👍

PR #3950, opened 2024-10-29
add task solution

taniabarkovskya:


Krrampuss:
Wow! Good job!

Krrampuss (approved):
Wow! Good job!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
```
BudnikOleksii: result/res/arr/array/obj/object/el/elem/element and etc. are bad names. Name should describe variable

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+  let currentState = { ...state };
+
+  for (const act of actions) {
```
```suggestion
  for (const action of actions) {
```
What act is it, first or second? 😄

src/transformStateWithClones.js
```diff
@@ -7,7 +7,36 @@
...
+          delete modifiedState[key];
+        }
+        break;
+
+      case 'clear':
+        modifiedState = {};
+        break;
+
+      default:
+        return 'error';
```
```suggestion
        throw new Error(`Invalid action type: ${action.type}`);
```
It doesn't make much sense to return an error. You should throw it and the message should be informative

BudnikOleksii (changes requested):
Good job 👍 
Just left a few comments to improve your solution

VenomMB (approved):
Great job!!!

PR #3961, opened 2024-10-30
Solution

LiliiaVol:


Yevhen-Filimonov (approved):
Good job!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+  const stateCopy = { ...state };
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
+        for (const key in extraData) {
+          stateCopy[key] = extraData[key];
+        }
```
BudnikOleksii: This could be simplified with `Object.assign`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+        break;
+
+      case 'clear':
+        for (const key in stateCopy) {
+          delete stateCopy[key];
+        }
+        break;
+
+      default:
+        return ['error, try again with other arguments'];
```
```suggestion
        throw new Error(`Invalid action type: ${action.type}`);
```
It doesn't make much sense to return an error. You should throw it and the message should be informative

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions to improve your solution

polosanya (approved):
Well done! 👍

PR #3963, opened 2024-10-30
Solution

AsyaDev14:


natalia646 (approved):
Great job!

src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  let newState = { ...state };
+  const array = [];
```
```suggestion
  const statesHistory = [];
```
array/arr/obj/object/el/elem/res/result and so on are bad names. The name should describe variable

src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
...
+
+  for (const value of actions) {
```
BudnikOleksii: value as well is a bad name, I have no idea what value is it

src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
...
+    switch (value.type) {
+      case 'addProperties':
+        const editObj = { ...newState, ...value.extraData };
```
BudnikOleksii: The same, fix naming in all places

src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
...
+      case 'removeProperties':
+        for (const key of value.keysToRemove) {
+          delete newState[key];
+        }
+        array.push(newState);
+        break;
+
+      case 'clear':
+        newState = {};
+        array.push({});
```
BudnikOleksii: [DONT REPEAT YOURSELF] - If you perform same action in all switch cases - do it just once afterwards.

Update your newState and just push it once

src/transformStateWithClones.js
```diff
@@ -8,6 +8,37 @@
...
+
+        break;
+
+      default:
+        break;
```
BudnikOleksii: You can throw an error if someone passes an incorrect action type

BudnikOleksii (changes requested):
Good work 👍 
But fix naming and a few moments

src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
...
+
+      case 'clear':
+        newState = {};
+        break;
+
+      default:
+        throw new Error(`Unknown action type: ${action.type}`);
+    }
+    statesHistory.push({ ...newState });
```
undeletable, line 33: [Not critical]
In fact there's no need to perform an object copy here. The requirement is not to modify the original `state` object, and this is already handled by creating `newState` as its clone.

undeletable (approved):
Well done! I just have a couple of non-critical comments - please check that.

src/transformStateWithClones.js
```diff
@@ -8,6 +8,32 @@
...
+          delete newState[key];
+        }
+        break;
+
+      case 'clear':
+        newState = {};
+        break;
+
+      default:
+        throw new Error(`Unknown action type: ${action.type}`);
```
undeletable, line 31: [Not critical]
Instead of throwing an error, personally I'd prefer to return `newState` with no changes and log the error into console. Throwing an error means that you'd have to handle it in the places which use this function.

But that's beyond the requirements of current task, so you're not required to do this change. Just mentioning that there's another way of handling this case, so that you can consider it in a real work.

PR #3968, opened 2024-10-30
StateWithClones (Solution)

andrushchenkoo:


src/transformStateWithClones.js
```diff
@@ -7,7 +8,44 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [];
```
Eeoneishin: you should change name this const, maybe: stateHistory / historyState or something like that

src/transformStateWithClones.js
```diff
@@ -7,7 +8,44 @@
...
+
+      case 'clear': {
+        for (const key in currentState) {
+          delete currentState[key];
+        }
+        break;
+      }
+
+      default: {
+        return `Unknown action type: ${action.type}`;
```
Eeoneishin: it`s right, it`s work,  but in this task we need 'throw new Error'

Eeoneishin (approved):
i can approve it 👍 , check my advices :)

src/transformStateWithClones.js
BudnikOleksii: > you should change name this const, maybe: stateHistory / historyState or something like that

Good point

src/transformStateWithClones.js
BudnikOleksii: > it`s right, it`s work, but in this task we need 'throw new Error'

Good point

BudnikOleksii (changes requested):
Looks good 👍 
Check the previous reviewer comments

src/transformStateWithClones.js
```diff
@@ -7,7 +8,44 @@
...
+
+      case 'clear': {
+        for (const key in currentState) {
+          delete currentState[key];
+        }
+        break;
+      }
+
+      default: {
+        return 'throw new Error';
```
BudnikOleksii, line 42:
```suggestion
        throw new Error(`Unknown action type: ${action.type}`);
```
You should throw an error instead of returning a string

BudnikOleksii (approved):
👍

PR #3973, opened 2024-10-31
task solution

so3r:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+  const history = [];
+  let newState = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        newState = { ...newState, ...action.extraData };
+        break;
+      case 'removeProperties':
+        newState = { ...newState };
```
Daniil-102: This is an unnecessary string, it must be removed. That doesn't have any sense, you just make a copy of this object in this object.

Daniil-102 (changes requested):
Good job. You have good result, just delete one unnecessary string.

Daniil-102 (approved):
Great job!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete newState[key];
+        }
+        break;
+      case 'clear':
+        newState = {};
+        break;
+      default:
+        break;
```
BudnikOleksii, line 27: [CODE STYLE]: switch/case should always have default case for error handling.

BudnikOleksii (approved):
👍

PR #4006, opened 2024-11-12
Solution

OksanaW8:


KarolVlasova (changes requested):
push only files with your solution
did you use GPT?

KarolVlasova (approved):
"push only files with your solution" means that you shouldn't push .json files and others

PR #4039, opened 2024-11-27
Solution

fromjkeee16:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const STRING_ADD_CASE = 'addProperties';
+  const STRING_REMOVE_CASE = 'removeProperties';
+  const STRING_CLEAR_CASE = 'clear';
```
alexander-ignatow, line 12: You should declare such constants outside of the function - with your implementation that are needlessly created on each function call

alexander-ignatow (approved):
Looks good! 👍 

Just mind my comments for the future

src/transformStateWithClones.js
fromjkeee16: okay, thank you!

PR #4069, opened 2024-12-07
solution

Olifan:


etojeDenys (changes requested):
better to rewrite it using the swith case instead of if else

etojeDenys (approved):

PR #4070, opened 2024-12-08
js_stateful-clones solution

aCTORr8:
added task solution

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -10,24 +10,23 @@ function transformStateWithClones(state, actions) {
   let currentState = { ...state };
   const states = [];
 
-  for (const action of actions) {
-    if (action.type === 'addProperties') {
-      currentState = { ...currentState, ...action.extraData };
+  for (const { type, extraData, keysToRemove } of actions) {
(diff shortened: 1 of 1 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      for (const key of action.keysToRemove) {
+        delete currentState[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      currentState = {};
+    }
+
+    states.push(currentState);
```
volodymyr-soltys97: Use switch case instead of if conditions

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let currentState = { ...state };
+  const states = [];
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```

volodymyr-soltys97 (changes requested):
Good job!
Let's improve your code

volodymyr-soltys97 (approved):
Great work! 🔥

PR #4095, opened 2024-12-17
solution

b0hdan1:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let currentState = { ...state };
+  const result = [];
+
+  for (const action of actions) {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      currentState = { ...currentState };
+
+      for (const key of action.keysToRemove) {
+        delete currentState[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      currentState = {};
+    }
```
volodymyr-soltys97: Use switch case construction instead of if conditions

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/transformStateWithClones.js
b0hdan1: done

src/transformStateWithClones.js
b0hdan1: done

VitaliyBondarenko1982 (approved):
Good job

PR #4156, opened 2025-01-10
Solution

kovalmaria:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+      newState = { ...newState };
+
+      for (const key of action.keysToRemove) {
+        delete newState[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      newState = {};
+    }
```
alexander-ignatow: It looks good, but please consider using a switch/case here, since it fits better.

In this case, you doan  extra job by checking all three `if` conditions on each iteration.

alexander-ignatow (changes requested):
Good job, but please fix the issue in your code

alexander-ignatow (approved):
Now it looks better! 

But one important thing about the switch-case - always add the `default` section.

PR #4194, opened 2025-01-23
Добавлен файл transformState.js с функцией трансформации состояния

mmaksymov:


transformState.js
```diff
@@ -0,0 +1,29 @@
+function transformStateWithClones(state, actions) {
```
Stepico, line 1: You have two the same files right for some reason just with different names

transformState.js
```diff
@@ -0,0 +1,29 @@
...
+  const stateHistory = []; // Массив для хранения состояний
```
Stepico, line 2: All comments should be in English

Stepico (changes requested):
Some changes needed, please check the comments

PR #4196, opened 2025-01-23
Solution

IvanIlnytskyy:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+
+    if (action.type === 'removeProperties') {
+      action.keysToRemove.forEach(function (key) {
+        delete currentState[key];
+      });
+    }
+
+    if (action.type === 'clear') {
+      currentState = {};
+    }
```
volodymyr-soltys97: Use switch case here instead of if conditions, it will be more readable

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+  const states = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(currentState, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      action.keysToRemove.forEach(function (key) {
```
volodymyr-soltys97: Use arrow function
```suggestion
      action.keysToRemove.forEach((key) => {
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const states = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

vadiimvooo (approved):
Well done

PR #4274, opened 2025-02-10
add task solution

AnOvramenko:


PiekharievaK (approved):
Good job!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arrStates = [];
```
```suggestion
  const transformedStates = [];
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+  let stateCopy = { ...state };
+
+  for (const obj of actions) {
+    let newState;
```
ArturErmolenko: why we need one more variable? we could work with stateCopy

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const arrStates = [];
+  let stateCopy = { ...state };
+
+  for (const obj of actions) {
```
```suggestion
  for (const action of actions) {
```
its not obj, it's action inside actions

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+    let newState;
+
+    switch (obj.type) {
```
```suggestion
   const { type, extraData, keysToRemove } = action; --> obj in your version
    switch (type) {
    ...
```
use destruction

src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
...
+        break;
+      case 'clear':
+        newState = {};
+        break;
+      default:
+        throw new Error(`Unknown action type: ${obj.type}`);
+    }
+
+    arrStates.push({ ...newState });
+    stateCopy = newState;
```
```suggestion
```
confused ;)

src/transformStateWithClones.js
AnOvramenko: I was trying to solve the problem that AiBody said:
Direct Mutation of stateCopy: It's crucial to avoid directly mutating the stateCopy object within the switch cases. Instead, create a new copy of the object at the start of each switch case using Object.assign or the spread operator. This will help maintain the integrity of the state across different actions.

alexander-ignatow (changes requested):
Consider @ArturErmolenko comments

PR #4283, opened 2025-02-11
Solution js stateful clones

Nadiia-developer:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,25 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateHistory = [];
+  let currentState = { ...state };
```
OleksiiChystiakov: Please use a more detailed variable name.
Instead of `currentState`, use `stateCopy`. This makes it clear that we are working with a copy of the original state.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,25 @@
...
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
OleksiiChystiakov: Please use `switch` instead of `if-else`
Since `action.type` can only have a limited number of values, a `switch` statement is better than multiple `if-else` conditions.
Like this:
switch (action.type) {
      case 'addProperties':
      .....

src/transformStateWithClones.js
```diff
@@ -7,7 +7,25 @@
...
+      currentState = { ...currentState, ...action.extraData };
```
OleksiiChystiakov: Use `Object.assign` for clarity
Instead of `{ ...currentState, ...action.extraData }`, we can use `Object.assign(stateCopy, action.extraData);`
This improves understanding and is commonly used for merging objects. Spread operator creates a new object while Object.assign modifies target.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,25 @@
...
+    } else if (action.type === 'removeProperties') {
```
OleksiiChystiakov: It should be inside `switch-case`
This should be inside a `switch` statement instead of an `if-else` block.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,25 @@
...
+      currentState = { ...currentState };
+
+      for (const key of action.keysToRemove) {
+        delete currentState[key];
+      }
+    } else if (action.type === 'clear') {
```
OleksiiChystiakov: Instead of an `if-else`, this should be inside a `switch-case` and also, please add a `default` case to handle unexpected `action.type` values.

OleksiiChystiakov (changes requested):
In general, your code performs the necessary actions to meet the task requirements, but some improvements should be made. Please check my comments.

Stepico (approved):
Good job

PR #4285, opened 2025-02-11
solution

reb0rned:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,34 @@
  *
  * @return {Object[]}
  */
+
 function transformStateWithClones(state, actions) {
```
Iwantschenko, line 9: No comments. Have a nice CODE )

Iwantschenko (approved):

src/transformStateWithClones.js
```diff
@@ -6,8 +6,33 @@
...
-  // write code here
+  const arrStates = [];
```
```suggestion
  const transformedStates = [];
```
suits better

src/transformStateWithClones.js
```diff
@@ -6,8 +6,33 @@
...
+  const stateCopy = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
```
```suggestion
  for (const action of actions) {
  const { type, extraData, keysToRemove } = action;
    switch (type) {
    ...
```
Always use desctruction

alexander-ignatow (changes requested):
please fix @ArturErmolenko 's comments

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        currentState = { ...currentState, ...extraData };
+        break;
+      case 'removeProperties':
+        const newState = {};
+
+        for (const key in currentState) {
+          if (!keysToRemove.includes(key)) {
+            newState[key] = currentState[key];
+          }
+        }
```
alexander-ignatow, line 27: No need to change, but possibly this is not the best solution in terms of productivity
`includes` has a linear complexity (think of it like it iterates array internally) and you are doing it inside the loop- so it gives you square complexity

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+
+        currentState = newState;
+        break;
+      case 'clear':
+        currentState = {};
+        break;
+      default:
+        throw new Error(`Unknown action type: ${type}`);
```
alexander-ignatow, line 35: :+1:

alexander-ignatow (approved):
Good job!

PR #4298, opened 2025-02-12
made solution

maxbilfed:
Solution

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -8,24 +8,32 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
-  const newState = { ...state };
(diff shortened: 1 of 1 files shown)


src/transformStateWithClones.js
```diff
@@ -8,6 +8,27 @@
  */
 function transformStateWithClones(state, actions) {
   // write code here
+  const newState = { ...state };
+  const clonedStates = [];
+
+  for (const elem of actions) {
+    if (elem.type === 'addProperties') {
```
VladRoz95: You can use 'switch case statement' to perform different actions based on different conditions.

VladRoz95 (approved):

PR #4303, opened 2025-02-12
Solution

ansofidev:


SofiiaPelenska (approved):
Good job! But in this case, it is better to use `switch` instead of `if-else`.

PR #4318, opened 2025-02-14
first

anastasiia-zu:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,55 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const finalState = [];
+  let newState = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
```
alexander-ignatow: minor thing, but for / of loop would look slightly better here (you won't need extra variable for the index)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,55 @@
...
+    const action1 = actions[i];
+
+    switch (action1.type) {
```
alexander-ignatow: 1. Weird variable name (better options - `currentAction`, etc)
2. To avoid creating it, consider using for/of :)

src/transformStateWithClones.js
```diff
@@ -7,7 +7,55 @@
...
+      case 'addProperties':
+        const addedResults = addProperties(action1, newState);
+
+        finalState.push(addedResults);
```
alexander-ignatow: Avoid intermediate variables, where you don't need them

```suggestion
        finalState.push( addProperties(action1, newState));
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,55 @@
...
+    newState = { ...finalState[finalState.length - 1] };
+  }
+
+  return finalState;
+}
+
+function addProperties(action1, newState) {
+  for (const key in action1.extraData) {
+    newState[key] = action1.extraData[key];
+  }
```
alexander-ignatow, line 46: Why not for/ of?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,55 @@
...
+
+  return newState;
+}
+
+function removeProperties(action1, newState) {
+  for (let j = 0; j < action1.keysToRemove.length; j++) {
```
alexander-ignatow: Why not for/ of?

src/transformStateWithClones.js
```diff
@@ -7,7 +7,55 @@
...
+        throw new Error('Invalid type');
+    }
+
+    newState = { ...finalState[finalState.length - 1] };
+  }
+
+  return finalState;
+}
+
+function addProperties(action1, newState) {
```
alexander-ignatow: It would be a better practice to put such functions at the top of the file, as in such case it would be more human-readable

src/transformStateWithClones.js
```diff
@@ -7,7 +7,55 @@
...
+        break;
+
+      case 'removeProperties':
+        const removedResults = removeProperties(action1, newState);
+
+        finalState.push(removedResults);
+        break;
+
+      default:
+        throw new Error('Invalid type');
```
alexander-ignatow: 👍 

good, but lets make a message more informative. You may use `template literals`. (ex: `Invalid action type: ${action1.type}`)

alexander-ignatow (changes requested):
Good solution overall! You have a nice code structure, still there are some issues to fix

alexander-ignatow (approved):
Nice job!

PR #4320, opened 2025-02-15
Solution

abaskova:


OksanaKuziv (approved):
Your solution looks very nice!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const stateObject = [];
+
+  let currentState = { ...state };
+
+  for (let i = 0; i < actions.length; i++) {
+    const { type, extraData, keysToRemove } = actions[i];
```
alexander-ignatow, line 15: for..of loop works better here

```suggestion
  for (const  { type, extraData, keysToRemove } of actions) {
   ...
```

src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        break;
+
+      case 'addProperties':
+        nextState = { ...nextState, ...extraData };
+        break;
+
+      case 'removeProperties':
+        for (let j = 0; j < keysToRemove.length; j++) {
+          delete nextState[keysToRemove[j]];
+        }
```
alexander-ignatow, line 30: For..of is also a better option here

```suggestion
        for (const keyToRemovle of keysToRemove) {
          delete nextState[keyToRemove];
        }
```

alexander-ignatow (approved):
Overal, good job

Please review my comments and consider them for future!

PR #4324, opened 2025-02-17
Solution

Yevhen2802:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let currentState = { ...state };
+  const stateHistory = [];
+
+  for (const types of actions) {
```
volodymyr-soltys97: Consider will be better if you use `forEach` method instead of for loop

src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
...
+    }
+
+    if (types.type === 'clear') {
+      currentState = {};
+    }
+
+    if (types.type === 'removeProperties') {
+      currentState = { ...currentState };
+      types.keysToRemove.forEach((key) => delete currentState[key]);
+    }
```
volodymyr-soltys97: Use switch case instead of if condition, it will be better

src/transformStateWithClones.js
```diff
@@ -7,7 +7,27 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  let currentState = { ...state };
+  const stateHistory = [];
+
+  for (const types of actions) {
+    if (types.type === 'addProperties') {
```
volodymyr-soltys97: Use destructuring for `types`, also you can rename `types` to `action`

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (changes requested):
.

VitaliyBondarenko1982 (approved):
Looks good

PR #4382, opened 2025-03-05
add solution

ychaiun:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+      newState = { ...newState, ...extraData };
+    }
+
+    if (type === 'removeProperties') {
+      newState = removeProperties(newState, keysToRemove);
+    }
+
+    if (type === 'clear') {
+      newState = {};
+    }
```
volodymyr-soltys97: Use switch case instead of if conditions here

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Great! 🔥

PR #4522, opened 2025-04-14
Solution

KOleksandrivna:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,35 @@
...
+      for (const key in action.extraData) {
+        newState[key] = action.extraData[key];
+      }
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete newState[key];
+      }
+    }
```
etojeDenys: you can simplify it using the switch case

etojeDenys: image

etojeDenys (changes requested):

etojeDenys (approved):

PR #4538, opened 2025-04-16
StateWithClones 

YesupenkoVladislav:


natalliiaa (changes requested):
Remove calculateRentalCost.js and calculateRentalConst.test.js from the PR. They are not the files of current task

PR #4553, opened 2025-04-21
Update transformStateWithClones.js

mslol21:


mslol21:
function addProperties(state, extraData) {
  Object.assign(state, extraData);
}

function removeProperties(state, keysToRemove) {
    delete state[key];
  }
}

function clearProperties(state) {
  for (const key in state) {
    if (Object.hasOwn(state, key)) {
      delete state[key];
    }
  }
}

PR #4564, opened 2025-04-22
add stateful clones solution

DenisPitsul:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+        for (const keyToRemove of action.keysToRemove) {
+          delete currentState[keyToRemove];
+        }
+        break;
+
+      case 'clear':
+        for (const key in currentState) {
+          delete currentState[key];
+        }
+        break;
```
yelyzavetaz, line 34: Missing `default` in `switch`. This is important for catching invalid or unexpected action types.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,37 @@
...
+    }
+
+    stateHistory.push(structuredClone(currentState));
```
yelyzavetaz, line 39: You are repeating `stateHistory.push(structuredClone(currentState))` in each loop. It would be cleaner and less error-prone if handled just once at the end of each iteration.

yelyzavetaz (changes requested):
Great progress overall - just needs some cleanup ദ്ദി（• ˕ •マ.ᐟ

Missing `default` in `switch`: missing error handling for unknown action types.
Repetition: pushing to stateHistory happens every iteration instead of once at the end of each iteration.

src/transformStateWithClones.js
DenisPitsul: I don't understand this comment. 
'It would be cleaner and less error-prone if handled just once at the end of each iteration.' it works like you described.

yelyzavetaz: This comment was more of a suggestion. Your code works fine and looks great! But for better optimization and easier debugging, it would be better to move `stateHistory.push`  outside the loop. This way, you reduce the number of steps in each iteration. But totally optional - your current approach is valid too!

yelyzavetaz (approved):
Great job! /ᐠ > ˕ <マ
Thanks for addressing the feedback.

yelyzavetaz (approved):

PR #4565, opened 2025-04-22
Solution of transformStateWithClones

AndSLASH:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,26 @@
...
+      newState = { ...newState, ...act.extraData };
+    } else if (act.type === 'removeProperties') {
+      newState = { ...newState };
+
+      for (const key of act.keysToRemove) {
+        delete newState[key];
+      }
+    } else {
+      newState = {};
+    }
```
koliiis: 2. [CODE STYLE]: Use switch statement if you have limited amount of conditions.

koliiis (changes requested):
Good job!
But use a `switch` statement instead of `if`, and make sure it has default case.

PR #4567, opened 2025-04-22
add solution transformStateWithClones

yelyzavetaz:


src/transformStateWithClones.js
```diff
@@ -1,13 +1,44 @@
...
-/**
- * @param {Object} state
- * @param {Object[]} actions
- *
- * @return {Object[]}
- */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let currentState = state;
```
hryhoriikorsun, line 5: Now `currentState`  and `state` are the same. You should create copy, like: `currentState = { ...state }` or `Object.assign({}, state)`. And use your `currentState` замість `clonedState `.

src/transformStateWithClones.js
```diff
@@ -1,13 +1,44 @@
...
+            delete clonedState[key];
+          }
+        }
+        break;
+
+      default:
+        throw new Error(`Unknown action type: ${action.type}`);
+    }
+
+    result.push(clonedState);
```
hryhoriikorsun, line 37: If u use `currentState` corectly, u can write like this: `result.push(currentState);` and then delete line `currentState = clonedState;`.

hryhoriikorsun (approved):
God job👍 My comments like alternatives. If it work, it work)

PR #4572, opened 2025-04-23
add solution

AzartiksMan:


ystyslo (approved):
Great job 👏 But it's better to write 'Смерть путіну' in `default`.

PR #4583, opened 2025-04-24
solution

tempiqq:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
-}
```
RTSaitama, line 11: delete this stroke plz. function content is "under function" actually.

RTSaitama (approved):
Good work, i've found only 1 grammar mistake.
in clear we may create empty object, i think instead of use heaveweight 'delete' in loop, but it should work now correctly, so well done)

PR #4595, opened 2025-04-25
Solution

stanislav-sn:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const RESULT = [];
+  let CURRENT_STATE = { ...state };
```
jasialazyrka: Use camelCase for variables (for ex: currentState)
We Use upper case only for Constants. 

- currentState changes (you assign it new values)

- result changes (you push new states into it))

Pls fix it everywhere

jasialazyrka (changes requested):
Good job! It's almost done. 
Please see the comment and fix it

src/transformStateWithClones.js
```diff
@@ -8,28 +8,28 @@
  */
 function transformStateWithClones(state, actions) {
   const RESULT = [];
-  let CURRENT_STATE = { ...state };
+  let currentState = { ...state };
 
   for (const ACTION of actions) {
```
jasialazyrka: action, key , result: should be lower case

jasialazyrka (changes requested):
fix comments

PR #4618, opened 2025-05-01
Develop-2

valchuk-ivan:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,31 @@
...
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in currentState) {
+        delete currentState[key];
+      }
+    }
+    finalArray.push({ ...currentState });
+  }
```
koliiis: 2. [CODE STYLE]: Use switch statement if you have limited amount of conditions.

koliiis (changes requested):
Good job! 
But use the switch statement instead of if.

valchuk-ivan (commented):
Solution

PR #4634, opened 2025-05-07
Solution

Whslv:
Solve the task


src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newStates = [];
+  let newState = { ...state };
+
+  for (const action of actions) {
```
volodymyr-soltys97: You can use `forEach` method here

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
...
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete newState[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      newState = {};
```
volodymyr-soltys97: Use a switch case instead of if conditions, it will be better

src/transformStateWithClones.js
```diff
@@ -7,7 +7,28 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newStates = [];
+  let newState = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

etojeDenys (approved):

PR #4724, opened 2025-06-11
Solution

fabiocosta123:


fabiocosta123 (commented):
Modified if/else to switch and also renamed history to stateCopy

PR #4802, opened 2025-07-10
task solution

iastivka:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,26 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const history = [];
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'clear') {
```
horidor: 2. [CODE STYLE]: Use switch statement if you have limited amount of conditions.

для всіх if else у циклі. Також не забудь про

horidor (changes requested):
Good job! The code is concise and utilizes lesson material quite well, however there are some problems according to the checklist that you need to fix.

horidor (approved):
Good job! Code is now correct and in accordance with all the checklist requirements.

PR #4820, opened 2025-07-11
Added task solution

Yurii-Kupchynskyi:


maxrori228 (changes requested):
good work but you need to use switch case

2. [CODE STYLE]: Use `switch` statement if you have limited amount of conditions.

PR #4821, opened 2025-07-11
add task solution

maxrori228:


bebelekele (changes requested):
Seems to be all correct, but you should do a destructured assignment. It will make easier to get values from object.

maxrori228:
> Seems to be all correct, but you should do a destructured assignment. It will make easier to get values from object.


But this is an optional edit, the task can be done using my method

PR #4850, opened 2025-07-17
Solution

Ann-ovocheviy:


lvlashnin (approved):
The task was completed taking into account the checkpoints, the state was not mutate. Well done! But next time try to avoid if/else hell.

PR #4866, opened 2025-07-22
add clone

DenysNikiforov:


DenysNikiforov:
add Solution

DenysNikiforov:
`function transformStateWithClones(state, actions) {
  const resultingStates = [];
  
    let nextState = { ...currentState };

    switch (action.type) {
      case 'clear':
        nextState = {};
        break;

      case 'addProperties':
        if (action.extraData) {
          nextState = { ...currentState, ...action.extraData };
        }
        break;

      case 'removeProperties':
        if (Array.isArray(action.keysToRemove)) {
            delete nextState[key];
          }
        }
        break;
    }
    
    currentState = nextState;
    
    resultingStates.push(currentState);
  }

  return resultingStates;
}

module.exports = transformStateWithClones;`

PR #4964, opened 2025-08-31
add solution

Kobeliukh:


Kobeliukh:
I understand the suggestion about creating a new object each iteration for immutability. In my current implementation, the original state is not modified, and each state after an action is saved as a separate copy in logs. This meets the task requirements, so using a fully immutable approach is not required.

PR #4970, opened 2025-09-03
_stateful-clones solution_

S-Khraban:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,40 @@
...
+        );
+        break;
+      }
+
+      default:
+        break;
+    }
+
+    history.push({ ...current });
+  }
```
mkyloa, line 41: can we rewrite it and make it shorter and more readable (maybe without using the break statement)?

mkyloa (approved):
gj 👍

PR #4974, opened 2025-09-04
Add stateful clones task

plaaasio:


Zibi95 (changes requested):
Hey. Do you have any problems with pushing your changes on the origin branch?

PR #4979, opened 2025-09-08
ready

MaksimShe:


.github/workflows/test.yml-template
```diff
@@ -0,0 +1,23 @@
+name: Test
```
voloshynmykyta, line 1: remove this file from PR, please

package-lock.json
voloshynmykyta, line 1: remove this file from the PR, you have to push only ./src directory

package.json
voloshynmykyta, line 1: remove this file from the PR, you have to push only ./src directory

src/transformStateWithClones.js
```diff
@@ -7,7 +7,26 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const newStates = [];
```
voloshynmykyta: newStates - its ok, but it is more looks like stateHistory. It's not a point but just my thoughts.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,26 @@
...
+  let currentState = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
voloshynmykyta: please use constants for types
like

`const ADD_TYPE = 'addProperties'`

and use them like: 

`action.type === ADD_TYPE`

src/transformStateWithClones.js
```diff
@@ -7,7 +7,26 @@
...
+      currentState = { ...currentState, ...action.extraData };
+    } else if (action.type === 'removeProperties') {
```
voloshynmykyta: Use switch/case statement. It is in checklist.md file in your project.

src/transformStateWithClones.js
```diff
@@ -7,7 +7,26 @@
...
+      currentState = { ...currentState };
```
voloshynmykyta: I don't really sure you have to do that. You alreade have currentState === currentState, why do you do that copy again? 

you can just remove this line and all gonna work in the same way

voloshynmykyta (changes requested):
General: 

1. remove extra files from PR
2. use constants
3. use switch instead of if/else
4. remove unexpected assignment 
5. recheck checklist.md

PR #4981, opened 2025-09-09
implement state transfering by actions

voloshynmykyta:


Volodymyr-Bauzi (commented):
Мені подобається...!

Volodymyr-Bauzi (approved):
Так тримати..!

voloshynmykyta:
@mateacademy-ai-mentor structuredClone(state) its more complex solution than spread operator or Object.assign and now its supporting by all browsers. So it's really even better solution because it covers nested objects so we can copy even them. I know about spec expilicity but structuredClone works here even better than them. 

About requirement 3.2. There will be code duplication if ill write the new object, because i want to add `stateHistory.push({ ...stateCopy });` only once in the end of cycle instead of adding them everywhere. So please approve

PR #5003, opened 2025-09-11
Solution

Artur2912:


vtmnkmarta (changes requested):
You did a great job!😀 Your solution keeps the initial state immutable and applies actions correctly. You can still simplify the code in the addProperties case by using the spread operator, like this: newState = { ...lastState, ...action.extraData };
instead of looping through extraData.

vtmnkmarta (approved):

PR #5019, opened 2025-09-15
add solution

nazkov:


DmytroLymar (approved):
Good job, but think about optimization

You could do result.push({ ...currentState }) and create a new reference only for push and not for each operation, it would be easier to read your code

PR #5107, opened 2025-11-07
- ✍🏼 `transformStateWithClones` function to handle ➕add, ❌remove, and…

de-vasta:
… 🧹clear (based on `actions`) props of the `state` object without reassigning function's params and returning the array of modified object of every action.

### DIFF
--- a/src/transformStateWithClones.js
+++ b/src/transformStateWithClones.js
@@ -7,7 +7,38 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const ADD_ACTION = 'addProperties';
+  const REMOVE_ACTION = 'removeProperties';
+  const CLEAR_ACTION = 'clear';
+
+  // Array to hold clones of `state` after each action
+  const stateClones = [];
+  const prevClone = { ...state };
+
+  for (const action of actions) {
+    switch (action.type) {
+      case ADD_ACTION:
+        Object.assign(prevClone, action.extraData);
+        break;
+
+      case REMOVE_ACTION:
+        for (const key of action.keysToRemove) {
+          delete prevClone[key];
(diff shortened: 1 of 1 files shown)


src/transformStateWithClones.js
```diff
@@ -7,7 +7,38 @@
...
+        }
+        break;
+
+      case CLEAR_ACTION:
+        for (const key in prevClone) {
+          delete prevClone[key];
+        }
+        break;
+      default:
+        throw new Error(`Unknown action type: ${action.type}`);
```
DemonWayne: It’s better to wrap each `case` in `{}` in a `switch` statement

DemonWayne (changes requested):

src/transformStateWithClones.js
de-vasta: Ok! The requested little adjustments: a4cd18a

DemonWayne (approved):
Good work!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const ADD_ACTION = 'addProperties';
+  const REMOVE_ACTION = 'removeProperties';
+  const CLEAR_ACTION = 'clear';
+
+  // Array to hold clones of `state` after each action
```
```suggestion
```
remove comments

src/transformStateWithClones.js
```diff
@@ -7,7 +7,43 @@
...
+  const stateClones = [];
+  const prevClone = { ...state };
```
```suggestion
  const prevClone = { ...state };
```
why prev? It's looks like copy ;)

stateCopy sounds better for me

ArturErmolenko (approved):
Nice job 🔥

PR #5113, opened 2025-11-10
Solution

DemonWayne:


src/transformStateWithClones.js
```diff
@@ -6,8 +6,66 @@
...
+        clearState(stateCopy);
+        break;
+      }
+
+      default: {
+        break;
+      }
+    }
+
+    states.push(structuredClone(stateCopy));
```
dmytro-halieba: Why do we need structuredClone here?
We are cloning stateCopy after mutating it in each action handler.
This produces a deep copy of an already mutated object.
Could you clarify the intention here?

src/transformStateWithClones.js
```diff
@@ -6,8 +6,66 @@
...
+  for (const key of keys) {
+    delete state[key];
+  }
+}
+
+/**
+ * Clear state object
+ * @param {Object} state State object
+ */
+function clearState(state) {
```
dmytro-halieba, line 72: The clearState function removes all keys from the existing state object.
However, in the task description, clear might refer to producing a brand-new empty object rather than mutating the current one.
Should clear reset the state by creating {} or by deleting keys in place?
Just checking to ensure the behavior matches the intended specification.

src/transformStateWithClones.js
```diff
@@ -6,8 +6,66 @@
...
+
+  return states;
+};
+
+/**
+ * Add properties to state object
+ * @param {Object} state Current state
+ * @param {Object} data Data to add
+ */
+function addProperties(state, data) {
```
dmytro-halieba, line 47: All helper functions (addProperties, removeProperties, clearState) directly mutate the stateCopy object.
I'm not fully sure whether the task requires us to avoid mutating only the initial input state,
or to avoid any mutation at all and instead create a new state at every step.
Could you clarify whether mutating intermediate states is acceptable under the task requirements?

dmytro-halieba (approved):
Great work on the task. I really like how the code has become much more structured with the use of clearly defined function calls - I like it.
If you don’t mind, could you also share any mentor comments later on, especially if they provide suggestions or corrections regarding my feedback? That would be very helpful. Thanks in advance.

src/transformStateWithClones.js
DemonWayne: Actually, we don't need to use `structuredClone` here.
For the given tests, it will pass with simple object reconstruction.
But there are some situations when using `structuredClone` is better, like when there is some object reference inside an object that we are using, like:
```js
const address = { city: 'NYC', zip: '10001' } ;

const state = { 
  name: 'John', 
  address
};
```
So then, after modifying some of the values first object have we will have the update on every reference, and I think we don't need this for update history like this.
Over time, there may be some changes to values and variables, so in my opinion, this solution is better

src/transformStateWithClones.js
DemonWayne: So, for now, I think it needs to be re-assigned to `{}` (if I correctly understood AI feedback)

src/transformStateWithClones.js
DemonWayne: Good catch actually

PR #5137, opened 2025-11-20
Solution

dmytro-halieba:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,39 @@
...
+          delete nextState[key];
+        }
+        break;
+
+      case 'clear':
+        nextState = {};
+        break;
+
+      default:
+        continue;
```
DemonWayne: There, we need to throw an error about an unsupported type of action or something similar to this, at your choice.

DemonWayne (changes requested):
Mostly, this solution looks OK for me, and for the main problem, I am requesting changes below.

src/transformStateWithClones.js
dmytro-halieba: Fixed that in commit [43d3942]

DemonWayne (approved):
LGTM, nice work!

PR #5166, opened 2025-12-07
done Stateful clones task

maximtsyrulnyk:
done Stateful clones task


maximtsyrulnyk:
Edited according requirements

maximtsyrulnyk:
edited according requirements

PR #5180, opened 2025-12-10
Solution

mark-chep:


src/transformStateWithClones.js
```diff
@@ -8,6 +8,38 @@
...
+  const STATE_HISTORY = [];
+  const ADD_PROPERTIES = 'addProperties';
+  const REMOVE_PROPERTIES = 'removeProperties';
+  const CLEAR = 'clear';
+  let currentState = structuredClone(state);
+
+  for (const action of actions) {
+    let copyState = structuredClone(currentState);
+
+    if (action.type === ADD_PROPERTIES) {
```
SergiyKusiy: . [CODE STYLE] – Use switch!!! instead of multiple if statements

You currently use three separate if checks for action.type.
Since there is a limited and known set of action types, a switch statement is the proper choice according to the checklist.

src/transformStateWithClones.js
```diff
@@ -8,6 +8,38 @@
...
+      for (const key in action.extraData) {
+        copyState[key] = action.extraData[key];
```
SergiyKusiy: The checklist specifically recommends:
BAD: copy
GOOD: stateCopy
Your name copyState is not ideal.
Fix: Rename to stateCopy for clarity.

src/transformStateWithClones.js
```diff
@@ -8,6 +8,38 @@
...
+    if (action.type === REMOVE_PROPERTIES) {
+      for (const key of action.keysToRemove) {
+        delete copyState[key];
+      }
+      STATE_HISTORY.push(copyState);
+    }
+
+    if (action.type === CLEAR) {
+      copyState = {};
+      STATE_HISTORY.push(copyState);
```
SergiyKusiy: [DONT REPEAT YOURSELF]
Each branch of your logic ends with the same line:
STATE_HISTORY.push(copyState)
This violates the DRY principle.
Fix:
Call STATE_HISTORY.push once after the switch block.

SergiyKusiy (changes requested):
Good joob!) 
But try to make changes according to the comments, and your code will be perfect!

PR #5183, opened 2025-12-10
solution

Milana-Danylivska:


AdelinaY18 (approved):
Great job!

src/transformStateWithClones.js
```diff
@@ -7,7 +7,32 @@
...
+        break;
+
+      case 'removeProperties':
+        currentState = { ...currentState };
+
+        for (const key of action.keysToRemove) {
+          delete currentState[key];
+        }
+        break;
+    }
```
mark-chep, line 31: good job! But Luke complained about me because of this:
- You need to add default: break; in case none of the conditions match

mark-chep (changes requested):

src/transformStateWithClones.js
mark-chep: you need to add default: break, don't remove it, everything was fine

mark-chep (changes requested):

mark-chep (approved):
Great! Great solution. Good luck with the upcoming tasks.

PR #5191, opened 2025-12-11
Solution

Renata11111-afk:


NastyBaster (approved):
Nice and clean solution. I like your naming approach. Just a quick reminder to use UPPERCASE for constant names.

NastyBaster (approved):

PR #5343, opened 2026-02-28
add task solution

SergGun:


MaxShaggy (changes requested):
The task was completed correctly

PR #5390, opened 2026-03-25
done the task

whomngmnt:


linska (approved):
A couple of small suggestions:

- consider using a switch instead of multiple if statements for better clarity.
- minor naming: i could be renamed to action for readability.

But overall - clean and solid solution 🔥

PR #5403, opened 2026-04-01
Solution

yana-karpovych:


Evgeniy45 (approved):
Good solution!
I approved your job, but think about this:
when you are checking that action is "clear", do you realy need one more loop in this case.(you can use just an empty object)

PR #5405, opened 2026-04-01
add task solution

rlysunets:


DmitriyS2k (approved):
The task is completed, but some recommendations from the checklist not followed:

2. [CODE STYLE]: Use switch statement if you have limited amount of conditions.

Roma be more careful next time ;)

PR #5410, opened 2026-04-01
Stateful-clones_solution

Palfinger2002:


vaynexx (approved):
code is correct for the task. Also u can use "default"

PR #5414, opened 2026-04-02
Solved

Misharkus:


Tysiachnyk (approved):
That's a really interesting solution. 
I hadn't thought about it that way. 
Thanks for the new perspective!

PR #5421, opened 2026-04-02
Solution

YYarik:
all done


timurradkevic (approved):
Your function works correctly but should:
- Use a switch instead of if/else.
- Create a state copy once per action, not multiple times.
- Include a default case in the switch for unknown action types.
- Use consistent variable names for state copies (stateCopy instead of generic copy).

PR #5422, opened 2026-04-03
git add solution

andriy-stetsula:


impo10nt (commented):
good job mate

timurradkevic (changes requested):
The code works correctly and produces the expected results, but it does not fully follow the required code style guidelines. There is no `default` case in the `switch` statement, which is important for proper error handling. The variable name `array` is too generic and should be replaced with something more descriptive like `stateHistory`. There is also a bit of unnecessary duplication in the `removeProperties` case when copying the object. Overall, the logic is solid, but the code could be improved to better match the style requirements.

PR #5446, opened 2026-04-14
Solution

bstroshek:


timurradkevic (approved):
The implementation works correctly and handles state cloning properly 👍
However, according to the requirements, it would be better to use a switch statement instead of multiple if conditions, and include a default case for error handling.

PR #5603, opened 2026-08-03
add task solution

AchekcSs:


HiljSV (approved):
Solution is correct: all 12 tests pass locally and CI `build (20.x)` is green.
The checklist is respected - the incoming `state` is never mutated, `switch` is used,
and the common `push` is done once after the `switch`.

One small thing: item 3 of `checklist.md` asks for a `default` case *for error handling*,
and `default: break;` silently ignores an unknown action type.
`throw new Error('Unknown action type: ' + action.type);` matches the checklist better.
Not blocking.

PR #5607, opened 2026-08-04
solution

recurse1337:
Checklist:
- [x] don't mutate object or arrays - it will cause unexpected results later on. You should make copy using Object.assign or spread operator

- [x] Use switch statement if you have limited amount of conditions.

- [x] switch/case should always have default case for error handling.

- [x] If you perform same action in all switch cases - do it just once afterwards.

- [x] use proper names for object copy


ggylyon (approved):
Well done!

PR #5608, opened 2026-08-04
Solution

bakugan1-os:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const history = [];
+  let current = { ...state };
```
recurse1337: 5. [NAMING] - use proper names for object copy

BAD EXAMPLE:

const current = { ...state }

GOOD EXAMPLE:

const currentState = { ...state }

src/transformStateWithClones.js
```diff
@@ -7,7 +7,30 @@
...
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
recurse1337: 2. [CODE STYLE]: Use switch statement if you have limited amount of conditions.
If you keep the if statements, consider using else if to avoid unnecessary condition checks.

recurse1337 (approved):
Good job!

PR #5611, opened 2026-08-04
Solution

itisandrew25-ship-it:


src/transformStateWithClones.js
```diff
@@ -7,7 +7,35 @@
  * @return {Object[]}
  */
 function transformStateWithClones(state, actions) {
-  // write code here
+  const result = [];
+  let copyState = { ...state };
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
arkatovnikita: Good job! The solution is clear and easy to read. As a small improvement, you could replace the three separate if statements with if / else if, since an action can only have one type. This would make the code a bit cleaner and avoid unnecessary checks.

arkatovnikita (changes requested):

PR #5616, opened 2026-08-04
first attempt

arkatovnikita:


sofiatsebenko (commented):
good job!

skuzmenk (approved):
Good job! The solution is clear and easy to follow. As a small improvement, you could use if...else if instead of multiple if statements, since action.type can only match one condition. This avoids unnecessary checks.

